PR #46, opened 2018-02-28
Cleanup plugin code

jrudolph:


sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -0,0 +1,42 @@
...
+import sbtprotoc.ProtocPlugin
+
+object AkkaGrpcPlugin extends AutoPlugin {
+  import sbtprotoc.ProtocPlugin.autoImport._
+
+  // don't enable automatically, you might not want to run it on every subproject automatically
+  override def trigger = noTrigger
+  override def requires = ProtocPlugin
+
+  trait Keys { _: autoImport.type =>
```
dwijnand: What's the advantage/point of doing things this way?

jrudolph: Do you mean the complete change or the weird self type?

dwijnand: the trait/self type setup

jrudolph: Anyway, I cannot remember why I've done it this way (the best PRs are the ones without any comment on why things were changed one way or the other), so if you feel like it we should probably change it to a less confusing variant.

dwijnand: Nope, just wondering if there were some learnings you'd gained that led you to this configuration.

PR #135, opened 2018-03-28
[not for merge] Client compression tests (io.grpc client against akka-scala server)

raboof:


raboof:
This has made it into grpc-java 1.12.0 as expected so since #213 compression is tested (with the exception of #214) on master and this branch is no longer needed.

PR #144, opened 2018-04-04
Compress responses from the Java server side

raboof:


runtime/src/main/scala/akka/grpc/Codecs.scala
```diff
@@ -12,8 +12,8 @@ object Codecs {
   private val supported = supportedCodecs.map(_.name)
   private val byName = supportedCodecs.map(c ⇒ c.name → c).toMap
 
-  def negotiate(request: HttpRequest): Codec =
-    `Message-Accept-Encoding`.findIn(request.headers)
+  def negotiate(request: jm.HttpRequest): Codec =
```
ktoso: good

PR #145, opened 2018-04-05
[WiP] Maven plugin for Java

raboof:


maven-plugin/src/main/scala/akka/grpc/maven/Generate.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.maven.plugin.AbstractMojo
+import org.slf4j.LoggerFactory
+import protocbridge.{ JvmGenerator, Target }
+import org.apache.maven.project.MavenProject
+import scalapb.ScalaPbCodeGenerator
+
+import scala.annotation.tailrec
+import scala.beans.BeanProperty
+
+class Generate @Inject() (project: MavenProject) extends AbstractMojo {
```
ignasi35: I think the naming convention is using "Mojo" as a suffix: `GenerateMojo`.

.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: scala
+
+scala:
+  - 2.12.5
+
+jobs:
+  include:
+    - stage: publishLocal
+      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal
+    - stage: tests
```
2m: You do not need to repeat the `stage` name. It is possible to have consecutive `- script` entries which then uses previous stage name.

docs/src/main/paradox/server.md
```diff
@@ -52,7 +57,50 @@ protobuf {
...
+  <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
+    <akka.grpc.project.version>$projectversion$</akka.grpc.project.version>
+    <grpc.version>$grpc.version$</grpc.version>
+    <project.encoding>UTF-8</project.encoding>
+  </properties>
+</project>
+```
+@@@
```
2m: We should include these from the tester projects instead of having the xml in the markdown.

PR #149, opened 2018-04-05
(sbt) Watch .proto files in protobuf folder to trigger PB.recompile

### LINKED ISSUE #148: Hook  .proto files to sbt's resource watch

ignasi35:
Fixes #148

This PR is heavily inspired by https://github.com/lagom/sbt-lagom-descriptor-generator/blob/master/lagom-descriptor-generator-sbt-plugin/src/main/scala/com/lightbend/lagom/spec/sbt/LagomOpenApiPlugin.scala#L22-L56 . There are several `sbt` details that I'm not familiar with and I'd rather have a second pair of eyes review this.

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -38,8 +38,6 @@ object AkkaGrpcPlugin extends AutoPlugin {
       akkaGrpcCodeGenerators := GeneratorAndSettings(ScalaServerCodeGenerator, akkaGrpcCodeGeneratorSettings.value) :: Nil,
       akkaGrpcModelGenerators := Seq[Target]((JvmGenerator("scala", ScalaPbCodeGenerator), akkaGrpcCodeGeneratorSettings.value) -> sourceManaged.value),
 
-      PB.protoSources in Compile += target.value / "protobuf",
-
       sourceDirectory in PB.recompile := sourceDirectory.value / "protobuf",
       resourceDirectory in PB.recompile := sourceDirectory.value / "protobuf",
 
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -5,7 +5,7 @@
 package akka.grpc.sbt
 
 import protocbridge.{ JvmGenerator, Target }
-import sbt._
+import sbt.{ GlobFilter, _ }
 import Keys._
 import akka.grpc.gen.scaladsl.ScalaServerCodeGenerator
 import akka.grpc.gen.CodeGenerator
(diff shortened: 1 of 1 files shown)


sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -38,13 +38,42 @@ object AkkaGrpcPlugin extends AutoPlugin {
       akkaGrpcCodeGenerators := GeneratorAndSettings(ScalaServerCodeGenerator, akkaGrpcCodeGeneratorSettings.value) :: Nil,
       akkaGrpcModelGenerators := Seq[Target]((JvmGenerator("scala", ScalaPbCodeGenerator), akkaGrpcCodeGeneratorSettings.value) -> sourceManaged.value),
 
+      PB.protoSources in Compile += target.value / "protobuf",
+
+      sourceDirectory in PB.recompile := sourceDirectory.value / "protobuf",
+      resourceDirectory in PB.recompile := sourceDirectory.value / "protobuf",
```
ignasi35: `PB.protoSources in Compile += target.value / "protobuf"` added here so users don't have to add it on their build.sbt.

`sourceDirectory.value / "protobuf"` hardcoded (or we can switch to `proto`) to prefer convention over config.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -38,13 +38,42 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+
+      target in PB.recompile :=
+        crossTarget.value / "protobuf" / Defaults.nameForSrc(configuration.value.name),
```
ignasi35: IIUC, this goes hand in hand with line `41`

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -38,13 +38,42 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+      // TODO: review managedSources
+      managedSourceDirectories += (target in PB.recompile).value,
+      unmanagedResourceDirectories += (resourceDirectory in PB.recompile).value,
+      watchSources in Defaults.ConfigGlobal ++= (sources in PB.recompile).value,
```
ignasi35: This is the key setting.

ignasi35:
Needs #145 or we copy `.travis.yml` here.

ignasi35:
#145 was merged already. This is now rebased on top.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -38,13 +38,42 @@ object AkkaGrpcPlugin extends AutoPlugin {
       akkaGrpcCodeGenerators := GeneratorAndSettings(ScalaServerCodeGenerator, akkaGrpcCodeGeneratorSettings.value) :: Nil,
       akkaGrpcModelGenerators := Seq[Target]((JvmGenerator("scala", ScalaPbCodeGenerator), akkaGrpcCodeGeneratorSettings.value) -> sourceManaged.value),
 
+      PB.protoSources in Compile += target.value / "protobuf",
```
raboof: I'll have to take a bit of time to digest and play with this monday - haven't worked that closely with those plugins yet. Which task puts files into `target/protobuf`?

ignasi35: You are right, I think this is a copy/paste left-over we can get rid of.

ignasi35: To answer your question: no task puts files in `target/protobuf`.

raboof:
I can confirm this works, but I'm not sure I understand all of the PR. This seems to basically-work by boiling it down to `watchSources in Defaults.ConfigGlobal ++= PB.protoSources.value,` - what are we missing when we do it like that? Something around caching it looks like?

ignasi35:
> I can confirm this works, but I'm not sure I understand all of the PR. This seems to basically-work by boiling it down to watchSources in Defaults.ConfigGlobal ++= PB.protoSources.value, - what are we missing when we do it like that? Something around caching it looks like?

That works for regular sbt `~task`'s but won't work for [Lagom's devmode reloader](https://github.com/lagom/lagom/blob/55c5d1a92edf89c8ec2bd39b959e889c919627ee/dev/maven-plugin/src/main/scala/com/lightbend/lagom/maven/ServiceManager.scala#L105-L108) which watches the source folders using a different impl.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
ignasi35: Correction: these three lines are key (first two for dev-mode in play/lagom) and last for regular `~task`'s in `sbt`.

Issue #42, opened 2018-02-27
Integrate with Gradle for Scala projects


raboof:
Related to #6 

For Scala we could choose to invoke scalapb as a protoc plugin, or use https://github.com/CharlesAHunt/scalapb-gradle-plugin

2m:
Fixed in https://github.com/akka/akka-grpc/pull/154

gortiz:
Are there plans to upload the gradle plugin to Maven Central or somewhere else?

raboof:
Hi @gortiz - yes, ideally I think we would upload the plugin to the Gradle plugin portal (https://guides.gradle.org/publishing-plugins-to-gradle-plugin-portal), though right now we do not have this in place just yet.

PR #154, opened 2018-04-06
Gradle plugin for scala and java

### LINKED ISSUE #6: Integrate with Gradle protobuf plugin for Java projects
As long as we have a plain protoc plugin we should be able to integrate with gradle nicely as well: 

```
task.plugins {
  // Add grpc output without any option.  grpc must have been defined in the
  // protobuf.plugins block.
  // This yields
  // "--grpc_out=/path/to/output" on the protoc commandline.
  grpc { }
}
```

https://github.com/google/protobuf-gradle-plugin

Note that that's the official google gradle plugin

2m:
Generates both server and client side.

Fixes #6 and #42

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,21 @@ scala:
 
 jobs:
   include:
-    - stage: publishLocal
-      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal publishM2
     - stage: tests
       script: sbt test
     - script: sbt scripted
-    - script: cd plugin-tester-java && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate compile
-    - script: cd plugin-tester-scala && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate scala:compile
-    - script: cd plugin-tester-java && sbt compile
-    - script: cd plugin-tester-scala && sbt compile
-    - script: cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
-    - script: cd plugin-tester-scala && ./gradlew --include-build ../gradle-plugin clean compileScala
+    - script: sbt publishLocal && cd plugin-tester-java && sbt compile
+    - script: sbt publishLocal && cd plugin-tester-scala && sbt compile
+    - script: sbt publishM2 && cd plugin-tester-java && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate compile
+    - script: sbt publishM2 && cd plugin-tester-scala && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate scala:compile
+    - script: sbt publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
+    - script: sbt publishM2 && cd plugin-tester-scala && ./gradlew --include-build ../gradle-plugin clean compileScala
+
+before_cache:
+  - find $HOME/.ivy2 -name "ivydata-*.properties" -print -delete
+  - find $HOME/.sbt  -name "*.lock"               -print -delete
+
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt/boot
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,13 @@ scala:
 jobs:
   include:
     - stage: publishLocal
-      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal
+      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal publishM2
     - stage: tests
       script: sbt test
     - script: sbt scripted
-    - script: sbt akka-grpc-runtime/publishM2 && sbt akka-grpc-codegen/publishM2 && sbt akka-grpc-maven-plugin/publishM2 && cd plugin-tester-java && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate compile
-    - script: sbt akka-grpc-runtime/publishM2 && sbt akka-grpc-codegen/publishM2 && sbt akka-grpc-maven-plugin/publishM2 && cd plugin-tester-scala && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate scala:compile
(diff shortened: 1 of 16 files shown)


codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -26,8 +26,8 @@ object Main extends App {
 
   val req = CodeGeneratorRequest.parseFrom(inBytes)
   val out =
-    if (req.getParameter.toLowerCase.contains("language=scala")) ScalaServerCodeGenerator.run(req)
-    else JavaServerCodeGenerator.run(req)
+    if (req.getParameter.toLowerCase.contains("language=scala")) ScalaBothCodeGenerator.run(req)
+    else JavaBothCodeGenerator.run(req)
```
raboof: I guess we want to make this configurable - ticketify?

gradle-plugin/build.gradle
```diff
@@ -0,0 +1,33 @@
+plugins {
+    id 'groovy'
+    id 'java-gradle-plugin'
+    id 'maven-publish'
+    id 'com.palantir.git-version' version '0.10.1'
+}
+
+group = "com.lightbend.akka.grpc"
+version versionDetails().gitHashFull.substring(0, 8)
```
raboof: :+1: for now at least :+1:

plugin-tester-java/build.gradle
```diff
@@ -1,49 +1,26 @@
+buildscript {
+  repositories {
+    mavenLocal()
+  }
+
+  dependencies {
+    classpath 'com.lightbend.akka.grpc:gradle-plugin:0.0.1'
```
raboof: Should we call it akka-grpc-gradle-plugin?

.travis.yml
```diff
@@ -6,11 +6,13 @@ scala:
 jobs:
   include:
     - stage: publishLocal
-      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal
+      script: sbt ++$TRAVIS_SCALA_VERSION publishLocal publishM2
```
raboof: Hard to say how this pans out... also it doesn't seem like it's being picked up - which is strange, does that mean `publishLocal` didn't actually have any effect either?

build.sbt
```diff
@@ -55,6 +58,9 @@ lazy val scalapbProtocPlugin = Project(
       art.withClassifier(Some("assembly"))
     },
     mainClass in assembly := Some("akka.grpc.scalapb.Main"),
+    assemblyOption in assembly := (assemblyOption in assembly).value.copy(
+      prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultShellScript)
+    ),
```
raboof: Weird that this didn't seem necessary before... but :+1:

gradle-plugin/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+
+jar {
+    manifest {
+        attributes 'Implementation-Version': version
+    }
+}
+
+dependencies {
+    // to bring in protobuf dependency trasitively
+    implementation 'com.google.protobuf:protobuf-gradle-plugin:0.8.5'
```
johanandren: positively trasitive

gradle-plugin/build.gradle
2m: Yea. We might wanna tag `v0.0.1` soon just to be able to find all the places that need to be changed.

plugin-tester-java/build.gradle
2m: Yea. I think that way it will align with other plugins better.

.travis.yml
2m: Ahh, from https://docs.travis-ci.com/user/build-stages/#Data-persistence-between-stages-and-jobs:

raboof: Right - so we could perhaps use the pattern from https://docs.travis-ci.com/user/build-stages/warm-cache/ to warm up the ivy cache (when we configure it as a cache) but not rely on it for the pipeline. Makes sense. For now remove the publishLocal stage then?

codegen/src/main/scala/akka/grpc/gen/Main.scala
2m: https://github.com/akka/akka-grpc/issues/155

.travis.yml
2m: Yea. Will remove the stage.

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+    Project project
+
+    @Override
+    void apply(Project project) {
+        this.project = project
+        project.gradle.addListener(this)
+        final boolean isScala = project.plugins.hasPlugin("scala")
+
+        project.configure(project) {
```
raboof: This looks pretty cool

plugin-tester-java/build.gradle
```diff
@@ -1,49 +1,26 @@
+buildscript {
+  repositories {
+    mavenLocal()
+  }
+
+  dependencies {
+    classpath 'com.lightbend.akka.grpc:akka-grpc-gradle-plugin:0.0.1'
```
raboof: Maybe add a comment here explaining that this will be replaced with the 'link' to the currently-checked-out-version?

plugin-tester-scala/src/main/scala/io/grpc/examples/helloworld/GreeterServiceImpl.scala
```diff
@@ -1,5 +1,5 @@
 //#full-service-impl
-package io.grpc.examples
+package io.grpc.examples.helloworld
```
raboof: Do we still want to move these files when we configure `flat_package` by default?

2m: Yes, if we want to have the same semantics as java files.

raboof: I think that makes sense as a default

plugin-tester-scala/src/main/protobuf/helloworld.proto
```diff
@@ -1,7 +1,7 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "io.grpc.examples";
+option java_package = "io.grpc.examples.helloworld";
```
raboof: This is more consistent with upstream though, so OK... we should get our `helloworld.proto`s consistent again though :)

PR #156, opened 2018-04-10
Spell 'gRPC' with lowercase 'g'

raboof:


patriknw:
+1

Interesting that their official logo is with capital letters, but text is gRPC: https://grpc.io/docs/

renatocaval:
I guess the reason why they have a logo with all uppercase is because of the arrow decoration they have crossing it. If it was was a lowercase g, it won't align correctly. Pure graphical design issue.

raboof:
Bah, form should follow function :laughing:

patriknw:
Yeah, completely unacceptable oversight. We should submit a PR. ;-)

PR #160, opened 2018-04-10
Configuring where to find your proto files

raboof:


docs/src/main/paradox/client.md
```diff
@@ -179,3 +179,15 @@ java.util.logging.ConsoleHandler.formatter=java.util.logging.SimpleFormatter
...
+
+You can configure where your .proto files are located like this:
+
+sbt
+:   @@snip[build.sbt]($root$/../plugin-tester-java/build.sbt) { #protoSources }
+
+TODO gradle, maven
+
+TODO describe settings for code generation: scala/java, client/server/both
+
```
johanandren: I started on these as well in the Both-PR. I'm thinking it makes more sense with a separate section for each build tool than mashing them all together?

plugin-tester-java/build.sbt
```diff
@@ -6,12 +6,16 @@ enablePlugins(AkkaGrpcPlugin)
 
 javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.7" % "runtime"
 
+//#protoSources
 inConfig(Compile)(Seq(
+  PB.protoSources += new File("src/main/proto"),
+  //#protoSources
```
johanandren: Hmm. I'd want this to be just `PB.protoSources += ...` also, I think the sbt docs says you should _never_ use relative paths but always base of off the sources key or such.

docs/src/main/paradox/client.md
raboof: Yeah it's tricky having so many axes: client/server, java/scala, maven/gradle/sbt, I'm undecided. Perhaps we should keep the 'getting started' above 'shared' between build tools, but do separate 'configuration' sections for each one?

plugin-tester-java/build.sbt
raboof: That makes sense, let me give that a quick try

docs/src/main/paradox/client.md
johanandren: Yeah, that was exactly my thinking!

plugin-tester-java/build.sbt
```diff
@@ -6,12 +6,16 @@ enablePlugins(AkkaGrpcPlugin)
 
 javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.7" % "runtime"
 
+//#protoSources
 inConfig(Compile)(Seq(
+  PB.protoSources += sourceDirectory.value / "proto",
+  //#protoSources
```
raboof: Oh the commend below this line needs to be removed now

plugin-tester-java/build.sbt
```diff
@@ -6,12 +6,16 @@ enablePlugins(AkkaGrpcPlugin)
 
 javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.7" % "runtime"
 
+//#protoSources
 inConfig(Compile)(Seq(
+  PB.protoSources += sourceDirectory.value / "proto",
```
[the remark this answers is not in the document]
ignasi35: Isn't `sourceDirectory.value`  pointing to `src`? shouldn't the code be:

```
inConfig(Compile)(Seq(
  PB.protoSources += sourceDirectory.value / "main" / "proto",
```

In any case, it'd be good to explain in the docs what is the matching folder for `xyz `in  `PB.protoSources += xyz`

docs/src/main/paradox/index.md
```diff
@@ -15,10 +15,12 @@ connection.
 Based on a protobuf service definition, akka-grpc can generate:
 
 * Model classes (using plain protoc for Java or scalapb for Scala)
-* The API (as an interface for Java or a trait for Scala), expressed in Akka Streams `Source`s
+* The API @scala[trait]@java[interface], expressed in Akka Streams `Source`s
```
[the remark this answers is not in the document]
raboof: In the first bullet I figured it would be useful to make the reader aware that we're generating something else for Java than for Scala. 

For the second bullet that didn't seem so important anymore and kind of distracting - but a bit inconsistent, I guess I agree :)

plugin-tester-java/build.sbt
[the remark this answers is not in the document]
raboof: Though indeed perhaps spelling out the configuration and the resulting path would be nice,  will do :)

ignasi35:
BTW: I just realised that when testing this I made the mistake of testing on a project that had `.proto` files in `src/main/protobuf` but that is the default so I just realised that testing this should be with `src/main/proto-somthing-else`.

raboof:
Ah - the plugin-tester-java should be suitable as it places its proto files in src/main/proto

raboof:
(processed review comments)

docs/src/main/paradox/client.md
ktoso: +1 for separation by build tool here

docs/src/main/paradox/index.md
```diff
@@ -15,10 +15,12 @@ connection.
 Based on a protobuf service definition, akka-grpc can generate:
 
 * Model classes (using plain protoc for Java or scalapb for Scala)
-* The API (as an interface for Java or a trait for Scala), expressed in Akka Streams `Source`s
+* The API trait or interface, expressed in Akka Streams `Source`s
```
ktoso: `@scala[trair or]` java people react with panic sometimes to scala wording

docs/src/main/paradox/server.md
```diff
@@ -99,13 +99,12 @@ Maven
...
 
-
-TODO describe settings for code generation: scala/java, client/server/both
+For a complete overview of the configuration options see the chapter for your build tool, @ref[sbt](sbt-plugin.md), Gradle or Maven. (TODO Gradle and Maven)
 
 ## Generate and implement
 
 Define the interfaces you want to implement in your project's
-@sbt[src/main/protobuf]@gradle[src/main/proto] file(s).
+@sbt[src/main/protobuf]@gradle[src/main/proto]@maven[src/main/proto] file(s).
```
ktoso: those are all the same for all build tools tho?

docs/src/main/paradox/sbt-plugin.md
```diff
@@ -50,6 +50,13 @@ akkaGrpcCodeGenerators := Seq(GeneratorAndSettings(JavaServerCodeGenerator, akka
...
+You can configure where your .proto files are located like this:
+
+sbt
+:   @@snip[build.sbt]($root$/../plugin-tester-java/build.sbt) { #protoSources }
+
 ## Loading proto files from artifacts
 
-TODO more stuff - generating from classpath rather than directory for example
\ No newline at end of file
+TODO more stuff - generating from classpath rather than directory for example
```
ktoso: Ah too bad, I came here looking to see how you did it here :-)

docs/src/main/paradox/index.md
johanandren: 🙀

docs/src/main/paradox/server.md
johanandren: I vote for follow the build tool convention

Issue #159, opened 2018-04-10
Default to both client and server generation from sbt


johanandren:
And document how to do only one of them

PR #161, opened 2018-04-10
Default to generate both server and client

johanandren:
Simplified sbt plugin config
and some more docs

Refs #159

### DIFF
--- a/docs/src/main/paradox/index.md
+++ b/docs/src/main/paradox/index.md
@@ -23,5 +23,6 @@ Based on a protobuf service definition, akka-grpc can generate:
 
 * [server](server.md)
 * [client](client.md)
+* [sbt plugin details](sbt-plugin.md)
 
 @@@
--- a/docs/src/main/paradox/sbt-plugin.md
+++ b/docs/src/main/paradox/sbt-plugin.md
@@ -1,0 +1,53 @@
+# sbt plugin details
+
+The sbt plugin is added to a project like so:
+
+@@@vars
+```scala
+// in project/plugins.sbt:
+addSbtPlugin("com.lightbend.akka.grpc" % "akka-grpc-sbt-plugin" % "$projectversion$")
+// in build.sbt:
+enablePlugins(AkkaGrpcPlugin)
+```
+@@@
+
+It will then by default look for `.proto` files under `src/main/protobuf`.
+
+## Configuration
+
+By default the plugin generates both a client and a server stub for Scala. 
+
+It can be configured to just generate either server or client like so:
+
+```scala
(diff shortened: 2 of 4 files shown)


docs/src/main/paradox/sbt-plugin.md
```diff
@@ -0,0 +1,53 @@
...
+
+It will then by default look for `.proto` files under `src/main/protobuf`.
+
+## Configuration
+
+By default the plugin generates both a client and a server stub for Scala. 
+
+It can be configured to just generate either server or client like so:
+
+```scala
```
raboof: Should we use Scala/Java tabs here?

docs/src/main/paradox/index.md
```diff
@@ -23,5 +23,6 @@ Based on a protobuf service definition, akka-grpc can generate:
 
 * [server](server.md)
 * [client](client.md)
+* [sbt plugin details](sbt-plugin.md)
```
raboof: 'sbt configuration'? but :+1: as discussed on slack

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -30,12 +30,15 @@ object AkkaGrpcPlugin extends AutoPlugin {
   }
   import autoImport._
 
-  override def projectSettings = configSettings(Compile) ++ configSettings(Test)
+  override def projectSettings = defaultSettings ++ configSettings(Compile) ++ configSettings(Test)
+
+  private def defaultSettings =
+    Seq(
+      akkaGrpcCodeGeneratorSettings := Seq("flat_package"),
+      akkaGrpcCodeGenerators := GeneratorAndSettings(ScalaBothCodeGenerator, akkaGrpcCodeGeneratorSettings.value) :: Nil)
```
raboof: :+1:

docs/src/main/paradox/sbt-plugin.md
johanandren: You mean for the generate-scala/java sources? I thought about that when writing it and that this is more for completeness in a complicated setup than two alternatives as it normally is with Java/Scala

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
johanandren: Couldn't move the third setting out of the config-scope for some reason, but couldn't figure out why

PR #164, opened 2018-04-11
Add cancellation barrier

raboof:
Refs #147

Issue #158, opened 2018-04-10
Documentation/getting started improvements - Scala/sbt


johanandren:
This is what I found/had problems/questions about with when starting from zero on the Scala side of things, a bit unfocused but I guess we can split of separate tickets as needed later.

## Server

* What does java_package in the .proto-file do? I played around with it a bit and couldn't see that it makes a difference, even removing it worked fine, remove it from the sample or at least provide a comment explaining?
* Question of taste: do we recommend the `MyService` + `MyServiceImpl` naming style (I think I'd personally would prefer `MyServiceApi` and `MyService`)?
* We should document extra that it must be bound using Http2 (as it binds fine with Http 1 but doesn't work and gives weird errors) and that the ALPN agent is needed for that.

## Client

* What about lifecycle, should it be managed, does it need to be stopped?
* Does it keep connections alive, what happens if those break, does it reconnect automagically, do I need to care?
* Should I create a new one for every call or reuse a single one for all calls, is it thread safe?

## General

* Sharing .proto-files through an artifact or copying. Should we list some pro's/con's or simply recommend a versioned artifact when all involved parts are JVM for example?
* Should we say something about structuring a larger/more complex service? Should one prefer having a single gRPC API or multiple ones?

patriknw:
> Question of taste: do we recommend the MyService + MyServiceImpl naming style (I think I'd personally would prefer MyServiceApi and MyService)?

Yeah, tasty. I prefer MyService + MyServiceImpl. Isn't that the most common java style of such things?

raboof:
We used to append `Service` to the service name defined in the `proto`, but that lead to `TestServiceService`, so we dropped that ;). Perhaps the chance of a `FooApiApi` is smaller, but it seems safer to leave the service name from the contract untouched?

patriknw:
We should not add something automatically. I think this was only about what convention we would use in the examples.

raboof:
Ah - there I'm fine with either, but in the `TestService` case this is something that is also used upstream (https://github.com/grpc/grpc-java/blob/master/interop-testing/src/main/proto/io/grpc/testing/integration/test.proto) so it makes sense to follow that convention?

johanandren:
Ok! No strong feelings on that one, I guess one can argue either way, just wanted to make sure it was thought through. Maybe we should explicitly note in the docs what names of things get generated etc. even if that is also shown in the source samples.

raboof:
Server:
[...]

PR #165, opened 2018-04-11
Docs and references for the .proto files format

raboof:


raboof:
Failures are #164

docs/src/main/paradox/proto.md
```diff
@@ -0,0 +1,40 @@
...
+
+@@snip [helloworld.proto]($root$/../plugin-tester-scala/src/main/protobuf/helloworld.proto) { #messages }
+
+## Services
+
+Those messages can be used to define services:
+
+@@snip [helloworld.proto]($root$/../plugin-tester-scala/src/main/protobuf/helloworld.proto) { #services }
+
+Both the request and the response consists of either a single message or a stream of messages.
```
johanandren: Something I wondered as a newbie was, how can I make a Unit/void function, or one with no parameters. Maybe add that it is not possible to do that here? (Unless I'm wrong ofc)

docs/src/main/paradox/proto.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Code generation options
+
+There are a number of options that can be set in the .proto definition that influence how the code is generated:
+
+@@snip [helloworld.proto]($root$/../plugin-tester-scala/src/main/protobuf/helloworld.proto) { #options }
+
+The 'package' in the `.proto` is used to resolve references from one `.proto` file to another.
+It can also be used for the package name in the generated code, but it is
+common to use the separate `java_package` option to override it.
```
johanandren: Is the photo package required or optional?

johanandren: Damn you, autocorrect!

raboof: :camera:

docs/src/main/paradox/proto.md
raboof: Good one: I think the convention is to use the 'empty' object that Google provides for this (https://github.com/google/protobuf/blob/master/src/google/protobuf/empty.proto), but it might take a small section (also documenting how to generate code for proto's from dependencies). I'll try to work it in.

johanandren:
Refs #158

PR #183, opened 2018-04-26
Configure whitesource

### LINKED ISSUE #146: Add whitesource

raboof:
fixes #146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,18 @@ jobs:
     - script: sbt publishM2 && cd plugin-tester-scala && mvn -Dakka.grpc.project.version=`echo $TRAVIS_COMMIT | head -c8` akka-grpc:generate scala:compile
     - script: sbt publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
     - script: sbt publishM2 && cd plugin-tester-scala && ./gradlew --include-build ../gradle-plugin clean compileScala
+    - stage: whitesource-master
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -14,6 +14,18 @@ jobs:
...
     - script: sbt publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
     - script: sbt publishM2 && cd plugin-tester-scala && ./gradlew --include-build ../gradle-plugin clean compileScala
+    - stage: whitesource-master
+      script: sbt whitesourceCheckPolicies whitesourceUpdate
+    - stage: whitesource-release
+      script: sbt whitesourceCheckPolicies whitesourceUpdate
+
+stages:
+  - name: tests
+    if: NOT tag =~ ^v
```
dwijnand: Why not run tests on tag builds?

2m: Tagged commit is already tested when that commit lands on master. Skipping tests on release makes the release process a bit quicker as well.

dwijnand: Hmm, yeah I think in most cases this would be absolutely fine. True!

(The case I'm thinking about is if a tag is created before the target commit finished testing.)

2m: There is another possible race condition here. If a tag is created before the target commit even starts testing (or starts getting the sources) then both jobs become release jobs. That situation that can happen when cutting a release immediately after merging the final PR.

dwijnand: Yep it affects all sbt-dynver users too.  But I think you have that fixed by another condition below that only publishes on tags.

Issue #103, opened 2018-03-16
make helloworld.proto consistent


raboof:
We use an example `helloworld.proto` in a number of places, but they have grown out of sync. Let's make them consistent again.

johanandren:
They are now in sync.

PR #184, opened 2018-04-26
Make proto's and service names consistent

raboof:
Refs #103

### DIFF
--- a/sbt-plugin/src/sbt-test/gen-scala-server/01-gen-basic-server/src/main/protobuf/helloworld.proto
+++ b/sbt-plugin/src/sbt-test/gen-scala-server/01-gen-basic-server/src/main/protobuf/helloworld.proto
@@ -1,13 +1,13 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "io.akka.grpc";
+option java_package = "io.grpc.examples.helloworld";
 option java_outer_classname = "HelloWorldProto";
 
 package helloworld;
 
 // The greeting service definition.
-service Greeter {
+service GreeterService {
     // Sends a greeting
     rpc SayHello (HelloRequest) returns (HelloReply) {}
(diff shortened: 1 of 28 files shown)


sbt-plugin/src/sbt-test/gen-scala-server/01-gen-basic-server/src/main/protobuf/helloworld.proto
```diff
@@ -1,21 +1,21 @@
 syntax = "proto3";
 
 option java_multiple_files = true;
-option java_package = "io.akka.grpc";
+option java_package = "io.grpc.examples.helloworld";
```
johanandren: In Akka we always put the generated proto sources in a `akka.bbb.ccc.proto` should we recommend that here as well (but a `.grpc` ) to discern handwritten classes from generated ones?

ktoso: We can recommend that, but not religiously :)

raboof: so something like `example.myapp.helloworld.proto` and `example.myapp.echo.proto`?

johanandren: I'd say `example.myapp.helloworld.grpc`, doing that in samples isn't really religiously enforcing anything, is it?

ktoso: I dont mind, as long as wording mentions that packages are a bit up to personal preference, but that's a default you can stick to

PR #187, opened 2018-04-26
Motivate/explain when to use gRPC a bit better

raboof:


docs/src/main/paradox/index.md
```diff
@@ -1,10 +1,14 @@
 # Akka gRPC
 
-Support for building streaming gRPC servers and clients on top
-of Akka Streams.
+[gRPC](https://grpc.io) is a transport mechanism for request/response
+and non-persistent streaming use cases. Use it for:
+
+* connections between internal services
+* connecting to external services, even ones written in other languages.
```
ktoso: external services that expose an gRPC API

docs/src/main/paradox/index.md
```diff
@@ -19,7 +23,15 @@ Based on a protobuf service definition, akka-grpc can generate:
...
 * On the client side, a client for the API.
 
+### gRPC vs REST
+
+* Where REST is more flexible about transport and encoding, gRPC standardizes on HTTP/2 and Protobuf.
+* Where REST can be either schemaless or use a 3rd-party schema, gRPC always declares the service and messages in a Protobuf schema definition.
+
+### gRPC vs SOAP
 
+* Where SOAP is more flexible about transport and encoding, gRPC standardizes on HTTP/2 and Protobuf.
```
ktoso: is it more flexible really?

docs/src/main/paradox/index.md
```diff
@@ -19,7 +23,15 @@ Based on a protobuf service definition, akka-grpc can generate:
...
+* Where in SOAP protocols are often set in stone once defined, Protobuf is explicitly intended to support schema evolution.
```
johanandren: Second time someone reminds me of SOAP today. Yuck!

"defined, and may require a new path for every version of the service, " ?

ktoso: Been years since I used SOAP, was that really the case?

docs/src/main/paradox/index.md
```diff
@@ -19,7 +23,15 @@ Based on a protobuf service definition, akka-grpc can generate:
 * On the server side, code to create an Akka HTTP route based on your implementation of the API
 * On the client side, a client for the API.
 
+### gRPC vs REST
+
+* Where REST is more flexible about transport and encoding, gRPC standardizes on HTTP/2 and Protobuf.
```
johanandren: Huh, this I never heard before, I thought it was very tightly coupled with the GET, POST, UPDATE, DELETE methods of HTTP?

docs/src/main/paradox/index.md
```diff
@@ -19,7 +23,15 @@ Based on a protobuf service definition, akka-grpc can generate:
 * On the server side, code to create an Akka HTTP route based on your implementation of the API
 * On the client side, a client for the API.
 
+### gRPC vs REST
```
ktoso: Some discussion that it is "synchronous" in the sense that other side needs to be online when we make the call, but non-blocking server/client implementations lead to resource efficient hardware utilisation?

docs/src/main/paradox/index.md
raboof: That's true, perhaps let's just say `Where REST is more flexible about encoding`

docs/src/main/paradox/index.md
raboof: Heh, should we avoid mentioning it at all? Good addition.

docs/src/main/paradox/index.md
raboof: That would be true for both gRPC and REST, right?

docs/src/main/paradox/index.md
johanandren: I also "know" that, even heard someone claim they did it over alternative transport in financial industry

johanandren: I think it makes sense, because I think this is a pro of gRPC compared to SOAP (otoh pretty much anything is a pro compared to SOAP ;) )

docs/src/main/paradox/index.md
ktoso: Yes, as seen in the ongoing debate though this seems to be an area of confusion so perhaps worth calling out all these are synchronous, more or less decoupled, yet nonblocking implementations?

johanandren: Maybe that goes in a "when not to use it" section? Where we can hint towards message queues etc?

raboof:
Perhaps we should also contract gRPC with message buses here? 

* When using gRPC both sides need to be up, while when using a message bus only the bus needs to be up, the consuming side does not need to be available.
* gRPC supports streaming in either direction, when using a message bus this is decoupled.

docs/src/main/paradox/index.md
raboof: Not about encoding I guess, transport kinda (but I guess that might not be widely used https://en.wikipedia.org/wiki/SOAP#Transport_methods )

docs/src/main/paradox/index.md
```diff
@@ -19,7 +23,36 @@ Based on a protobuf service definition, akka-grpc can generate:
...
+
+* Where REST is more flexible about encoding, gRPC standardizes Protobuf.
```
johanandren: "standardizes on"

raboof: woops

PR #192, opened 2018-05-02
Simpler sbt config

### LINKED ISSUE #190: Easier configuration for what to generate
Right now we can configure:
* akkaGrpcCodeGenerators (`Seq[GeneratorAndSettings]`, default: ScalaBoth)
* akkaGrpcModelGenerators (`Seq[Target]`, default scalaPB)

Perhaps we should have one `akkaGrpcTargetLanguage` (Java/Scala) to set the default. We could simplify the rest - for the `CodeGenerators` we currently have `Server`, `Client`, and `Both`, so for now that doesn't need to be a `Seq`, but perhaps in the future we'd want to add more specific generators e.g. for Play?

Related to #155

johanandren:
Fixes #190


johanandren:
I didn't quite understand the interop-test setup with the magic stuff, so made it pass but maybe it is meant to actually use the plugin (which it doesn't now)

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -20,34 +21,45 @@ object AkkaGrpcPlugin extends AutoPlugin {
   override def requires = ProtocPlugin
 
   trait Keys { _: autoImport.type =>
+
+    object AkkaGrpc {
+      sealed trait TargetStub
```
raboof: I'm not sure I like the `stub` term, it seems to be a bit overloaded.. perhaps just `Target`?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -20,34 +21,45 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+      case object Client extends TargetStub
+      case object Server extends TargetStub
+
+      sealed trait TargetLanguage
+      case object Scala extends TargetLanguage
+      case object Java extends TargetLanguage
+    }
+
+    val akkaGrpcTargetLanguages = settingKey[Seq[AkkaGrpc.TargetLanguage]]("Which languages to generate service and model classes for (AkkaGrpc.Scala, AkkaGrpc.Java)")
+    val akkaGrpcTargetStubs = settingKey[Seq[AkkaGrpc.TargetStub]]("Which of the stubs to generate (AkkaGrpc.Server, AkkaGrpc.Client)")
```
raboof: (here as well)

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -20,34 +21,45 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
     val akkaGrpcCodeGeneratorSettings = settingKey[Seq[String]]("Settings to pass to the code generator")
-    val akkaGrpcCodeGenerators = settingKey[Seq[GeneratorAndSettings]]("The configured source generator")
-    val akkaGrpcModelGenerators = settingKey[Seq[Target]]("The configured source generator for model classes")
   }
 
-  object autoImport extends Keys {
-    final case class GeneratorAndSettings(generator: CodeGenerator, settings: Seq[String] = Nil)
-  }
+  object autoImport extends Keys
```
raboof: Nice to lose the `GeneratorAndSettings`, that was a bit weird

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -20,34 +21,45 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
 
   def configSettings(config: Configuration): Seq[Setting[_]] =
     inConfig(config)(Seq(
-      akkaGrpcModelGenerators := Seq[Target]((JvmGenerator("scala", ScalaPbCodeGenerator), akkaGrpcCodeGeneratorSettings.value) -> sourceManaged.value),
-
       unmanagedResourceDirectories ++= (resourceDirectories in PB.recompile).value,
       watchSources in Defaults.ConfigGlobal ++= (sources in PB.recompile).value,
 
       // configure the proto gen automatically by adding our codegen:
+      // FIXME: actually specifying separate Compile and Test target stub and languages does not work
```
raboof: Seems fine, perhaps let's add a ticket making it explicit that we currently don't support this?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,53 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+    // TODO no way to provide per language + stub settings/target dirs - not sure if needed
+    generators.map { case (generator, settings) => protocbridge.Target(generator, targetPath, settings) }
+  }
+
+  // creates a seq of generator and per generator settings
+  private def generatorsFor(stubs: Seq[AkkaGrpc.TargetStub], languages: Seq[AkkaGrpc.TargetLanguage], defaultSettings: Seq[String]): Seq[(protocbridge.Generator, Seq[String])] = {
+    // these two are the model/message (protoc) generators
+    def ScalaGenerator: (protocbridge.Generator, Seq[String]) = (protocbridge.JvmGenerator("scala", ScalaPbCodeGenerator), defaultSettings)
+    // we have a default flat_package, but that doesn't play with the java generator (it fails)
+    def JavaGenerator: (protocbridge.Generator, Seq[String]) = (PB.gens.java, defaultSettings.filterNot(_ == "flat_package"))
```
raboof: I think we should filter this 'the other way around', and only pass through options that the Java generator supports. I think we do that somewhere but can't find it right now...

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,53 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+      (protocbridge.JvmGenerator(codeGenerator.name, adapter), defaultSettings)
+    }
+
+    (stubs match {
+      case Seq(_, _) =>
+        languages match {
+          case Seq(_, _) => Seq(ScalaGenerator, toGenerator(ScalaBothCodeGenerator), JavaGenerator, toGenerator(JavaBothCodeGenerator))
+          case Seq(AkkaGrpc.Scala) => Seq(ScalaGenerator, toGenerator(ScalaBothCodeGenerator))
+          case Seq(AkkaGrpc.Java) => Seq(JavaGenerator, toGenerator(JavaBothCodeGenerator))
+        }
```
raboof: The `Seq(_, _)` is a bit strange.. but I can't come up with something succinct right now either ;).

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,53 @@ object AkkaGrpcPlugin extends AutoPlugin {
           unmanagedResources := { Defaults.collectFiles(unmanagedResourceDirectories, includeFilter, excludeFilter).value },
           resources := managedResources.value ++ unmanagedResources.value)))
 
-  def adaptAkkaGenerator(targetPath: File)(generatorAndSettings: GeneratorAndSettings): Target = {
-    val adapted = new ProtocBridgeSbtPluginCodeGenerator(generatorAndSettings.generator)
-    val generator = JvmGenerator(generatorAndSettings.generator.name, adapted)
-    (generator, generatorAndSettings.settings) -> targetPath
+  private def targetsFor(targetPath: File, defaultSettings: Seq[String], stubs: Seq[AkkaGrpc.TargetStub], languages: Seq[AkkaGrpc.TargetLanguage]): Seq[protocbridge.Target] = {
+    val generators = generatorsFor(stubs, languages, defaultSettings)
+    // TODO no way to provide per language + stub settings/target dirs - not sure if needed
```
raboof: Seems OK, users who need that kind of control could probably populate `PB.targets` directly themselves

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,53 @@ object AkkaGrpcPlugin extends AutoPlugin {
           unmanagedResources := { Defaults.collectFiles(unmanagedResourceDirectories, includeFilter, excludeFilter).value },
           resources := managedResources.value ++ unmanagedResources.value)))
 
-  def adaptAkkaGenerator(targetPath: File)(generatorAndSettings: GeneratorAndSettings): Target = {
-    val adapted = new ProtocBridgeSbtPluginCodeGenerator(generatorAndSettings.generator)
-    val generator = JvmGenerator(generatorAndSettings.generator.name, adapted)
-    (generator, generatorAndSettings.settings) -> targetPath
+  private def targetsFor(targetPath: File, defaultSettings: Seq[String], stubs: Seq[AkkaGrpc.TargetStub], languages: Seq[AkkaGrpc.TargetLanguage]): Seq[protocbridge.Target] = {
```
raboof: just `settings` rather than `defaultSettings` ;)

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,53 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+          case Seq(AkkaGrpc.Scala) => Seq(ScalaGenerator, toGenerator(ScalaServerCodeGenerator))
+          case Seq(AkkaGrpc.Java) => Seq(JavaGenerator, toGenerator(JavaServerCodeGenerator))
+        }
+    })
+  }
+
+  /**
+   * Adapts existing [[akka.grpc.gen.CodeGenerator]] into the protocbridge required type
+   */
+  private[akka] class ProtocBridgeSbtPluginCodeGenerator(impl: akka.grpc.gen.CodeGenerator) extends protocbridge.ProtocCodeGenerator {
```
raboof: Nice cleanup :+1:

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/build.sbt
```diff
@@ -0,0 +1,7 @@
+
+enablePlugins(JavaAgent)
+enablePlugins(AkkaGrpcPlugin)
+
+javaAgents += "org.mortbay.jetty.alpn" % "jetty-alpn-agent" % "2.0.6" % "runtime"
```
raboof: 2.0.7 is already out

raboof:
@johanandren interop-tests is not meant to actually run the plugin, we have the 'scripted' tests for that (but those are really slow, which is why the interop-tests are still good to have)

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
johanandren: I wanted to avoid just `Target` alone since `Target` is the end result of this exercise.  
`GeneratedSources`? (a bit too narrow perhaps, since that also includes the model which isn't configurable)

docs/src/main/paradox/sbt.md
```diff
@@ -4,39 +4,29 @@ To get started with Akka gRPC read the [client](client.md) or [server](server.md
...
-import akka.grpc.gen.javadsl.JavaClientCodeGenerator
-akkaGrpcCodeGenerators := Seq(GeneratorAndSettings(JavaClientCodeGenerator, akkaGrpcCodeGeneratorSettings.value))
-
-// only server
-import akka.grpc.gen.javadsl.JavaServerCodeGenerator
-akkaGrpcCodeGenerators := Seq(GeneratorAndSettings(JavaServerCodeGenerator, akkaGrpcCodeGeneratorSettings.value))
+akkaGrpcTargetLanguages := Seq(AkkaGrpc.Java)
 
+// Both Java and Scala
+akkaGrpcTargetLanguages := Seq(AkkaGrpc.Java, AkkaGrpc.Scala)
```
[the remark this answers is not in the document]
2m: +1 What is more, do you know if it is possible to control the package name of the generated gRPC model classes from the outside? Because after turning off the `flat_package` we get ourselves in a similar situation where Play API is, where everything one cryptic package name deeper is scaladsl.

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/build.sbt
johanandren: copy-pasta, I can upgrade across the project if it matters

docs/src/main/paradox/sbt.md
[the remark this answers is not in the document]
johanandren: I auto-disable it for Java anyways, (because it made the Java generator fail with an error) so I think that problem is solved. But let me verify.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaCompositeCodeGenerator.scala
```diff
@@ -5,8 +5,9 @@
 package akka.grpc.sbt
 
 import akka.grpc.gen.CompositeCodeGenerator
+import akka.grpc.sbt.AkkaGrpcPlugin.ProtocBridgeSbtPluginCodeGenerator
 
-/** An easy to use accessor to be used reflectively in the main sbt project */
+/** An easy to use accessor to be used reflectively in the main sbt project, not really used except for that */
```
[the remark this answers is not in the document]
johanandren: :+1:

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
2m: IMO  `stub` is only extensively used in the gRPC context therefore it felt natural as the common name of the client and server. Only having `Target` overloads it as we also have `TargetLanguage`.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
2m: The `case Seq(_, _) =>` could be moved from the first to the last case and changed to `case _ =>` but I am not sure if that is better either.

docs/src/main/paradox/sbt.md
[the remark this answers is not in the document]
johanandren: I had it wrong, I'll add a comment.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
raboof: right, but gRPC uses 'stub' mainly to describe the client AFAICS (https://grpc.io/docs/guides/concepts.html), it might in fact be confusing to also use it for other generated code?

I agree that just `Target` isn't very nice either though ;). `GeneratedSources` might be fine, and then in the setting comment explain it is 'which sources to generate in addition to the gRPC protobuf model classes'?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -69,9 +81,52 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+    val generators = generatorsFor(stubs, languages, settings)
+    generators.map { case (generator, generatorSettings) => protocbridge.Target(generator, targetPath, generatorSettings) }
+  }
+
+  // creates a seq of generator and per generator settings
+  private def generatorsFor(stubs: Seq[AkkaGrpc.GeneratedSource], languages: Seq[AkkaGrpc.Language], defaultSettings: Seq[String]): Seq[(protocbridge.Generator, Seq[String])] = {
+    // these two are the model/message (protoc) generators
+    def ScalaGenerator: (protocbridge.Generator, Seq[String]) = (protocbridge.JvmGenerator("scala", ScalaPbCodeGenerator), defaultSettings)
+    // we have a default flat_package, but that doesn't play with the java generator (it fails)
+    def JavaGenerator: (protocbridge.Generator, Seq[String]) = (PB.gens.java, defaultSettings.filterNot(_ == "flat_package"))
```
raboof: If the Java generator crashes on parameters it doesn't understand I think we should whitelist only the parameters we know it understands, rather than blacklisting the ones we know of that it doesn't understand. We do something similar for the scalapb plugin in `ScalaCodeGenerator`: that also crashes on parameters it doesn't expect.

OK for now though, it's not the point of this PR after all ;).

Issue #67, opened 2018-03-08
Move io.grpc.Channel logic into TestServiceServiceClient


raboof:
To abstract this away for the end user. Does that make sense @renatocaval ?

2m:
What is the plan with the Channel logic? I see that the scala client builds the channel itself (after #173). However the java client still requires the channel to be passed in to the factory method.

I found this while working on Alpakka Google PubSub gRPC connector. Usage example here: https://github.com/2m/alpakka/blob/CremboC-pub-sub-grpc/google-cloud-pub-sub-grpc/src/main/scala/akka/stream/alpakka/googlecloud/pubsub/grpc/impl/GrpcApi.scala#L22-L30 Here I create a scala client just to extract the channel for the java client as a workaround. :)

raboof:
Ah sorry it seems I forgot to update the Java API side.

The current channel logic is Netty-specific, so it was neater to pass in settings and make the channel creating an implementation detail. I think this interface makes sense, but in the future I could imagine multiple clients sharing an Akka HTTP HTTP/2 'connection pool' somehow, but it seems premature to anticipate that in the API.

raboof:
We now removed the `io.grpc.Channel`-specific things out of the akka-grpc client API, 'manually' managing the lifecycle of the client. As discussed in the last comment above this is not the 'final form' but closes this issue for now.

PR #198, opened 2018-05-03
Java client creation API aligned with Scala

johanandren:
Refs #67 

Additionally:
Cleanup of things that need not be templates and moved those into runtime.


raboof:
:tada:

PR #200, opened 2018-05-04
Publish on tag

raboof:
Does not ping tech hub yet

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -121,7 +121,6 @@ lazy val interopTests = Project(
   )
   .dependsOn(runtime)
   .enablePlugins(akka.grpc.ReflectiveCodeGen)
-  .enablePlugins(akka.grpc.NoPublish)
   // needed to be able to override the PB.generate task reliably
   .disablePlugins(ProtocPlugin)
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -121,7 +121,6 @@ lazy val interopTests = Project(
   )
   .dependsOn(runtime)
   .enablePlugins(akka.grpc.ReflectiveCodeGen)
-  .enablePlugins(akka.grpc.NoPublish)
```
johanandren: Why do we want to publish the interop tests?

raboof: They are used by the scripted tests. Might be more elegant to only `publishLocal` them, but can't really hurt either...

project/Publish.scala
```diff
@@ -0,0 +1,32 @@
...
+object NoPublish extends AutoPlugin {
+  override def requires = plugins.JvmPlugin
+
+  override def projectSettings = Seq(
+    publishArtifact := false,
+    publish := {},
+    publishLocal := {},
+    skip in publish := true,
+  )
+}
```
2m: All of this accumulated throughout the years. `skip in publish := true` should be the only one required to do the job. Lets try to adjust the folklore to only that setting. :)

Issue #188, opened 2018-04-26
Deadline propagation


raboof:
gRPC has a way to specify a deadline for a call.

* Client: introduce a way to set call deadlines
* Server: fail the stream when the deadline is missed

patriknw:
I added deadline on client side in https://github.com/akka/akka-grpc/pull/245
Server side still remaining, but probably not for MVP.

hollinwilkins:
Is there a way to specify a deadline outside of the GrpcClientSettings, on a per-call basis?

PR #201, opened 2018-05-07
Client power user API

### LINKED ISSUE #191: Client-side power API
The current generated client has an API suitable for message streams, but does not provide access to underlying protocol details.

For some use cases, such as adding per-call authentication principal information or reading custom response headers, we should have a 'power API'.  Likely we'll have a method on the 'normal' client to get an instance of the 'powerful' client, which would also have methods for each call, but those would then return an object on which further specifics can be specified before the call is issued, something like:

```
val myServiceClient: MyServiceClient
val s: Source[Foo] = myServiceClient.getFoos(query)
val a: AdvancedMyServiceClient = myServiceClient.advanced
val (s: Source[Foo], rh: Seq[HttpHeader]) = a.getFoos.addRequestHeader("Authentication", "Complicated").exposeResponseHeaders.execute()
```

(all naming and other specifics are still up for debate ;) )

johanandren:
Fixes #191

Went for 
```
client.method()
  .bladi()
  .dadi()
  .execute(param)
```

instead of 
```
client.advanced.method(param)
  .bladi()
  .dadi()
  .execute()
```


johanandren:
No usage samples yet, but the "normal" calls are now expressed through the new RequestBuilder version of the method, and interop-tests passing.

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
```diff
@@ -0,0 +1,147 @@
...
+ * Request builder for requests giving providing per call metadata capabilities
+ *
+ * Not for user extension
+ */
+@DoNotInherit
+sealed trait RequestBuilder[Req, Res] {
+
+  def addMetadata(key: String, value: String): RequestBuilder[Req, Res]
+  def withDeadline(deadline: FiniteDuration): RequestBuilder[Req, Res]
+  def execute(req: Req): Res
```
renatocaval: What about `invoke` instead of `execute`?

renatocaval: Me neither. ;-) Lagom uses `invoke` hence my slight preference. 

Btw, that API style means that we always have a lifted API representation instead of a switch giving access to the lifted/advanced one. 
I thought we wanted to provide a non-lift, straight, variation for the most common cases.

raboof:
Right! Moving the parameter to the end of the chain makes a lot of sense: this way a library user can even reuse the `RequestBuilder` for multiple requests that have the same characteristics. That seems like a cool advantage and I don't see any disadvantage.

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
johanandren: Yeah. Not sure that it is always available rather than hidden behind a switch is a problem though. The old non-lifted API is still there (and should be the default)

johanandren: Aligning with Lagom seems like sensible motivation for `invoke` :)

renatocaval: Ah ok, now I see it. Because in gRPC there is always one param, we can overload the method with a paramless variation that gives access to the lifted version. Neat! Much better than a switch.

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -25,56 +26,30 @@ final class @{service.name}Client(settings: GrpcClientSettings)(implicit mat: Ma
...
-        ClientCalls.asyncBidiStreamingCall(
-          channel.newCall(@{method.name}Descriptor, options),
-          responseObserver)
-      }
-    in.via(flow)
+    }
   }
+
+  /**
+   * For access to method metadata use the parameter less version of ${method.name}
```
raboof: ('parameterless')

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
```diff
@@ -0,0 +1,147 @@
...
+  options: CallOptions,
+  delegatedExecute: (Req, CallOptions) => Ret) extends RequestBuilder[Req, Ret]() {
+
+  def addMetadata(key: String, value: String): RequestBuilderImpl[Req, Res, Ret] = {
+    // FIXME support values other than string values?
+    copy(options = options.withOption(CallOptions.Key.of(key, ""), value))
+  }
+
+  def withDeadline(deadline: FiniteDuration): RequestBuilderImpl[Req, Res, Ret] = {
+    copy(options = options.withDeadline(Deadline.after(deadline.length, deadline.unit)))
```
raboof: Nice, we don't support this at the server side yet (#188) but since this is :+1:

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * INTERNAL API
+ */
+@InternalApi
+final case class RequestBuilderImpl[Req, Res, Ret](
+  options: CallOptions,
+  delegatedExecute: (Req, CallOptions) => Ret) extends RequestBuilder[Req, Ret]() {
+
+  def addMetadata(key: String, value: String): RequestBuilderImpl[Req, Res, Ret] = {
+    // FIXME support values other than string values?
```
raboof: do 'options' correspond to 'headers' or is there more indirection there?

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -25,56 +26,30 @@ final class @{service.name}Client(settings: GrpcClientSettings)(implicit mat: Ma
...
-        ClientCalls.asyncClientStreamingCall(
-          channel.newCall(@{method.name}Descriptor, options),
-          responseObserver)
+  /**
+   * Lower level version of the method, prefer @{method.name}(@method.parameterType) if possible.
+   */
+  def @{method.name}(): RequestBuilder[@method.parameterType, @method.returnType] = {
+    val descriptor = @{method.name}Descriptor
+    @if(method.methodType == akka.grpc.gen.Unary) {
+      RequestBuilderImpl.unary(descriptor, channel, options)
```
raboof: Do we want to organize the template so that the indentation of the generated code makes sense? No real preference here.

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
johanandren: I think there is a bit more indirection, there is also some marshalling infra in the grpc library around this. Not sure if we care about that though, maybe strings are good enough for now, and also avoids tying us down to something client specific. 

It's annotated as an experimental API, but so is most of these CallOptions.

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
johanandren: Ah, I thought I did follow that, since it is how it was, but maybe having the templates readable is more important for maintainability than not having too many indentation levels in the generated sources?

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
johanandren: Yeah, just picked this randomly to have some functionality to provide through the request builder.

johanandren:
One thing I couldn't figure out was a reasonable way to deal with response metadata, for single response calls it may be relatively straight forward API wise (`Future[(Resp, Metadata)]`) but for streaming data it just doesn't make sense to put it inside of the stream, as you'd potentially want to act on it before consuming the response stream.

Any thoughts around that?

ktoso:
I had proposed a wrapper type for that I believe; `GrpcResponse(stream or future).with(all the stuff)`, WDYT?

johanandren:
Thinking about this some more, maybe the primary use case will be to _set_ metadata on the request, not deal with response metadata, so perhaps there should be the current "simple" `invoke(req)` and then a more complete but also more messy `invokeWithAnnoyingReturnType(req)`?

renatocaval:
I think a better way of doing that is to have a method that changes the return type. That's also how it was done in Lagom.

For instance:
```
val res: Future[Res] =
  client.method()
    .addMetadata(key, value)
    .invoke(param)

val res: Future[GrpcResponse[Res]] =
  client.method()
    .addMetadata(key, value)
    .withResponseHeaders
    .invoke(param)
```

This requires two impl of `RequestBuilder` and the method `withResponseHeaders` works as a one direction switch.

johanandren:
I don't see why a stateful/extra type is better than an additional `invoke` method with different return type. Am I missing something? Just to align with Lagom?

TimMoore:
Yeah, I think it is pretty useful to be able to write decorators that can manipulate either/both the request and response metadata and payloads. I don't see a way with this API to be able to write a decorating function that can intercept the response without using internal APIs.

johanandren:
Hmm, I still don't get it, why would a special type with transformations be better than just having a `Future[ResponseAndHeadersType]`/`CompletionStage<ResponseAndHeadersType>` which you can transform in exactly the same way as all your other async thingies and that you already know the API of?

johanandren:
I think all those cases you describe should be covered by having two methods on the builder: the one that currently exists and then one that returns a `Future[ResponseWithHeaders]`. Just calling `map` on that then allows you to transform the response to any type you want while also looking at the response headers.

renatocaval:
Sure, but does that mean that in case of streaming we will get a `Future[StreamResponseWithHeaders[T, R]]` from each I could read the headers and the `Source`?

This sounds good and it's worth experiment. I has some implications though.

 If the general API returns `Source[T, R]`, the power user API will have to return `Future[StreamResponseWithHeaders[T, R]]`.

It also means that there is effectively two calls, one to get the `Future` and one when we effectively materialize the `Source`, but I guess we can't avoid it. Unless we return `Source[(Headers, T), R]` which has an obvious overhead penalty.

And then we may have a another problem, for the moment mostly theoretical. This two calls may create a situation where the metadata differs between them.

renatocaval:
For completeness and clarity. 

The implications I mentioned above (two calls) also holds for the `handleResponseHeader` API. I didn't mean to say that using `Future[ResponseWithHeaders]` was forcing us into that situation.

renatocaval:
Just a crazy thought. 

The only way to associate headers to a stream and make sure that they relate (make part of the same call) is `Source[(Headers, T), R]`. We also want to keep the semantics of `Source` and be able to re-run it (make the call again) whenever we want.

So, maybe...

```scala
sealed trait Headers
case object EmptyHeaders extends Headers
case class NonEmptyHeaders extends Headers
```
Than is up to the implementers to decide when to send the headers. On the first element, on each element, etc.

Didn't thought thoroughly on the implications of this idea. Maybe stupid, maybe crazy, maybe good. I let this go through your brains first. :-)

raboof:
I'm not entirely convinced we want to keep the semantics of Source and be able to re-run it (make the call again) whenever we want: I don't think we do that for `Source`-backed entities in `HttpRequest`s either.

`Source[(Headers, T), R]` suggests each `T` may have headers, which is not the case (as it's the response that has (leading and trailing) headers, not each element in the response individually) - so from that perspective returning the future containing headers and source seems to make more sense.

renatocaval:
> suggests each `T` may have headers
You are totally right on that. Not a good idea.

johanandren:
Internally the headers will arrive when the flow is materialized, and then there is a `Future[Trailers]` since those won't arrive until the stream completes. Haven't thought through how to provide this to the user yet, but the underlying stage I just wrote can at least be materialized multiple times and each materialization leads to one call - one set of metadata.

ignasi35:
If a `Source` can be materialised several times causing a new call each time, then could headers and trailers be the materialised value? (I don't even know if that makes sense)

`Source[ T, ( Future[Headers], Future[Trailers] ) ]`

johanandren:
Yeah, likely. But I think a specific type capturing it would be better than a tuple. The alternative would be to side effect from `mapMatValue` to complete some promises/futures for then, that way it could be something like `StreamingGrpcResponse(stream, futureHeaders, futureTrailers)`.

ignasi35:
I find counterintuitive that materializing the `Source` several times causes new calls without using `client.method()...invoke(param)`.

IIRC Lagom's client produces a `Source` that's only meant to be materialized once and new calls only happen on new `invoke`. Also, in the 4 possible combinations (unary-basic, streamed-basic, unary-powerUser, stream-powerUser) I think this would be the only combination where materializing twice creates two calls, right? Or is a `Source` obtained using the basic (non-powerUSer) API also behaving like this?

Maybe renaming akka-grpc's `invoke` to `callFactory` or something that's not `invoke` could help make it more obvious what's actually happening.

johanandren:
Both of the source variation could be materialized many times. If that is an issue we can protect against that and fail. I don't have a gut feeling about it being surprising or not yet, I'll see when the actual user API is in place (which is what I am currently working on).

plugin-tester-scala/src/main/scala/example/myapp/helloworld/LiftedGreeterClient.scala
```diff
@@ -0,0 +1,99 @@
...
+
+    implicit val sys = ActorSystem()
+    implicit val mat = ActorMaterializer()
+    implicit val ec = sys.dispatcher
+
+    val client = new GreeterServiceClient(GrpcClientSettings(
+      "127.0.0.1",
+      8080,
+      overrideAuthority = Some("foo.test.google.fr"),
+      None,
```
raboof: perhaps name this parameter as well?

plugin-tester-scala/src/main/scala/example/myapp/helloworld/GreeterClient.scala
```diff
@@ -56,7 +66,7 @@ object GreeterClient {
       val responseStream = client.itKeepsReplying(HelloRequest("Alice"))
       val done: Future[Done] =
         responseStream.runForeach(reply => println(s"got streaming reply: ${reply.message}"))
-      Await.ready(done, 1.minute)
+      Await.ready(done, 1.minute) // just to keep sample simple - dont do Await.ready in actual code
```
raboof: Perhaps we should avoid it here as well then - but let's keep that for another PR

runtime/src/main/scala/akka/grpc/GrpcClientResponse.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Optional
+import java.util.concurrent.CompletionStage
+
+import akka.annotation.DoNotInherit
+import io.grpc.Metadata
+
+import scala.concurrent.Future
+
+// FIXME should we provide our own immutable/thread safe response metadata abstraction?
```
raboof: Indeed the upstream `Metadata` is not too nice, might be good to at least wrap it in some immutable interface?.

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
```diff
@@ -0,0 +1,314 @@
...
+ * INTERNAL API
+ */
+@InternalApi
+trait OptionsModifications[T <: OptionsModifications[T]] {
+  def options: CallOptions
+  def updated(options: CallOptions): T
+
+  def addMetadata(key: String, value: String): T = {
+    // FIXME Key is instance equal to allow for replacing of the same key but still also allowing multiple
+    // values with the same name-key, not sure if it is important we support that
```
raboof: Let's keep an eye out if we ever see that (multiple `role`s?).

If so another way to do it might be to have this method overwrite keys, and introduce a `def addMetadata(key: String, value: Seq[String])` for multi-valued metadata

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+trait OptionsModifications[T <: OptionsModifications[T]] {
+  def options: CallOptions
+  def updated(options: CallOptions): T
+
+  def addMetadata(key: String, value: String): T = {
```
raboof: Since later the `GrpcResponseMetadata` contains headers, perhaps `addHeader` would make more sense here?

runtime/src/main/scala/akka/grpc/internal/UnaryCallAdapter.scala
```diff
@@ -0,0 +1,101 @@
...
+  def future: Future[Res] = responsePromise.future
+  def cs: CompletionStage[Res] = future.toJava
+}
+
+/**
+ * gRPC Netty based client listener transforming callbacks into a future response
+ *
+ * INTERNAL API
+ */
+// needs to be a separate class because of CompletionStage error handling not bubling
```
raboof: bubbling ;)

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
```diff
@@ -0,0 +1,83 @@
...
+   * Invoke the gRPC method with the additional metadata added
+   *
+   * FIXME for streaming response this doesn't really make sense, left it to keep parity with single response but maybe it should go
+   */
+  def invoke(request: Req): Source[Res, NotUsed]
+
+  /**
+   * Invoke the gRPC method with the additional metadata added and provide access to response metadata
+   */
+  def invokeWithMetadata(request: Req): Source[Res, Future[GrpcResponseMetadata]]
```
raboof: Encoding the metadata in the materialized value looks good.

It seems a 'clever' solution which sometimes backfires, but in this case it looks like it works out elegantly.

I agree since the metadata is in the materialized value the need for a separate `invokeWithMetadata` kind of disappears... unless perhaps there's opportunity for more optimization in `invoke` later?

runtime/src/main/scala/akka/grpc/internal/AkkaGrpcGraphStage.scala
```diff
@@ -4,48 +4,188 @@
...
+ *  Flow in       ------ I------>  |                | ------- O ------->  Flow out
+ *                                 |  Netty client  |
+ *  upstream pull <-- onReady ---  |      call      | <-- request(1) ---  downstream pull
+ *                                 ------------------
+ *
+ *
+ * @param streamingResponse Do we expect a stream of responses or does more than 1 response mean a faulty server?
+ */
+@InternalApi
+private final class AkkaNettyGrpcClientGraphStage[I, O](
```
raboof: cool

runtime/src/main/scala/akka/grpc/scaladsl/RequestBuilder.scala
johanandren: Yeah, that may be a good argument for keeping the two separate methods

runtime/src/main/scala/akka/grpc/internal/UnaryCallAdapter.scala
johanandren: Less doubling in my bubbling!

plugin-tester-scala/src/main/scala/example/myapp/helloworld/LiftedGreeterClient.scala
johanandren: ups, keep missing out on the plugin-testers as they are not part of the aggregated build

runtime/src/main/scala/akka/grpc/GrpcClientResponse.scala
johanandren: Done!

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
johanandren: It turned out that what I thought was the same thing, the CallOptions.options and the metadata headers, was in fact not at all the same thing, so this is gone now.

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
johanandren: Yeap, I renamed when changing it into metadata. I realize I renamed it wrong though, changed it to `withHeader` as I thought it replaces, but I realize now that it should be additive and not replace, so `add` would be better.

johanandren:
Should be ready for merge now! :)

johanandren:
Ah, crap, I missed that it maybe failed validation, github refuses to show me what failed right now though.

raboof:
No worries, I think I fixed it on master

PR #224, opened 2018-06-01
No compilation warnings for generated Java client

### LINKED ISSUE #216: Fix warnings in the generated code
At least the generated Java code yields warnings about unsafe usage, we should make sure it does not cause warnings.

johanandren:
Fixes #216

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,10 @@ val commonSettings = Seq(
     "-language:_",
     "-encoding", "UTF-8"
   ),
+  javacOptions ++= List(
+    "-Xlint:unchecked",
+    "-Xlint:deprecation"
+  )
 )
 
 val akkaGrpcRuntimeName = "akka-grpc-runtime"
--- a/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
@@ -60,21 +60,21 @@ public final class @{service.name}Client implements @{service.name} {
 
     @for(method <- service.methods) {
       @if(method.methodType == akka.grpc.gen.Unary) {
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -11,6 +11,10 @@ val commonSettings = Seq(
     "-language:_",
     "-encoding", "UTF-8"
   ),
+  javacOptions ++= List(
+    "-Xlint:unchecked",
+    "-Xlint:deprecation"
```
raboof: Does this mean the warnings are still shown when the generated Java code is compiled in, say, a Maven project? If so then it might be worth seeing if we can avoid them entirely instead of suppressing them here? Could be a subsequent issue though.

johanandren: This enabled printing them so I could figure out what was wrong. No suppression going on!

raboof: Ah of course, excellent!

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -60,21 +60,21 @@ public final class @{service.name}Client implements @{service.name} {
...
+            @{method.name}RequestBuilder = new JavaServerStreamingRequestBuilder<>(
               @{method.name}Descriptor, "@fqName", channel, defaultOptions);
           } else if(method.methodType == akka.grpc.gen.BidiStreaming){
-            @{method.name}RequestBuilder = new JavaBidirectionalStreamingRequestBuilder(
+            @{method.name}RequestBuilder = new JavaBidirectionalStreamingRequestBuilder<>(
               @{method.name}Descriptor, "@fqName", channel, defaultOptions);
           }
         }
-      }
+      }G
```
raboof: Oops

johanandren: Dang

Issue #205, opened 2018-05-11
Document server side concurrency requirements


johanandren:
The service implementation on the server side needs to be thread safe as the same instance is used for all requests. We need to document this.

patriknw:
We should also document how to implement stateful services. Keeping state in actors is probably a good approach. Interaction with `Flow.ask` or ordinary `ask` is easiest, but other ways also possible.

PR #228, opened 2018-06-01
Sample of a stateful service added 

johanandren:
Refs #205

### DIFF
--- a/docs/src/main/paradox/server.md
+++ b/docs/src/main/paradox/server.md
@@ -141,6 +141,10 @@ That service can then be handled by an Akka HTTP server via the generated `Greet
 which is a @scala[partial ]function from `HttpRequest` to @scala[`Future[HttpResponse]`]@java[`CompletionStage<HttpResponse>`].
 @scala[The partial function should be made total before giving it to Akka HTTP by for example providing 404 as as default response].
 
+The server will run one instance of the implementation and that is then shared between requests,
+this mean that it must be thread safe. In the sample above there is no mutable state, for more about safely implementing
+servers with state see @link-ref-how-do-you-do-it?[stateful]
+
 A main program that starts a Akka HTTP server with the `GreeterService` looks like this:
 
 Scala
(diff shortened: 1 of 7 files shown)


docs/src/main/paradox/server.md
```diff
@@ -141,6 +141,10 @@ That service can then be handled by an Akka HTTP server via the generated `Greet
 which is a @scala[partial ]function from `HttpRequest` to @scala[`Future[HttpResponse]`]@java[`CompletionStage<HttpResponse>`].
 @scala[The partial function should be made total before giving it to Akka HTTP by for example providing 404 as as default response].
 
+The server will run one instance of the implementation and that is then shared between requests,
+this mean that it must be thread safe. In the sample above there is no mutable state, for more about safely implementing
+servers with state see @link-ref-how-do-you-do-it?[stateful]
```
raboof: heh I guess `[stateful](#Stateful-services)`? Not entirely sure.

plugin-tester-scala/src/main/scala/example/myapp/statefulhelloworld/GreeterServiceImpl.scala
```diff
@@ -0,0 +1,33 @@
...
+
+// #stateful-service
+class GreeterServiceImpl(system: ActorSystem) extends GreeterService {
+
+  val greeterActor = system.actorOf(GreeterActor.props("Hello"), "greeter")
+
+  def sayHello(in: HelloRequest): Future[HelloReply] = {
+    // timeout and execution context for ask
+    implicit val timeout: Timeout = 3.seconds
+    import system.dispatcher
```
raboof: I think we want to show `implicit val ec: ExecutionContext = system.dispatcher`?

plugin-tester-java/src/main/proto/stateful-helloworld.proto
```diff
@@ -0,0 +1,38 @@
...
+package statefulhelloworld;
+
+//#services
+// The greeting service definition.
+service GreeterService {
+    // Sends a greeting
+    rpc SayHello (HelloRequest) returns (HelloReply) {}
+
+    // changes the used greeting
+    rpc ChangeGreeting (ChangeRequest) returns (ChangeResponse) {}
```
raboof: Should we also show streaming?

plugin-tester-java/src/main/java/example/myapp/statefulhelloworld/GreeterServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+import static akka.pattern.PatternsCS.ask;
+
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+// #stateful-service
+public final class GreeterServiceImpl implements GreeterService {
+
+  private final ActorSystem system;
```
raboof: I guess `system` doesn't even need to be a field here - though perhaps good to show since real-life implementations might often need it?

plugin-tester-java/src/main/java/example/myapp/statefulhelloworld/GreeterServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+  public GreeterServiceImpl(ActorSystem system) {
+    this.system = system;
+    this.greeterActor = system.actorOf(GreeterActor.props("Hello"), "greeter");
+  }
+
+  public CompletionStage<HelloReply> sayHello(HelloRequest in) {
+    return ask(greeterActor, GreeterActor.GET_GREETING, timeout)
+        .thenApply(message ->
+          HelloReply.newBuilder()
+            .setMessage(((GreeterActor.Greeting) message).greeting)
```
raboof: Do we have a `mapTo` in scaladsl?

plugin-tester-java/src/main/java/example/myapp/statefulhelloworld/GreeterServiceImpl.java
johanandren: Yeah, actually I thought I'd need it for ask, but then CS-ask doesn't. Let's keep it anyway

plugin-tester-java/src/main/java/example/myapp/statefulhelloworld/GreeterServiceImpl.java
johanandren: `mapTo` is only for Scala as far as I could see (probably since it cannot be an extension method on `CS` like it is in Scala)

plugin-tester-java/src/main/proto/stateful-helloworld.proto
johanandren: I was thinking about that, but let's keep it simple like this to start with.

docs/src/main/paradox/server.md
johanandren: Yes, that did the trick. Thanks!

PR #230, opened 2018-06-07
fix broken certs in plugin-tester-java

patriknw:
* using same code as in plugin-tester-scala


patriknw:
> Can we also remove it from build.gradle and pom.xml?

I thought I did that. More places?

raboof:
Oh I guess I missed that then... let's merge when green!

patriknw:
some unrelated integration tests failing, I think we already have ticket for that

PR #237, opened 2018-06-15
[DO NOT MERGE] Allowing downstream projects to add its own generators

ignasi35:
By refactoring the sbt plugin downstream projects may use the machinery
setup here to add their own generators.

Do not merge.


raboof:
I refined this further into #246, WDYT?

raboof:
(seems superseded by #246, feel free to reopen when anything is missing)

Issue #176, opened 2018-04-13
Documentation/getting started improvements - Java/Maven


johanandren:
A bit messy, sorry, but what I noted when rewriting the notifications sample for Java:
- [x] no proto directory rendered in the docs “add .proto files to your projects `` directory”
- [x] minor and maybe just me - isn’t the “normal” way to put maven properties up top rather than the end?
- [x] not relevant for the long run, but needed to remember to publish to the maven repo rather than ivy
- [ ] maven plugin details missing
- [ ] maybe we shouldn’t be using someone else package-namespace in the sample? re “io.grpc.examples.helloworld” or is that pulled in from a jar?
- [x] missing hookin to the lifecycle - this was fixed already
- [x] generated sources end up in target/main/java/… which neither IntelliJ nor maven seem to understand out of the box - should be target/generated-sources/… - fixed already
- [x] m2e-lifecycle-mapping? - will be done as #171
- [ ] maybe we could skip the useTls/useTestCa boolean flags to simplify the starter examples (for Scala as well)
- [ ] wrong cert loaded “rootCA.key” but should be “ca.pem”
- [ ] alpn sample for maven in Akka HTTP docs did not work without some figuring out maven details

johanandren:
I think the relevant parts of this has been addressed now.

Issue #219, opened 2018-05-31
akka-grpc server scala with akka-grpc scala client tester should pass the custom_metadata integration test


raboof:
https://travis-ci.org/akka/akka-grpc/jobs/386240856
Looks like the `x-grpc-test-echo-trailing-bin` trailing header was missing

johanandren:
Was this fixed by #215 or did it happen just after that got merged?

raboof:
It looks like just after, though I haven't seen it since

raboof:
So far I've only seen it with the akka-grpc scala server and the akka-grpc scala or java client:

```
https://travis-ci.org/akka/akka-grpc/jobs/386589637
  akka-grpc server scala with akka-grpc scala client tester
https://travis-ci.org/akka/akka-grpc/jobs/386576573
  akka-grpc server scala with akka-grpc scala client tester
https://travis-ci.org/akka/akka-grpc/jobs/386589617
  akka-grpc server scala with akka-grpc scala client tester
https://travis-ci.org/akka/akka-grpc/jobs/386586816
  akka-grpc server scala with akka-grpc java client tester
```

That suggests we're doing something wrong in our server implementation, but let's keep an eye on this

johanandren:
👁

raboof:
Seems to fail fairly consistently (but not *all* the time) on travis, very rarely (but a couple of times) locally. 

Added some debug logging (branch `debugCustomMetadataServing`), this confirms the headers are added on the server side but are not presented to use from Netty on the client side.

Configured the test to run without TLS (branch `debugCustomMetadataServingWithoutTls`, with a snapshot akka-http version). Haven't seen it fail yet, but wireshark does show there appears to be padding missing from the final HEADERS frame. Might be something to look into monday.

raboof:
Looking at https://http2.github.io/http2-spec/#rfc.section.6.2 is seems not having padding is perfectly fine.

johanandren:
https://travis-ci.org/akka/akka-grpc/jobs/394514604

patriknw:
Seems to be failing rather frequently on travis.

```
Running test custom_metadata
[INFO] [06/21/2018 06:41:25.598] [pool-1-thread-1-ScalaTest-running-GrpcInteropSpec] [akka.actor.ActorSystemImpl(default)] Exception thrown, unbinding
org.scalatest.exceptions.TestFailedException
	at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:528)
[...]

PR #240, opened 2018-06-20
Documentation improvements

johanandren:
Covers the maven plugin, and some other cross-build-tool-fixes.


johanandren:
Test failure unrelated #219

johanandren:
Fixes #176

docs/src/main/paradox/maven.md
```diff
@@ -2,4 +2,95 @@
...
+The plugin can be configured to generate either Java or Scala classes, and then server and or client for the chosen language.
+By default both client and server in Java are generated.
+
+```xml
+<plugin>
+    <groupId>com.lightbend.akka.grpc</groupId>
+    <artifactId>akka-grpc-maven-plugin</artifactId>
+    <version>${akka.grpc.version}</version>
+    <configuration>
+      <language>JAVA</language> <!- or SCALA, must be uppercase -->
```
patriknw: can we fix that instead, and make it case insensitive?

docs/src/main/paradox/maven.md
```diff
@@ -2,4 +2,95 @@
...
+```
+
+## Loading proto files from artifacts
+
+TODO this is not supported yet See [#152](https://github.com/akka/akka-grpc/issues/152)
+
+
+## Starting your Akka gRPC server from Maven
+
+As the server requires a special Java agent for ALPN ([see Akka HTTP docs about HTTP2](https://doc.akka.io/docs/akka-http/current/server-side/http2.html#application-layer-protocol-negotiation-alpn-))
```
patriknw: HTTP/2

docs/src/main/paradox/sbt.md
```diff
@@ -57,4 +57,28 @@ that contain proto definitions to your build:
 ```scala
 libraryDependencies +=
   "com.example" %% "my-grpc-service" % "1.0.0" % "protobuf"
+```
+
+## Starting your Akka gRPC server from sbt
+
+As the server requires a special Java agent for ALPN ([see Akka HTTP docs about HTTP2](https://doc.akka.io/docs/akka-http/current/server-side/http2.html#application-layer-protocol-negotiation-alpn-))
```
patriknw: HTTP/2

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
```diff
@@ -28,7 +28,9 @@ class GenerateMojo @Inject() (project: MavenProject) extends AbstractMojo {
   var generateServer: Boolean = _
 
   override def execute(): Unit = {
-    val schemas = findProtoFiles(new File(protoPath))
+    val protoDirectory = new File(project.getBasedir, protoPath).getAbsoluteFile
+    if (!protoDirectory.exists()) throw new RuntimeException(s"Configured protoPath $protoDirectory does not exist")
```
patriknw: I'd put `[ ]` around `$protoDirectory` to make it more clear if it's empty string (and consistent with how we do it many other places)

docs/src/main/paradox/maven.md
```diff
@@ -2,4 +2,95 @@
...
+          <arguments>
+            <argument>-javaagent:${org.mortbay.jetty.alpn:jetty-alpn-agent:jar}</argument>
+            <argument>-classpath</argument>
+            <classpath />
+            <argument>com.example.MainClass</argument>
+          </arguments>
+        </configuration>
+     </plugin>
+  </plugins>
+```
```
patriknw: is this included in the plugin-tester-java project?

docs/src/main/paradox/maven.md
raboof: Agreed, #241

docs/src/main/paradox/server.md
```diff
@@ -164,25 +164,8 @@ akka.http.server.preview.enable-http2 = on
...
-Gradle
-:   ```
-./gradlew run
-```
-
-Maven
-:   ```
-mvn akka-grpc:generate compile exec:java -Dexec.mainClass=io.grpc.examples.helloworld.GreeterClient
-```
+See the detailed chapters on @ref[sbt](sbt.md#starting-your-akka-grpc-server-from-sbt), @ref[Gradle](gradle.md#starting-your-akka-grpc-server-from-gradle)
```
raboof: I agree adding this paragraph here.

I wonder if we should keep the 'simple case' sbt/gradle/maven snippets though, to make this section read more like a tutorial when you see it for the first time (while providing the links to dig deeper if you need to)?

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
raboof: :+1: - good idea to add the check!

plugin-tester-scala/project/plugins.sbt
```diff
@@ -1,3 +1,5 @@
+// #java-agent-plugin
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4")
+// #java-agent-plugin
```
raboof: :+1:

docs/src/main/paradox/maven.md
johanandren: Nop, I'll add it for completeness.

docs/src/main/paradox/server.md
johanandren: The reason I didn't do that is because there isn't really a simple case, not even for sbt where you still have to add stuff to both plugin.sbt and build.sbt before you can actually run anything, so that is three code blocks that needs to coexist with other code blocks for the other two build tools (which are also not simple steps for maven and I'm  guessing the same will be true for gradle?)

plugin-tester-scala/pom.xml
```diff
@@ -39,13 +54,50 @@
...
+          </arguments>
+        </configuration>
       </plugin>
     </plugins>
   </build>
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <grpc.version>1.12.0</grpc.version>
+    <akka.grpc.project.version>ce349193+20180621-1030</akka.grpc.project.version>
```
raboof: Should we remove this line? (when testing you can set it from the command line with `-Dakka.grpc.project.version=ce349193+20180621-1030`)

johanandren: With the comment up top I was thinking it is nicer to have it here as a placeholder, but sure, I'll remove it.

Issue #242, opened 2018-06-21
Protoc bridge and errors


johanandren:
protoc-bridge does not capture stderr from the proto process, this means that by luck you get errors from calling protoc shown in the terminal with sbt, it does however not end up in the actually logged error message which will/may be problematic when run from inside an IDE, and definitely for the maven plugin when run inside an IDE where the only error you will get is "java.lang.RuntimeException: protoc returned exit code: 1". 

Not sure if we can work around this in our build plugins or if we need to do an upstream PR capturing stderr.

johanandren:
Especially annoying in Eclipse as it fails without errors in the .proto-files for some reason that is not available in any logs, as far as I can figure out.

johanandren:
Which just redirects the output of the process to the JVM stderr/out so we can capture those (that's what I did for the maven plugin)

PR #243, opened 2018-06-21
m2e support

### LINKED ISSUE #171: Eclipse support for the maven plugin
Making a plugin work well in eclipse seems to require some eclipse specific changes (!!):

* add a m2e lifecycle mapping xml file
* use the plexus BuildContext to detect changes and to inform about generated classes

I made a quick try, but got stuck on figuring out how to get that build context injected.

See more docs here: https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html

johanandren:
Fixes #171

### DIFF
--- a/maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
+++ b/maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
@@ -1,6 +1,6 @@
 package akka.grpc.maven
 
-import java.io.File
+import java.io.{ ByteArrayOutputStream, File, PrintStream }
 
 import akka.grpc.gen.CodeGenerator
 import akka.grpc.gen.javadsl.{ JavaBothCodeGenerator, JavaClientCodeGenerator, JavaServerCodeGenerator }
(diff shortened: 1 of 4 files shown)


johanandren:
Can't quite test that it works because #242

johanandren:
Somewhat solved #242 for maven

In eclipse: 

screen shot 2018-06-21 at 13 41 27

In the terminal running maven:

screen shot 2018-06-21 at 13 43 52

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
```diff
@@ -1,22 +1,60 @@
...
+  def parseError(errorLine: String): Either[ProtocError, String] = {
+    errorLine match {
+      case ProtocErrorRegex(file, line, pos, message) =>
+        Left(ProtocError(file, line.toInt, pos.toInt, message))
+      case unknown =>
+        Right(unknown)
+    }
+  }
+
+  private def captureStdOutAnderr[T](block: => T): (String, String, T) = {
```
raboof: Nice, though unfortunate that it's needed. Added https://github.com/os72/protoc-jar/issues/48

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
```diff
@@ -28,76 +66,109 @@ class GenerateMojo @Inject() (project: MavenProject) extends AbstractMojo {
...
+
+    // generated sources should be compiled
+    val generatedSourcesDir = "target/generated-sources/akka-grpc-" + language.name().toLowerCase
+    val compileSourceRoot = new File(project.getBasedir, generatedSourcesDir)
+    project.addCompileSourceRoot(generatedSourcesDir)
+
+
+
+    // only do any work if needed (m2e integration)
+    if (!buildContext.isIncremental && buildContext.hasDelta(generatedSourcesDir)) {
```
raboof: This does not make intuitive sense to me, but I'm not familiar with how this works so it might be OK?

johanandren: Yeah, the reason todo it is that eclipse has two phases, one for config and one for actually building. This was pretty much copied ad lib from one of the Java WSDL-generating plugins.

We don't actually do much incremental support as that would require hooking into the compilation in more fine grained detail - keeping track of what .proto generated what files etc.

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
johanandren: Thanks

maven-plugin/src/test/scala/akka/grpc/maven/ProtocSpec.scala
```diff
@@ -0,0 +1,21 @@
...
+class ProtocSpec extends WordSpec with Matchers {
+
+  "The protoc error messages" must {
+    "be parsed into details" in {
+      GenerateMojo.parseError("notifications.proto:12:1: Expected top-level statement (e.g. \"message\").") should
+        === (Left(GenerateMojo.ProtocError("notifications.proto", 12, 1, "Expected top-level statement (e.g. \"message\").")))
+
+    }
+    "be kept if not parseable" in {
+      GenerateMojo.parseError("My hovercraft is full of eels") should ===(Right("My hovercraft is full of eels"))
```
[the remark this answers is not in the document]
johanandren: Wasn't sure we always know what is spit out of protoc so I thought it's best to keep whatever it outputs when it can't be parsed into a line error.

maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
johanandren: I wonder if it should really be generatedSourcesDir though

johanandren: Actually, looking at a different generator-style maven plugin (antlr) I see they don't do this, let's skip this condition and always try to compile.

maven-plugin/src/test/scala/akka/grpc/maven/ProtocSpec.scala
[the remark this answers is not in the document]
raboof: Yes, that seems like a good strategy!

raboof:
(failure was #219)

Issue #155, opened 2018-04-06
Select client/server/both


2m:
https://github.com/akka/akka-grpc/pull/154#discussion_r179753307

2m:
I think this is a concern in the sbt plugin as well. Currently it is possible to select what to generate, but requires configuring multiple settings to achieve the desired result. We need something more user friendly to select from the `{java,scala} x {client,server}`.

patriknw:
@raboof is this done?

raboof:
java/scala switching is in there, but client/server/both switching not yet (it currently defaults to 'both').

patriknw:
What is the reason for having a `Seq` for `akkaGrpcGeneratedLanguages`? Doesn't make much sense to generate both java and scala, since class names will be conflicting?

raboof:
If you disable the `flat_package` option the class names no longer conflict. This seems like it could be convenient in some cases (for example we use this in the interop testing project)

patriknw:
ok, then there is a valid reason

patriknw:
seems like the gradle plugin puts the message files in wrong directory: plugin-tester-scala/build/generated/source/proto/main/scalapb/example/myapp/helloworld/grpc/HelloReply.scala

The service classes are right.

patriknw:
ah, it's `scalapb` vs `akkaGrpc`
(I was expecting everything together)

PR #244, opened 2018-06-21
generateClient and generateServer options in gradle plugin, #155

patriknw:
Refs #155

I will add docs also, after https://github.com/akka/akka-grpc/pull/240


raboof:
(checked failures are #219)

Issue #225, opened 2018-06-01
Decouple client settings from Netty specifics


johanandren:
So that we can easily migrate to Akka HTTP client when we have HTTP2 support there

raboof:
:+1:, I think we should make sure to support at least everything in the `CallCredentials` (perhaps even supply methods for easily applying `CallCredentials` that may come from a library)

patriknw:
I'm trying to understand what we would like to do here.

We have `GrpcClientSettings`, which makes it possible to define `CallOptions`. Looks like `CallOptions` is rather specific implementation details of grpc-java and perhaps not something we would like to expose.

`CallCredentials` is more interesting and I think `GrpcClientSettings` should have `withCallCredentials`. I don't think we should hide or abstract `CallCredentials` because there are library implementations of it that we might want to support, e.g. `GoogleAuthLibraryCallCredentials` that is included in `grpc-auth`.

Another thing is if there are things in `NettyChannelBuilder` or `ManagedChannelBuilder` that we want to expose in `GrpcClientSettings`. The only candidate I can find is `ManagedChannelBuilder.userAgent`. What else?

patriknw:
I see a risk with exposing too much, because when we switch underlying implementation we might have trouble supporting those or want to have slightly different settings. I think we should be conservative with the "nice to have" settings. Then I'd rather support a backdoor exposing the underlying implementation specific settings but make it clear that it will change.

raboof:
Done in #245

PR #245, opened 2018-06-21
Client settings, #225

patriknw:
* remove CallOptions since those expose too much grpc-java implementation specifics
* support CallCredentials, userAgent and deadline
* deadline required some refactoring to pass in the settings to the request builders,
  but that might be useful for other things also later. New CallOptions with the given
  deadline is needed for each call.

Refs #225


raboof:
Checked failures are #248 and #219

PR #247, opened 2018-06-22
Case unsensitivity for language in maven

### LINKED ISSUE #241: Configure language with case-insensitive string
Right now it's an enum, which means it must be configured as uppercase, which is ugly.

johanandren:
Fixes #241 

On top of #243, will rebase when that is merged.


raboof:
Thx for fixing the conflict! (failures are #219)

PR #250, opened 2018-06-22
Move composite generator to separate package

raboof:
Make it clearer that these are only used in the interop tests


raboof:
(failure was #219)

Issue #209, opened 2018-05-16
Provide a gRPC server in a Play application


raboof:
Documentation, example project (perhaps based on https://github.com/raboof/play-scala-grpc-server), and utilities to make this smooth.

Requires https://github.com/akka/akka-http/issues/1966 and perhaps #150 

Design decisions: 
* Where do these utilities live? Putting them in the akka-grpc project seems convenient as long as we're still moving quickly, but once our API's stabilize we might want to move them to a Play gRPC subproject since generation might be different for different Play versions.
* Should the akka-grpc handlers be added 'automatically' (DI'ing the service implementations?) or should users add them to their routes definition file? Adding them to the routes definition file seems most 'native'.
* the generated handler will expect the server implementation to be injected. This can be done by type or by name. When injecting by type, this must be the `FooServiceImpl` type, since the `FooService` interface is also used by the client so will be ambiguous. Alternatively, we could avoid referring to `FooServiceImpl` in the router, but inject the correct `FooService` instance by name. This however would mean we would have to generate bindings from the `FooServiceImpl` to the name, which again introduces the need to refer to the implementation class name, so that does not buy us much. Expecting the implementation under `controllers` seems fine for now.

Progress:
* [x] Scala codegen (#251)
* [x] Java codegen (#274)
* [ ] Scala example project (perhaps based on https://github.com/raboof/play-scala-grpc-server)
* [ ] Java example project
* [x] Documentation

johanandren:
Working on the h2c support

raboof:
h2c `Upgrade` support progress is in https://github.com/akka/akka-http/pull/2075. Looks good with curl, nghttp and wireshark AFAICS, but the grpc-java client crashes on our (AFAICS correct) response to the original request request coming in as stream 1.

For Play we have a couple of options to have 'normal' and gRPC routes co-exist:
- Use `h2c` with `Upgrade`
- Introduce a third port next to `http` and `https`, serving the routes over `h2c` without `Upgrade`
- Introduce a flag to switch the `http` port to `h2c` without `Upgrade`
- Only support HTTP2 on the https port

- `h2c`+`Upgrade` seems nicest, but perhaps not possibe for the MVP as it seems it'd require switching to Akka HTTP at the client
- Introducing a third port might be tricky in Play, as it uses a 'port pair' in a lot of the code - though we may still want this in the future.
- Switching the `http` port to `h2c` makes it unusable from browsers, which seems like a showstopper
[...]

PR #251, opened 2018-06-25
Codegen for Play Scala serverside

raboof:
Add subproject to test the generated play code compiles

Refs #209

See it in action in https://github.com/raboof/play-scala-grpc-server/pull/2/files

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
@@ -66,7 +66,7 @@ object ScalaCodeGenerator {
   val generateServiceFile: Service => CodeGeneratorResponse.File = service => {
     val b = CodeGeneratorResponse.File.newBuilder()
     b.setContent(ApiTrait(service).body)
-    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}.scala")
+    b.setName(s"${service.packageDir}/${service.name}.scala")
     b.build
   }
 }
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
@@ -19,10 +19,10 @@ import templates.ScalaCommon.txt._
 
 abstract class ScalaCodeGenerator extends CodeGenerator {
   // Override this to add generated files per service
-  def perServiceContent: Set[Service ⇒ CodeGeneratorResponse.File] = Set(generateServiceFile)
+  def perServiceContent: Set[Service ⇒ CodeGeneratorResponse.File] = Set.empty
 
   // Override this to add service-independent generated files
-  def staticContent = Set.empty[CodeGeneratorResponse.File]
+  def staticContent: Set[CodeGeneratorResponse.File] = Set.empty
 
   override def suggestedDependencies = Seq(Artifact(BuildInfo.organization, BuildInfo.runtimeArtifactName, BuildInfo.version))
 
@@ -52,13 +52,6 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
     b.build()
   }
 
-  private def generateServiceFile(service: Service): CodeGeneratorResponse.File = {
-    val b = CodeGeneratorResponse.File.newBuilder()
-    b.setContent(ApiTrait(service).body)
-    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}.scala")
-    b.build
-  }
-
   private def parseParameters(params: String): GeneratorParams = {
     params.split(",").map(_.trim).filter(_.nonEmpty).foldLeft[GeneratorParams](GeneratorParams()) {
       case (p, "java_conversions") => p.copy(javaConversions = true)
@@ -67,5 +60,13 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
       case (p, "single_line_to_string") => p.copy(singleLineToProtoString = true)
       case (x, _) => x
     }
+  }
+}
+object ScalaCodeGenerator {
+  val generateServiceFile: Service => CodeGeneratorResponse.File = service => {
+    val b = CodeGeneratorResponse.File.newBuilder()
+    b.setContent(ApiTrait(service).body)
+    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}.scala")
+    b.build
   }
 }
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
@@ -16,7 +16,7 @@ object PlayScalaServerCodeGenerator extends ScalaCodeGenerator {
   private val generateRouter: Service => CodeGeneratorResponse.File = service => {
     val b = CodeGeneratorResponse.File.newBuilder()
     b.setContent(Router(service).body)
-    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}Router.scala")
+    b.setName(s"${service.packageDir}/${service.name}Router.scala")
     b.build
   }
 }
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
@@ -1,0 +1,18 @@
+package akka.grpc.gen.scaladsl.play
+
+import akka.grpc.gen.scaladsl.{ ScalaCodeGenerator, Service }
+import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse
+import templates.PlayScalaServer.txt.Router
+
+object PlayScalaServerCodeGenerator extends ScalaCodeGenerator {
+  override def name: String = "akka-grpc-play-server-scala"
+
+  override def perServiceContent = super.perServiceContent + generateRouter
+
+  private val generateRouter: Service => CodeGeneratorResponse.File = service => {
+    val b = CodeGeneratorResponse.File.newBuilder()
+    b.setContent(Router(service).body)
+    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}Router.scala")
+    b.build
+  }
+}
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
@@ -1,3 +1,7 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
 package akka.grpc.gen.scaladsl.play
 
 import akka.grpc.gen.scaladsl.{ ScalaCodeGenerator, Service }
--- a/codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
+++ b/codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
@@ -1,0 +1,37 @@
+@*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ *@
+
+@(service: akka.grpc.gen.scaladsl.Service)
+
+package @service.packageName
+
+import javax.inject._
+
+import scala.concurrent.Future
+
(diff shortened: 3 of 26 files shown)


codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
```diff
@@ -0,0 +1,37 @@
...
+class @{service.name}Router @@Inject()(cc: ControllerComponents, impl: controllers.@{service.name}Impl)(implicit mat: Materializer)
+  extends Router {
+
+  val handler = new AkkaHttpHandler {
+    val h = @{service.name}Handler(impl)
+    override def apply(request: HttpRequest): Future[HttpResponse] = h(request)
+  }
+
+  // could look at @{service.name}.name, but isn't this already covered in the route file?
+  override def routes: Routes = { case _ ⇒ handler }
```
renatocaval: I have been thinking about this when integrating it in Lagom. 

In Lagom we don't have a route file, but we need to merge the `routes` defined by the `ServiceDescriptor` with this `routes`.
In that case, the gRPC routes will become a `catch-all` matching. 

On the other hand, I don't think we should use `@{service.name}.name` for that. The user should be able to choose on which path the gRPC will be exposed, much like they do when defining the path in a Play routes file.

raboof: The [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests) mentions: "Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged. gRPC does not go out of its way to break users that are using this kind of override, but we do not actively support it, and some functionality (e.g., service config support) will not work when the path is not of the form shown above"

So it might be OK to give the user some control over it, but preferably not too obviously. In fact I think it's a bit of a shame we have to mention it in the `routes`, but there I included it to make it more consistent with the other lines.

raboof:
@renatocaval sure, especially if it's more broadly usable it could make sense to provide such a building block in Play. We'd probably still have to generate some binding boilerplate though - but perhaps a bit less.

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
renatocaval: Interesting. In that case, we could see it the other way around. The `Router` we build for a gRPC handler should bind it to the path as prescribed by the spec (ie: ":path" "/" Service-Name "/" {method name}).  

When plugin it in Play, we should instead compose the routes in the app loader and not add it to the `routes` file. The drawback is that this will be fragile as one could always try to override the path by adding something different to the `routes` file.

raboof: I'm on the fence about whether to add the gRPC routes to the app loader 'automatically' or via the `routes`: doing it via the `routes` might make things 'less magic' and perhaps eventually allow participating in action composition, which could be nice in the light of 'platform consolidation'. On the other hand somehow adding them automatically (not sure how hard that would be?) may be less 'boilerplatey'. I'd love to hear e.g. @richdougherty's thoughts on this.

richdougherty: For the MVP I'm OK if we generate one router class for each gRPC service then let the user put each route into the routes file manually. Maybe go as far as printing out a message to help them.

For v2 we could patch the main routes file and/or generate a `grpc.routes` file. If we generate a `grpc.routes` file then the user could map it as `-> / grpc.Routes` or `-> /my/grpc/path grpc.Routes` from the main `routes` file. (Credit to @marcospereira for this idea.)

One thing that we need to be careful of is that if we have a route like `-> / my.custom.Routes` in the `routes` file then I believe it will catch all paths prefixed with `/` (i.e. every request), so we need to make sure it's the last entry in the `routes` file. If we use a route with a prefix, e.g. `->  /helloworld.GreeterService controllers.GreeterServiceRouter` then it will only match requests prefixed with `/helloworld.GreeterService` so it can probably go anywhere in the `routes` file without causing problems.

richdougherty: Can we add an `isDefinedAt` to the `Routes` partial function?

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
```diff
@@ -0,0 +1,37 @@
...
+import akka.http.scaladsl.model.{HttpRequest, HttpResponse}
+import akka.stream.Materializer
+
+import play.api.mvc._
+import play.api.mvc.akkahttp.AkkaHttpHandler
+import play.api.routing.Router
+import play.api.routing.Router.Routes
+
+@@Singleton
+class @{service.name}Router @@Inject()(cc: ControllerComponents, impl: controllers.@{service.name}Impl)(implicit mat: Materializer)
```
richdougherty: We don't need to inject `ControllerComponents` here since we're not doing building an action, etc.

project/Dependencies.scala
```diff
@@ -38,6 +38,9 @@ object Dependencies {
     val protocJar = "com.github.os72" % "protoc-jar" % "3.5.1"
 
     val plexusBuildApi = "org.sonatype.plexus" % "plexus-build-api" % "0.0.7" % "optional"// Apache v2
+
+    val play = "com.typesafe.play" %% "play" % "2.7.0-M1" // Apache M2
```
richdougherty: :smile:

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
```diff
@@ -0,0 +1,37 @@
+@*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ *@
+
+@(service: akka.grpc.gen.scaladsl.Service)
+
+package @service.packageName
```
richdougherty: What's the default packageName, out of interest?

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
raboof: @renatocaval proposed extending `SimpleRouter` to achieve that effect in https://github.com/raboof/play-scala-grpc-server/pull/3 - that seems like a neat way to achieve that?

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
raboof: It's determined as described in https://scalapb.github.io/customizations.html#package-options . By default we enable the `flat_package` option.

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
```diff
@@ -0,0 +1,49 @@
...
+import play.api.mvc._
+import play.api.mvc.akkahttp.AkkaHttpHandler
+import play.api.routing.Router
+import play.api.routing.Router.Routes
+
+@@Singleton
+class @{service.name}Router @@Inject()(impl: controllers.@{service.name}Impl)(implicit mat: Materializer)
+  extends Router {
+
+  protected val prefix: String = @{service.name}.name
```
richdougherty: I'm not sure, but I think the prefix might be set by the parent router based on the path given in the `routes` file.

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
```diff
@@ -0,0 +1,49 @@
...
+
+  val handler = new AkkaHttpHandler {
+    val h = @{service.name}Handler(impl)
+    override def apply(request: HttpRequest): Future[HttpResponse] = h(request)
+  }
+
+  // could look at @{service.name}.name, but isn't this already covered in the route file?
+  override def routes: Routes = {
+    case rh if rh.path.startsWith(prefix) ⇒ handler
```
richdougherty: If the `prefix` is set by the parent router then the prefix could be something like `/` and using `startsWith(prefix)` might not be precise enough. It might match paths that it shouldn't.

It might be safer to call `isDefinedAt` on the underlying `PartialFunction`. E.g.

```scala
  private val serviceFunc: PartialFunction[HttpRequest, Future[HttpResponse]] = GreeterServiceHandler(serviceImpl)(mat)

  private def toAkkaHttpRequest(rh: RequestHeader): HttpRequest = {
    rh.headers.asInstanceOf[AkkaHeadersWrapper].request
  }

  override def routes: Routes = new Routes {
    private val playHandler = AkkaHttpHandler(serviceFunc)
    override def isDefinedAt(rh: RequestHeader): Boolean = {
      val request = toAkkaHttpRequest(rh)
      val defined = serviceFunc.isDefinedAt(request)
      println(s"GreeterServiceRouter: route to ${rh.path} is $defined")
      defined
    }
    override def apply(rh: RequestHeader): Handler = playHandler
  }
```

Since this code uses some internal Play APIs (e.g. `asInstanceOf[AkkaHeadersWrapper]` it might make sense to add a helper Play to create a `Routes` given a `PartialFunction[HttpRequest, Future[HttpResponse]]`. This is similar to what @renatocaval suggested elsewhere.

codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
```diff
@@ -69,3 +62,11 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
     }
   }
 }
+object ScalaCodeGenerator {
+  val generateServiceFile: Service => CodeGeneratorResponse.File = service => {
```
ktoso: why not def?

ktoso: I see, they way you use it in ScalaServerCodeGenerator makes this nicer; Okey

codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
```diff
@@ -69,3 +62,11 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
...
+    val b = CodeGeneratorResponse.File.newBuilder()
+    b.setContent(ApiTrait(service).body)
+    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}.scala")
```
ktoso: Time for some util for `packageName.replace('.', '/')`?

codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaServerCodeGenerator.scala
```diff
@@ -0,0 +1,18 @@
+package akka.grpc.gen.scaladsl.play
```
ktoso: Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
raboof: Yes. Given:
* Play will create the route and call `withPrefix` on it with the prefix from the `routes` file
* the gRPC is opinionated on using `@{service.name}.name` as the prefix ("[Some gRPC implementations may allow the Path format shown above to be overridden, but this functionality is strongly discouraged](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)").

in zoom @renatocaval and I figured the most reasonable behavior would be to default to the prefix encouraged by the spec, and tell Play users to use the name in the `routes` file (#257) but honor a custom prefix when they configure it.

I now realize that this implementation doesn't quite achieve that yet, I'll add tests to both document and check the behavior ;).

codegen/src/main/twirl/templates/PlayScalaServer/Router.scala.txt
raboof: For gRPC specifically it seems uncommon to have other routes 'next to' the gRPC routes, so if the user specifies `/` as the prefix I think it might be reasonable to treat this as "you asked for it, you got it" there.

For the more general 'Akka HTTP routes in Play' case it might indeed be useful to really treat the route as a partial function all the way. Not sure about the performance impact. Might be OK to postpone that for now?

codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
ktoso: It has that already actually `def packageDir = packageName.replace('.', '/')`

raboof: Just added it in response to your comment ;)

ktoso:
Seems good right?

raboof:
:+1: seems like the main feedback was addressed, let's merge since some other PR's depend on this one, we can always refine in follow-up PR's as needed

Issue #151, opened 2018-04-06
Configuration


raboof:
We've discussed before (e.g. #67) that we should introduce our own Configuration classes for the client, to configure things like host/port, TLS, compression, authentication etc. 

It would also be useful to be able to easily consume configuration from a `io.grpc.CallCredentials` into this configuration (#87).

Since a gRPC client is generated for each specific protocol, this configuration is probably usually more 'static' compared to more general-purpose clients such as HTTP. For this reason it probably makes sense to make it easy to load the configuration from typesafe-config. That could look something like:

```
akka.grpc.client {
  "grpc.testing.TestService" = {
    # ...
  }
}
```

In play-grpc we could even use this configuration to be able to generate a client configured with these settings and make it available for easy injection.

ktoso:
I would propose having globbing in the selector season as well as “include-feature = on” for enabling more powerful things so one can do it selectively

raboof:
Done in #254

PR #254, opened 2018-06-27
Allow configuring clients from application.conf

raboof:
Refs #151


runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala
```diff
@@ -0,0 +1,42 @@
+package akka.grpc
```
ktoso: header?

raboof: ah will fix in separate PR

raboof: #255

runtime/src/main/resources/reference.conf
```diff
@@ -0,0 +1,4 @@
+akka.grpc.client."*" {
+  host: ""
+  port: 0
+}
```
patriknw: newline important here

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -3,15 +3,48 @@
...
 import io.grpc.CallCredentials
 
 object GrpcClientSettings {
   /** Scala API */
   def apply(host: String, port: Int): GrpcClientSettings =
     new GrpcClientSettings(host, port)
 
+  /** Scala API */
+  def apply(serviceName: String, sys: ActorSystem): GrpcClientSettings = {
+    val akkaGrpcClientConfig = sys.settings.config.getConfig("akka.grpc.client")
```
patriknw: I prefer two steps so that any config section can be used

```
def apply(serviceName: String, sys: ActorSystem): GrpcClientSettings =
  apply(serviceName, sys.settings.config.getConfig("akka.grpc.client"))

def apply(serviceName: String, config: Config): GrpcClientSettings = ...
```

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -3,15 +3,48 @@
...
+    fromSubConfig(sys.settings.config, serviceConfig)
+  }
+
+  /** Java API */
+  def create(serviceName: String, sys: ActorSystem): GrpcClientSettings =
+    GrpcClientSettings(serviceName, sys)
+
+  def fromSubConfig(root: Config, c: Config): GrpcClientSettings = {
+    var result = GrpcClientSettings(c getString "host", c getInt "port")
+    if (c.hasPath("override-authority"))
```
patriknw: I'd prefer that we define those in reference conf (for documentation purposes) with default values such as `off`

raboof: Done for the deadline. Do we also want this for string parameters? would the default be an empty string or some 'magic value'?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -3,15 +3,56 @@
...
+      result = result.withOverrideAuthority(config.getString("override-authority"))
+    if (config.hasPath("trusted-ca-certificate"))
+      result = result.withTrustedCaCertificate(config.getString("trusted-ca-certificate"))
+    if (config.hasPath("user-agent"))
+      result = result.withUserAgent(config.getString("user-agent"))
+
+    result
+  }
+
+  private def getPotentiallyInfiniteDuration(underlying: Config, path: String): Duration = underlying.getString(path) match {
```
patriknw: akka.util.Helpers.toRootLowerCase

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -3,15 +3,56 @@
...
+  def apply(config: Config): GrpcClientSettings = {
+    var result = GrpcClientSettings(config getString "host", config getInt "port")
+      .withDeadline(getPotentiallyInfiniteDuration(config, "deadline"))
+
+    if (config.hasPath("override-authority"))
+      result = result.withOverrideAuthority(config.getString("override-authority"))
+    if (config.hasPath("trusted-ca-certificate"))
+      result = result.withTrustedCaCertificate(config.getString("trusted-ca-certificate"))
+    if (config.hasPath("user-agent"))
+      result = result.withUserAgent(config.getString("user-agent"))
```
patriknw: > Do we also want this for string parameters? would the default be an empty string or some 'magic value'?

Yes, add all of them and `""` is a good to represent undefined here

Issue #59, opened 2018-03-05
'concat' helper method to chain together multiple service implementations?


raboof:
As discussed in https://github.com/lightbend/akka-grpc/pull/56

raboof:
Perhaps also helping the user not need/write the `orElse` in https://github.com/akka/akka-grpc/blob/master/plugin-tester-scala/src/main/scala/example/myapp/helloworld/GreeterServer.scala#L41

Issue #89, opened 2018-03-14
utility to setup TLS to make the first developer experience nice


patriknw:
Either support plain text mode as the Netty client is already doing, or have a utility for setting up dummy test certificates on both sides.

Related https://github.com/akka/akka-http/issues/1849

ignasi35:
> or have a utility for setting up dummy test certificates on both sides.

Akka could ship fake certs on the testkit, or there could be a tool to crete a CA and a server cert on the fly.

For Lagom we spiked the second alternative and built a [`FakeKeyStoreGenerator.scala`](https://github.com/ignasi35/lagom/commit/12bbd8a99947cfe3858235336635284c507d8cef#diff-7b40ab76a5e229e3fe091531bef2421f9) based on Play's [`FakeKeyStore`](https://github.com/playframework/playframework/blob/5f576efda61d9a1bd4cd27215e1fe221ddbaa3a9/framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala).

raboof:
I think @richdougherty has been working on that, seeing if it is reasonable to use ssl-config here or we can make it convenient some other way.

ktoso:
Right, I think our approach here should be "whichever style play already does it for dev" :)
If ssl config helps then good, but if not then don't force it in I think. It may be nice to use just to declare from where to load certs, but not expose it in any APIs unless needed perhaps for example.

patriknw:
Remember to remove the certs from Quickstarts also, e.g. https://github.com/akka/akka-grpc-quickstart-scala.g8/pull/1#discussion_r202044228

ktoso:
The comment to do what play does still stands, but they what play does does not work very well now: https://github.com/playframework/playframework/issues/8562

ktoso:
For plain Akka to generate the certs we'd need to "steal" code that play has for these things I suppose:

from here

Rich Dougherty: 
https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/scala/play/core/server/ssl/FakeKeyStore.scala#L53-L82

Rich Dougherty:
https://github.com/playframework/playframework/blob/master/framework/src/play-server/src/main/scala/play/core/server/ssl/DefaultSSLEngineProvider.scala#L62

ignasi35:
raised https://github.com/akka/akka-grpc-quickstart-java.g8/issues/39

ignasi35:
(sorry, pressed the wrong buttons)

[...]

PR #256, opened 2018-06-27
concat helper method to chain together multiple service implementations, #59

patriknw:
I have to test and add documentation, but wanted to see if you agree with the approach first.

Refs #59

### DIFF
--- a/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
@@ -24,18 +24,23 @@ import akka.grpc.javadsl.GrpcExceptionHandler;
 import static @{service.packageName}.@{service.name}.Serializers.*;
 
 public class @{service.name}HandlerFactory {
+  private static final CompletionStage<HttpResponse> notFound = CompletableFuture.completedFuture(
+    HttpResponse.create().withStatus(StatusCodes.NOT_FOUND));
+
+  /**
+   * Creates a `HttpRequest` to `HttpResponse` handler that can be used in for example
+   * `Http.get(system).bindAndHandleAsync`. It ends with `StatusCodes.NotFound` if the request is not matching.
+   *
+   * Use `akka.grpc.javadsl.ServiceHandler.concatOrNotFound` when combining several services.
+   */
   public static Function<HttpRequest, CompletionStage<HttpResponse>> create(@service.name implementation, Materializer mat) {
     return (req -> {
       Iterator<String> segments = req.getUri().pathSegments().iterator();
       if (segments.hasNext() && segments.next().equals(@{service.name}.name) && segments.hasNext()) {
         String method = segments.next();
         return handle(req, method, implementation, mat).exceptionally(e -> GrpcExceptionHandler.standard(e));
-      }
-      else
-      {
-        CompletableFuture<HttpResponse> result = new CompletableFuture<>();
-        result.completeExceptionally(new UnsupportedOperationException("Unexpected path"));
-        return result;
+      } else {
+        return notFound;
       }
     });
   }
--- a/plugin-tester-java/src/main/java/example/myapp/CombinedServer.java
+++ b/plugin-tester-java/src/main/java/example/myapp/CombinedServer.java
@@ -1,0 +1,112 @@
+package example.myapp;
+
+import akka.actor.ActorSystem;
+import akka.http.javadsl.ConnectWithHttps;
+import akka.http.javadsl.ConnectionContext;
+import akka.http.javadsl.HttpsConnectionContext;
+import akka.stream.ActorMaterializer;
+import akka.stream.Materializer;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+import javax.net.ssl.KeyManagerFactory;
+import javax.net.ssl.SSLContext;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.FileInputStream;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.security.KeyFactory;
+import java.security.KeyStore;
+import java.security.PrivateKey;
+import java.security.SecureRandom;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.util.Base64;
+import java.util.concurrent.CompletionStage;
+
+//#import
+import akka.grpc.javadsl.ServiceHandler;
+import akka.http.javadsl.Http;
+import akka.http.javadsl.model.HttpRequest;
+import akka.http.javadsl.model.HttpResponse;
+import akka.japi.Function;
+
+//#import
+
+import example.myapp.helloworld.*;
+import example.myapp.helloworld.grpc.*;
+import example.myapp.echo.*;
+import example.myapp.echo.grpc.*;
+
+class CombinedServer {
+  public static void main(String[] args) throws Exception {
+      // important to enable HTTP/2 in ActorSystem's config
+      Config conf = ConfigFactory.parseString("akka.http.server.preview.enable-http2 = on")
+        .withFallback(ConfigFactory.defaultApplication());
+      ActorSystem sys = ActorSystem.create("HelloWorld", conf);
+      Materializer mat = ActorMaterializer.create(sys);
+
+      //#concatOrNotFound
+      Function<HttpRequest, CompletionStage<HttpResponse>> greeterService =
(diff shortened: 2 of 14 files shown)


runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -0,0 +1,54 @@
...
+import scala.annotation.varargs
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.util.control.NonFatal
+
+import akka.http.javadsl.model.HttpRequest
+import akka.http.javadsl.model.HttpResponse
+import akka.http.javadsl.model.StatusCodes
+import akka.japi.{Function => JFunction}
```
raboof: Did you find a reason not to use `java.util.function.Function` here?

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -24,18 +24,17 @@ import akka.grpc.javadsl.GrpcExceptionHandler;
...
         String method = segments.next();
         return handle(req, method, implementation, mat).exceptionally(e -> GrpcExceptionHandler.standard(e));
-      }
-      else
-      {
-        CompletableFuture<HttpResponse> result = new CompletableFuture<>();
-        result.completeExceptionally(new UnsupportedOperationException("Unexpected path"));
-        return result;
+      } else {
+        return notFound;
```
patriknw: changed this to NOT_FOUND because I guess that's a more valid response if a single handler is used, and felt better to use that as the signal to continue with next handler in `concat` rather than using exception

plugin-tester-scala/src/main/scala/example/myapp/helloworld/GreeterServer.scala
```diff
@@ -38,8 +39,8 @@ object GreeterServer {
     implicit val mat: Materializer = ActorMaterializer()
     implicit val ec: ExecutionContext = sys.dispatcher
 
-    val service: HttpRequest => Future[HttpResponse] = GreeterServiceHandler(new GreeterServiceImpl(mat))
-      .orElse { case _  => Future.successful(HttpResponse(StatusCodes.NotFound)) }
+    val service: HttpRequest => Future[HttpResponse] =
+      ServiceHandler(GreeterServiceHandler(new GreeterServiceImpl(mat)))
```
patriknw: would a single implementation be "wrong" if used without the NotFound?
```
val service: HttpRequest => Future[HttpResponse] = GreeterServiceHandler(new GreeterServiceImpl(mat))
```

It might be easy to miss that?
Should we change `GreeterServiceHandler.apply` to be a total function appending the NotFound by default, and make the partial function another method? `GreeterServiceHandler.partial`?

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -0,0 +1,54 @@
...
+
+object ServiceHandler {
+
+  private val notFound = Future.successful(HttpResponse.create().withStatus(StatusCodes.NOT_FOUND))
+
+  /**
+   * Creates a `HttpRequest` to `HttpResponse` handler that can be used in for example `Http.get(system).bindAndHandleAsync`
```
ktoso: maybe as [[]] links? (the response/request)

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
patriknw: I'll change that occurrence also together with the other one

patriknw: I'll do that in this PR

plugin-tester-scala/src/main/scala/example/myapp/helloworld/GreeterServer.scala
patriknw: Then the `ServiceHandler.apply` that I added wouldn't be needed

patriknw:
> I think the utility looks useful.

I think that for javadsl it's essential, "impossible" to compose several otherwise

patriknw:
> As for naming, though we started with it perhaps 'concat' (as we've called it before) isn't great since unlike Route.concat it also ends with returning a 404 (which I think is useful).

`concatOrNotFound`?

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -24,18 +24,17 @@ import akka.grpc.javadsl.GrpcExceptionHandler;
 import static @{service.packageName}.@{service.name}.Serializers.*;
 
 public class @{service.name}HandlerFactory {
+  private static CompletionStage<HttpResponse> notFound = CompletableFuture.completedFuture(
+    HttpResponse.create().withStatus(StatusCodes.NOT_FOUND));
```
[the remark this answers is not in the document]
patriknw: yes, this is for the path segments, the method match is still using UnsupportedOperationException

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -0,0 +1,54 @@
...
+              case NonFatal(e) => Future.failed(e)
+            }
+          }
+      }
+    }
+
+    new JFunction[HttpRequest, CompletionStage[HttpResponse]] {
+      override def apply(req: HttpRequest): CompletionStage[HttpResponse] =
+        cont(req, notFound, handlers.toList).toJava
+    }
```
ktoso: I see, since we generate functions and not `javadsl.Route`, looks good

ktoso:
> concatOrNotFound?

Sounds good to me

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -24,18 +24,17 @@ import akka.grpc.javadsl.GrpcExceptionHandler;
 import static @{service.packageName}.@{service.name}.Serializers.*;
 
 public class @{service.name}HandlerFactory {
+  private static CompletionStage<HttpResponse> notFound = CompletableFuture.completedFuture(
```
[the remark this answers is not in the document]
patriknw: thx

patriknw:
Updated with the things we discussed so far. Will continue with test and docs

patriknw:
Added documentation and test.
Also changed the implementation for javadsl to use plain CompletionStage.

This should be ready for final review now.

patriknw:
and the reason for the japi.Function was that bindAndHandleAsync expected that, so I reverted that change

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -0,0 +1,49 @@
+/**
+ * Copyright (C) 2018 Lightbend Inc. <http://www.lightbend.com>
+ */
```
raboof: I think sbt-headers expects a newline after this

sbt-plugin/src/sbt-test/gen-scala-server/02-multiple-services/src/main/scala/example/myapp/Main.scala
```diff
@@ -18,13 +19,12 @@ object Main extends App {
   implicit val system = ActorSystem()
   implicit val mat = ActorMaterializer()
 
-  val echoHandler = EchoServiceHandler(new EchoServiceImpl)
-  val greeterHandler = GreeterServiceHandler(new GreeterServiceImpl)
+  val echoHandler = EchoServiceHandler.partial(new EchoServiceImpl)
+  val greeterHandler = GreeterServiceHandler.partial(new GreeterServiceImpl)
+  val serviceHandler = ServiceHandler.concatOrNotFound(echoHandler, greeterHandler)
```
raboof: :+1:

plugin-tester-java/src/main/java/example/myapp/CombinedServer.java
```diff
@@ -0,0 +1,112 @@
...
+          serviceHandlers,
+          ConnectWithHttps.toHostHttps("127.0.0.1", 8080).withCustomHttpsContext(serverHttpContext()),
+          mat)
+      //#concatOrNotFound
+      .thenAccept(binding -> {
+        System.out.println("gRPC server bound to: " + binding.localAddress());
+      });
+  }
+
+  private static HttpsConnectionContext serverHttpContext() throws Exception {
```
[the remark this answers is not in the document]
raboof: Yes, finding a nicer alternative for that is part of #89

patriknw:
Fixed headerCheck, rebased and squashed

PR #259, opened 2018-06-27
[not for merge] Don't half-close unary requests

raboof:
To see if it affects #219 (though AFAICS this should really be allowed)


raboof:
(seeing timeouts in the test results, I guess we do have to close the connection ;) )

Issue #137, opened 2018-03-29
Use Service Discovery


ignasi35:
Use Service discovery so `host:port` are not hardcoded.

ktoso:
Relates to managed calls I think... Agree, should be able to plug in some client-side load balancing or similar mechanisms using however we decide to make this pluggable

raboof:
Is this something we'd want to consider when designing the akka-http API's for client-side HTTP2 support? Previously it seemed we'd need little API changes (just implementation), does this mean we should revisit that?

ktoso:
Perhaps, though it's not like HTTP/2 is "new API" really, it should mirror the HTTP one. So it is more about Akka HTTP clientside perhaps needing something. Though in general I think this belongs into the managed calls... Perhaps we need the client report back things to the managed calls, like "errors keep happening on this connection" and the managed calls would decide to try a new host or something.

ignasi35:
Did some investigation recently: https://github.com/ignasi35/akka-grpc-labs

The issue spans into `Channel` pooling.

_NOTE_: I use the name `Channel` since that's the `gRPC` naming convention (beyond language or implementation details) and not because that's the current implementation.

ignasi35:
@raboof I'm not familiar with Akka HTTP client pooling connections but this is highly related. Multiplexing traffic over a `Channel` just adds a bit more complexity since a `Channel` is suddenly used for unary and streamed calls all at once and the lifecycle management gets tricker.

ignasi35:
side note: I poked into this as part of spiking the needs to use akka-grpc in lagom. See more lab work specific to lagom at https://github.com/ignasi35/lagom-grpc-labs

ignasi35:
>  Previously it seemed we'd need little API changes (just implementation), does this mean we should revisit that?

I think so. I think the `ServiceClient` should get a `ServiceDiscovery` instance which could default to [`HardcodedServiceDiscovery`](https://github.com/ignasi35/akka-grpc-labs/blob/master/managed-client/src/main/scala/com/lightbend/akka/labs/utils/HardcodedServiceDiscovery.scala).

Alternatively we could provide more generators:

 * default generator provides a `ServiceClient(Channel)`
[...]

PR #260, opened 2018-06-27
Integrate akka service discovery

chbatey:
* Allow host and port via hard coded service API
* No reconnection
* Picks a random endpoint and doesn't re-resolve
* Assuming re-resolving/load balancing will be part of managed calls

Not sure how useful this is until we load balance / re-resolve.


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -4,33 +4,57 @@
...
+
+  private def hardcodedServiceDiscovery(host: String, port: Int) = new HardcodedServiceDiscovery(Resolved(host, immutable.Seq(ResolvedTarget(host, Some(port)))))
 }
 
 final class GrpcClientSettings private (
-  val host: String,
-  val port: Int,
+  val name: String,
+  val serviceDiscovery: SimpleServiceDiscovery,
+  val defaultPort: Int = 443,
```
raboof: Perhaps the default port should depend on whether TLS is enabled? (we currently determine that by checking if `trustedCaCertificate` is defined, might be good to make that explicit by introducing a 'def useTls() = trustedCaCertificate.isDefined`, this might change based on the work by @richdougherty for #89)

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -4,33 +4,57 @@
...
   val callCredentials: Option[CallCredentials] = None,
   val overrideAuthority: Option[String] = None,
   val trustedCaCertificate: Option[String] = None,
   val deadline: Duration = Duration.Undefined,
-  val userAgent: Option[String] = None) {
+  val userAgent: Option[String] = None,
+  val resolveTimeout: FiniteDuration = 1.second) {
```
raboof: refer to `DefaultResolveTimeout` defined above, or remove that one?

runtime/src/main/scala/akka/grpc/internal/HardcodedServiceDiscovery.scala
```diff
@@ -0,0 +1,17 @@
...
+
+package akka.grpc.internal
+
+import akka.discovery.SimpleServiceDiscovery
+import akka.discovery.SimpleServiceDiscovery.Resolved
+
+import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+
+class HardcodedServiceDiscovery(resolved: Resolved) extends SimpleServiceDiscovery {
```
raboof: Is this a utility that should be provided by akka-management in the future? Can we add a ticket to track that?

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -25,21 +27,28 @@ object NettyClientUtils {
...
+          .map(ctx => builder.negotiationType(NegotiationType.TLS).sslContext(ctx))
+          .getOrElse(builder.negotiationType(NegotiationType.PLAINTEXT))
+
+        builder = settings.overrideAuthority.map(builder.overrideAuthority(_)).getOrElse(builder)
+
+        builder = settings.userAgent.map(builder.userAgent(_)).getOrElse(builder)
+
+        builder.build
+      } else {
+        throw new IllegalStateException("No targets returned for name: " + settings.name)
```
raboof: Not super nice to throw... we could also replace `map` above with `flatMap` and return a failed future here?

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
```diff
@@ -160,26 +170,32 @@ final class ScalaClientStreamingRequestBuilder[I, O](
...
-      case (metadata, result) =>
-        new GrpcSingleResponse[O] {
-          def value: O = result
-          def getValue(): O = result
-          def headers = metadata.headers
-          def getHeaders() = metadata.getHeaders()
-          def trailers = metadata.trailers
-          def getTrailers = metadata.getTrailers()
+
+    channel.flatMap { c =>
```
raboof: (perhaps consider a for comprehension, but this is not too bad either)

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -4,33 +4,57 @@
...
+  val DefaultResolveTimeout = 5.seconds
+  val DefaultPort = 443
+
   /** Scala API */
   def apply(host: String, port: Int): GrpcClientSettings =
-    new GrpcClientSettings(host, port)
+    new GrpcClientSettings(host, hardcodedServiceDiscovery(host, port))
+
+  /** Scala API */
+  def apply(name: String, serviceDiscovery: SimpleServiceDiscovery) =
```
johanandren: Scaladoc describing what name is, for all these

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
```diff
@@ -33,38 +33,42 @@ import akka.util.OptionVal
...
-    listener.future
+    channel.flatMap(c => {
+      val listener = new UnaryCallAdapter[O]
+      val call = c.newCall(descriptor, callOptionsWithDeadline())
+      call.start(listener, headers.toGoogleGrpcMetadata())
+      call.sendMessage(request)
+      call.halfClose()
+      call.request(2)
+      listener.future
+    })
```
johanandren: I wonder if we should optimize this right away for the case where the channel future is already completed (which should be most of the invocations) and avoid one extra dispatch on the ec?

raboof:
This does the lookup once when the channel is created but never checks back for changed endpoints.

Since channels are long-running, I think this means this approach only really works if the service discovery returns stable hostnames. We should ticketify refreshing the discovery, but for the MVP perhaps we can get away without it?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -4,33 +4,57 @@
...
+import akka.grpc.internal.HardcodedServiceDiscovery
 
+import scala.concurrent.duration.{ Duration, FiniteDuration }
+import scala.collection.immutable
 import io.grpc.CallCredentials
+import scala.concurrent.duration._
 
 object GrpcClientSettings {
+  val DefaultResolveTimeout = 5.seconds
+  val DefaultPort = 443
```
patriknw: move to reference.conf, now that conf settings has been merged https://github.com/akka/akka-grpc/pull/254

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
patriknw: I think we should remove all defaults here, and read them from config in apply

runtime/src/main/scala/akka/grpc/internal/HardcodedServiceDiscovery.scala
patriknw: internal api markers in `internal` package

runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
patriknw: 👍

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -25,21 +27,28 @@ object NettyClientUtils {
...
-    builder = settings.trustedCaCertificate.map(c => GrpcSslContexts.forClient.trustManager(loadCert(c)).build)
-      .map(ctx => builder.negotiationType(NegotiationType.TLS).sslContext(ctx))
-      .getOrElse(builder.negotiationType(NegotiationType.PLAINTEXT))
-
-    builder = settings.overrideAuthority.map(builder.overrideAuthority(_)).getOrElse(builder)
-
-    builder = settings.userAgent.map(builder.userAgent(_)).getOrElse(builder)
-
-    builder.build
+  def createChannel(settings: GrpcClientSettings)(implicit ec: ExecutionContext): Future[ManagedChannel] = {
```
patriknw: How does it work now when the connection/channel is disconnected (e.g. server stopped)? All running streams will fail and new calls will fail?

Also, what happens after `client.close()` (`channel.shutdown()`) will that fail new calls immediately or do we have to add some fail fast for that case.

Shall we expose something like `def closed: Future[Done]`? Then we can complete that in `close()` and also complete it when the channel is disconnected. Assuming that we could listen to such even in the `channel`.

Then one can at least act on this and restart the client and thereby do a new discovery. Poor mans retry/reconnect until we have better things built in.

Wdyt?

raboof: Since it's a `ManagedChannel` it will try to reconnect, but of course even then it won't run our service lookup anymore.

If we can listen to the `channel` disconnecting I think we should replace the `channel` 'inside' the client rather than leaving this up to the API user.

TimMoore:
Ref: #137

chbatey:
Last thing I was looking at was dealing with a connection failure that isn’t resolved by the managed channel. I can do this under a different PR as this is very likely to conflict

Issue #212, opened 2018-05-25
Client close doesn't kill stream on server


peteremiljensen:
I've a question. The server streams out stock quotes to a client.
But I can't seem to close the stream again. If calling `client.close` on the client-side, then the server stil streams quotes. It only stops if I terminate the client.

Is client.close not closing any living connections? Or am I misunderstanding something?

Thank you :)

raboof:
Hi Peter, thanks for the feedback! It is indeed correct that `shutdown` will do an orderly shutdown in which preexisting calls continue but new calls are immediately cancelled.

Can you tell a bit more about your use case? Typically I'd say gRPC clients might be long-running processes, what is your reason for closing it? Are you winding down the entire application? Would you want/expect to cancel both existing and new calls to be shut down on close?

peteremiljensen:
Hi Arnout.
Thanks for the quick reply.

I my use case the client should ideally also be long-running. However, I have a variable amount of actors that get spawned at different times subscribing to multiple different quotes. And they also get torn down frequently. 

But, when all the actors using a specific quote stops, I would like to be able to stop the specific quote currently streaming.
Therefore, as a last resort I tried closing everything with `client.close`, which makes sense to be long-living :)
Is there a way to stop particular streams end-to-end? I tried failing the stream using `KillSwitches`, but that didn't close the stream on the server.

I hope I'am clear. Otherwise just ask :)
Thanks

raboof:
I think ideally terminating the stream should indeed close it all the way to the server - it's entirely possible that that is not currently the case yet, but I think I'd consider that a bug - possibly also related to the changes currently going on in #201.

peteremiljensen:
Yes. Unfortunately, it doesn't seem to close the stream on the server.
But I'm on release v0.1 and not the master branch.

patriknw:
I started looking at this.

It's correct that the streaming calls continue when after we have done `channel.shutdown`. Otherwise we should use `channel.shutdownNow`.

We could also install KillSwitch in the streams if that is a more graceful way to close things.

I'm not sure what we would want to support?

* shutdownNow
* shutdown
* shutdown + KillSwitch

Perhaps the full graceful shutdown (without KillSwitch) might be difficult to implement later when we switch to Akka HTTP client?

I think I'll start by simply changing it to `shutdownNow`.

This is perhaps also related to my comment about exposing the client close state in https://github.com/akka/akka-grpc/pull/260/files#r199136901

[...]

PR #264, opened 2018-06-29
Use shutdownNow in client close, #212

patriknw:
* and close client when ActorSystem has terminated

Refs #212

See my reasoning about more graceful shutdown in the ticket https://github.com/akka/akka-grpc/issues/212#issuecomment-401346253


patriknw:
@raboof Changed Java client also, thanks. 5905d5f

PR #266, opened 2018-07-03
Use a Java SSLConfig for client configuration

richdougherty:
Add an `SSLContext` setting to the `GrpcClientSettings` and deprecate the current certificate setting.

Goals:
- Make client settings API more flexible than we have at the moment. Currently we can only provide a path to a certificate to trust. By providing an `SSLContext` a wide variety of trust options are supported.
- No dependency from akka-grpc's public API to any other SSL libraries (such as Netty's `SslContext/Builder`, Lightbend's [`ssl-config`](https://github.com/lightbend/ssl-config), Apache's [commons-net utilities](https://commons.apache.org/proper/commons-net/apidocs/org/apache/commons/net/util/package-summary.html)). This does mean there is currently a bit of a vacuum about _how_ users should actually go about creating an `SSLContext` object.

Todo:
- Work out our preferred way for configuring SSL in Akka gRPC. I understand that we do not want to have a public dependency on Netty APIs like `SslContextBuilder` in Akka gRPC. We could leave this the details of creating an `SSLContext` unspecified and then downstream users could decide for themselves. E.g. Play could use `ssl-config`, since it already uses that library for `play-ws` configuration. However, since it is likely that Akka users will want to know how to create an `SSLContext` when using Akka gRPC, at some point we'll probably need to either write an API or come up with a recommendation for a third-party API to use.
- Within the private `NettyClientUtils` class, work out a better way to go from an `SSLContext` to a usable Netty `SslContext`. At the moment this is done by creating a default Netty `SslContext` object - which contains a Java `SSLContext` internally - then patching in a new `SSLContext` using reflection! This is not a nice way to do things. The reason to do this for now is because there are a lot of SSL settings that Netty sets up that we basically need to re-implement ourselves, and I've put off doing that for now. A proper solution that avoids using reflection will unfortunately involve replicating a lot of Netty's internal functionality ourselves.


runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -42,33 +44,39 @@ object NettyClientUtils {
...
+    // Create a Netty JdkSslContext object with all the correct ciphers, protocol settings, etc initialized.
+    val nettySslContext: JdkSslContext = buildJdkSslContext(GrpcSslContexts.forClient)
 
-    val out = new BufferedOutputStream(new FileOutputStream(tmpFile))
-    out.write(inBytes)
-    out.close()
+    // Patch the SSLContext value inside the JdkSslContext object
+    val nettySslContextField: Field = classOf[JdkSslContext].getDeclaredField("sslContext")
+    nettySslContextField.setAccessible(true)
+    nettySslContextField.set(nettySslContext, javaSslContext)
```
richdougherty: I humbly apologise for this code.

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -42,33 +44,39 @@ object NettyClientUtils {
     builder.build
   }
 
-  // FIXME couldn't we use the inputstream based trustManager() method instead of going via filesystem?
-  private def loadCert(name: String): File = {
```
richdougherty: As suggested in the FIXME, this code could be removed because the Netty API supports reading from an `InputStream` as well as a `File`.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -30,14 +36,21 @@ object GrpcClientSettings {
...
-    GrpcClientSettings(config getString "host", config getInt "port")
+  def apply(config: Config): GrpcClientSettings = {
+    var settings = GrpcClientSettings(config getString "host", config getInt "port")
       .copy(
         overrideAuthority = getOptionalString(config, "override-authority"),
         deadline = getPotentiallyInfiniteDuration(config, "deadline"),
-        trustedCaCertificate = getOptionalString(config, "trusted-ca-certificate"),
         userAgent = getOptionalString(config, "user-agent"))
 
+    // TODO: Decide whether to present a deprecation warning for this option.
```
patriknw: this hasn't even been included in a release yet as far as I know, so deprecation not needed

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -58,22 +71,37 @@ object GrpcClientSettings {
...
   val userAgent: Option[String] = None) {
 
   def withHost(value: String): GrpcClientSettings = copy(host = value)
   def withPort(value: Int): GrpcClientSettings = copy(port = value)
   def withCallCredentials(value: CallCredentials): GrpcClientSettings = copy(callCredentials = Option(value))
   def withOverrideAuthority(value: String): GrpcClientSettings = copy(overrideAuthority = Option(value))
-  def withTrustedCaCertificate(value: String): GrpcClientSettings = copy(trustedCaCertificate = Option(value))
+  def withSSLContext(context: SSLContext) = copy(sslContext = Option(context))
+  @deprecated("Use withSSLContext instead", "0.2")
+  def withTrustedCaCertificate(value: String): GrpcClientSettings = copy(sslContext = Option(GrpcClientSettings.trustCert(value)))
```
patriknw: Many things have changed since 0.1 so no need for deprecation. Just remove it.

runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala
```diff
@@ -38,9 +51,16 @@ class GrpcClientSettingsSpec extends WordSpec with Matchers {
       parsed.host should be("myhost")
       parsed.port should be(42)
       parsed.overrideAuthority should be(Some("google.fr"))
-      parsed.trustedCaCertificate should be(Some("ca.pem"))
+      parsed.sslContext should be(None)
```
richdougherty: All the other settings are simple types like `String`, `Int` or `Duration`. That almost makes me feel that `SSLContext` is the wrong type of object for us and we need something like an `SSLConfig` object. However, having an object like that, is pending on a decision about which API Akka gRPC wants to expose for configuring SSL. For now `SSLContext` is the least opinionated way for us to configure SSL.

raboof:
While it is elegant and useful not to rely on `ssl-config` in our library code, I think it would be fine to show it in the documentation and example code.

How is SSL typically configured in Play when using the Netty backend? For gRPC the inelegant bridge from `SSLContext` to Netty `SslContext` might not be such a problem, since we intend to replace the generated Netty client code with generated Akka HTTP client code eventually.

runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala
raboof: Having convenience constructors to use a `SSLConfig` to create the `SSLContext` here could make sense - that would mean pulling in ssl-config as a dependency, but still not requiring it to be used if the user wants something more flexible. I'm not sure we should go so far to make it the only way, though.

richdougherty:
> While it is elegant and useful not to rely on ssl-config in our library code, I think it would be fine to show it in the documentation and example code.

Great! I'll try that out.

> How is SSL typically configured in Play when using the Netty backend?

Play has an `SSLEngineProvider` class which can be overridden by the user. The default implementation reads settings from configuration. I don't believe it uses the same configuration format as `ssl-config`, although it might be similar, since @wsargent implemented both.

Also, the Play SSL configuration is only for server configuration, and I believe the `ssl-config` configuration is only for client configuration.

> For gRPC the inelegant bridge from SSLContext to Netty SslContext might not be such a problem, since we intend to replace the generated Netty client code with generated Akka HTTP client code eventually.

I like this option. :)

interop-tests/src/test/scala/io/grpc/testing/integration/AkkaGrpcJavaClientTester.java
```diff
@@ -17,6 +17,7 @@
 import com.google.protobuf.ByteString;
 import io.grpc.Status;
 import io.grpc.StatusRuntimeException;
+import io.grpc.testing.integration.test.SSLContextUtils;
```
richdougherty: For now we have one utility method in our tests. In the future we might make a helper class somewhere for users to use.

plugin-tester-java/src/main/resources/application.conf
```diff
@@ -1 +1,16 @@
 akka.loglevel = INFO
+akka.grpc.client {
+  "helloworld.GreeterService" {
+    host = 127.0.0.1
+    port = 8080
+    override-authority = foo.test.google.fr
+    ssl-config {
```
richdougherty: This section is parsed by the `ssl-config` library.

interop-tests/src/test/scala/io/grpc/testing/integration/test/SSLContextUtils.scala
```diff
@@ -0,0 +1,20 @@
...
+import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts
+import io.grpc.netty.shaded.io.netty.handler.ssl.{ JdkSslContext, SslProvider }
+import javax.net.ssl.SSLContext
+
+object SSLContextUtils {
+
+  def sslContextForCert(certPath: String): SSLContext = {
+    // Use Netty's SslContextBuilder internally to help us construct a SSLContext
+    val fullCertPath = "/certs/" + certPath
+    val certStream: InputStream = getClass.getResourceAsStream(fullCertPath)
```
ignasi35: I think requiring users to add their CA's pem on `/certs/` is a bit limiting.

Could we refactor this to 

```
def sslContextForCert(trustCertResourceName: String) ={
  val certStream: InputStream = getClass.getResourceAsStream(trustCertResourceName)
  ...
}
```
?

Or am I missing something and the limit is on purpose?

plugin-tester-java/src/main/java/example/myapp/helloworld/GreeterClient.java
```diff
@@ -26,14 +26,10 @@ public static void main(String[] args) throws Exception {
...
 
-    GrpcClientSettings settings = GrpcClientSettings.create(serverHost, serverPort)
-      // Note: In this sample we are using a dummy TLS cert so we need to fake the authority
-      .withOverrideAuthority("foo.test.google.fr")
-      .withTrustedCaCertificate("ca.pem");
-
     ActorSystem system = ActorSystem.create("HelloWorldClient");
     Materializer materializer = ActorMaterializer.create(system);
 
+    GrpcClientSettings settings = GrpcClientSettings.create("helloworld.GreeterService", system);
```
ignasi35: Can we replace `"helloworld.GreeterService"` with `GreeterService.name` or is the string used on purpose to ensure the correct value is used?

plugin-tester-scala/src/main/resources/application.conf
```diff
@@ -1 +1,18 @@
 akka.loglevel = INFO
+
+
+akka.grpc.client {
+  "helloworld.GreeterService" {
+    host = 127.0.0.1
+    port = 8080
+    override-authority = foo.test.google.fr
+    ssl-config {
+      disabledKeyAlgorithms = [] // Allow weak certificates
```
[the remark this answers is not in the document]
ignasi35: it's not _ours_. 😂 

I think the CA certs built with https://github.com/lagom/lagom/pull/1443/files#diff-7b40ab76a5e229e3fe091531bef2421f are strong enough.

interop-tests/src/test/scala/io/grpc/testing/integration/test/SSLContextUtils.scala
johanandren: And also allow for certs with a path not on the classpath would be good, when we make it not a test-utility :)

plugin-tester-java/src/main/java/example/myapp/helloworld/GreeterClient.java
johanandren: That's how we will do it in the autogenerated client providers in play, but maybe explicitly stating it is more clear in an example like this.

plugin-tester-java/src/main/resources/application.conf
```diff
@@ -1 +1,16 @@
...
+      disabledKeyAlgorithms = [] // Allow weak certificates
+      trustManager {
+        stores = [
+          {path = certs/ca.pem, classpath = true, type = PEM}
```
johanandren: is this a relative path or class path?

raboof: https://lightbend.github.io/ssl-config/KeyStores.html#configuring-a-trust-manager suggests it's a filesystem path

raboof: Ah, but you can set a `classpath` boolean to load it from the classpath (https://github.com/lightbend/ssl-config/blob/master/ssl-config-core/src/main/scala/com/typesafe/sslconfig/ssl/Config.scala#L641)

johanandren:
Build failure looks like another aspect of the one I was seeing, something around getting the sbt jars or them getting corrupted or I don't know...

raboof:
OK TO TEST

raboof:
PLS BUILD

raboof:
PLS BUILD

raboof:
PLS BUILD

plugin-tester-java/src/main/resources/application.conf
richdougherty: You're right, it's a classpath value.

richdougherty:
Thanks guys

PR #272, opened 2018-07-04
Codegen for Scala Client Play DI

raboof:
Also contains #251, so review that first for a smaller diff

See it in action at https://github.com/raboof/play-scala-grpc-server/pull/6/files

Refs #208

PR #274, opened 2018-07-05
Java Play Server Codegen

ktoso:
On top of https://github.com/akka/akka-grpc/pull/272
Review only last commit;

I was not able to find a nice way to test the java + scala in one project since I guess we expect/demand the impl to be in `controllers.` so there "can be only one!".

Hope to finish this up and clean up once Arnouts PR is in


codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaServerCodeGenerator.scala
```diff
@@ -0,0 +1,24 @@
...
+object PlayJavaServerCodeGenerator extends JavaCodeGenerator {
+  override def name: String = "akka-grpc-play-server-java"
+
+  override def perServiceContent: Set[Service ⇒ CodeGeneratorResponse.File] =
+    super.perServiceContent + generateRouter
+
+  private val generateRouter: Service => CodeGeneratorResponse.File = service => {
+    val b = CodeGeneratorResponse.File.newBuilder()
+    b.setContent(Router(service).body)
+    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}Router.java")
```
raboof: Can now use the new helper ;)

codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
```diff
@@ -0,0 +1,85 @@
...
+  @@Override
+  public Optional<Handler> route(Http.RequestHeader request) {
+    if (request.path().startsWith(prefix)) return Optional.of(handler);
+    else return Optional.empty();
+  }
+
+  @@Override
+  public Router withPrefix(String prefix) {
+    if (prefix == "/") return this;
+    else throw new IllegalStateException("Only one level of prefixes is supported for gRPC");
```
raboof: I think we should allow `withPrefix` at least once, since Play uses this when loading the route from the `routes` file. It should typically be under `@{service.name}.name`, but to be consistent with the rest of the `routes` configuration approach it makes sense to allow serving under another prefix.

play-interop-test/src/main/scala/controllers/GreeterServiceImpl.scala
```diff
@@ -1,22 +1,22 @@
...
-
-  override def sayHello(in: HelloRequest): Future[HelloReply] = Future.successful(HelloReply(s"Hello, ${in.name}!"))
-
-  override def itKeepsTalking(in: Source[HelloRequest, NotUsed]): Future[HelloReply] = ???
-
-  override def itKeepsReplying(in: HelloRequest): Source[HelloReply, NotUsed] = ???
-
-  override def streamHellos(in: Source[HelloRequest, NotUsed]): Source[HelloReply, NotUsed] = ???
-}
+//package controllers
```
raboof: I guess we have no choice but to split the play-interop-test into 2 subprojects, one for scala and one for java

raboof: Eventually perhaps we might want to allow configuring code generation to generate some services as java and some as scala, but I'd say that's beyond MVC

codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaServerCodeGenerator.scala
ktoso: naisu

codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
ktoso: okey

play-interop-test/src/main/scala/controllers/GreeterServiceImpl.scala
ktoso: yes, i did 
split it up

PR #278, opened 2018-07-05
Allow for benching without tls

johanandren:
Adds actually caring about the `--tls` flag for the server, without the server binds as unencrypted HTTP/2

Also adds a `.useTls` flag to the `GrpcClientSettings`

### DIFF
--- a/benchmark-java/src/main/scala/akka/grpc/benchmarks/Unencrypted.scala
+++ b/benchmark-java/src/main/scala/akka/grpc/benchmarks/Unencrypted.scala
@@ -1,0 +1,32 @@
+/**
+ * Copyright (C) 2009-2018 Lightbend Inc. <http://www.lightbend.com>
+ */
+package akka.grpc.benchmarks
+
+import java.util.concurrent.CompletionStage
+
+import akka.actor.ActorSystem
+import akka.http.javadsl.ConnectHttp
+import akka.http.javadsl.model.HttpRequest
+import akka.http.javadsl.model.HttpResponse
+import akka.http.scaladsl.Http.ServerBinding
+import akka.http.scaladsl.{ConnectionContext, Http, UseHttp2}
+import akka.stream.Materializer
+
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.Future
+
+object Unencrypted {
+
+  // needed to bind unencrypted http2 due to short coming in Javadsl with 10.1.3
+  // can be removed when https://github.com/akka/akka-http/issues/2110 is fixed
+  def connect(system: ActorSystem, f: akka.japi.Function[HttpRequest, CompletionStage[HttpResponse]], connect: ConnectHttp, mat: Materializer): CompletionStage[ServerBinding] = {
+    Http(system)
+      .bindAndHandleAsync(
+        r => f.apply(r).toScala.asInstanceOf[Future[akka.http.scaladsl.model.HttpResponse]],
+        connect.host,
+        connect.port,
+        ConnectionContext.noEncryption()(UseHttp2.Always))(mat)
+      .toJava
+  }
+}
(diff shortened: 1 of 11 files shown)


benchmark-java/src/main/scala/akka/grpc/benchmarks/Unencrypted.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  // needed to bind unencrypted http2 due to short coming in Javadsl with 10.1.3
+  // can be removed when https://github.com/akka/akka-http/issues/2110 is fixed
+  def connect(system: ActorSystem, f: akka.japi.Function[HttpRequest, CompletionStage[HttpResponse]], connect: ConnectHttp, mat: Materializer): CompletionStage[ServerBinding] = {
+    Http(system)
+      .bindAndHandleAsync(
+        r => f.apply(r).toScala.asInstanceOf[Future[akka.http.scaladsl.model.HttpResponse]],
+        connect.host,
+        connect.port,
+        ConnectionContext.noEncryption()(UseHttp2.Always))(mat)
```
raboof: Could be simply `ConnectionContext(UseHttp2.Always)` I think? But :+1:

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -67,13 +67,15 @@ final class GrpcClientSettings private (
   val overrideAuthority: Option[String] = None,
   val trustedCaCertificate: Option[String] = None,
   val deadline: Duration = Duration.Undefined,
-  val userAgent: Option[String] = None) {
+  val userAgent: Option[String] = None,
+  val useTls: Boolean = true) {
```
raboof: We could also implement `useTls` here as `def useTls = trustedCaCertificate.isDefined`?

I think this will be removed/replaced in #266 anyway so :+1:

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -31,12 +31,15 @@ object NettyClientUtils {
...
-      .getOrElse(builder.negotiationType(NegotiationType.PLAINTEXT))
-
-    builder = settings.overrideAuthority.map(builder.overrideAuthority(_)).getOrElse(builder)
-
+    if (!settings.useTls)
+      builder = builder.usePlaintext()
+    else {
+      builder = settings.trustedCaCertificate.map(c => GrpcSslContexts.forClient.trustManager(loadCert(c)).build)
+        .map(ctx => builder.negotiationType(NegotiationType.TLS).sslContext(ctx))
+        .getOrElse(builder.negotiationType(NegotiationType.PLAINTEXT))
```
raboof: I guess we want to throw an error instead of falling back to PLAINTEXT when `useTls` is enabled but `trustedCaCertificate` is nevertheless `None`?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
johanandren: Oh, I was under the impression that it already worked with the default keychain if you don't specify and the trusted is just for self signed?

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
johanandren: I didn't really change anything in this branch, just moved the original code down here, but maybe that is true.

benchmark-java/src/main/scala/akka/grpc/benchmarks/Unencrypted.scala
johanandren: I have some problems finding my ways around those APIs...

PR #282, opened 2018-07-06
Handle defaults better, fail to connect if no host has been set 

### LINKED ISSUE #271: Better error message when client settings are incomplete
When configuring a `Client` through `ClientSettings(serviceName, system)`, when there is no configuration found for `serviceName`, we currently get the exception `Caused by: java.net.URISyntaxException: Expected hostname at index 2: //:0`.

Validate that hostname is present when creating a Client, but not too early since it should be possible to load default 'generic' settings from the configuration and add host+port per client from code.

johanandren:
Fixes #271

### DIFF
--- a/runtime/src/main/resources/reference.conf
+++ b/runtime/src/main/resources/reference.conf
@@ -1,6 +1,8 @@
 akka.grpc.client."*" {
-  host = ""
-  port = 0
+  # these will only go in service specific blocks to identify
+  # host and port and are not set in the default config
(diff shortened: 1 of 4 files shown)


johanandren:
Could be we should get #260 merged first and then fix this on top of that to not make that PR more painful than necessary.

johanandren:
Actually, that changed the stage so much that it doesn't make sense to try to rebase. I'll create a new PR

PR #286, opened 2018-07-11
Provide a Future/CompletionStage to monitor if client is healthy

chbatey:
After a configured number of retries fails the future and closes
down the channel.

I don't get any "leaked" channel warnings with these changes.


runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.{ ExecutionContext, Future, Promise }
+
+/**
+ * INTERNAL API
+ */
+@InternalApi
+object ChannelUtils {
+
+  def createDonePromise(): Promise[Done] = Promise[Done]()
```
johanandren: Best method ever!

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+            }
+          })
+        }
+      }
+    }
+
+    monitor(channel.getState(false), 0)
+
+  }
+}
```
johanandren: Would a Channel wrapping class containing the promise and channel be cleaner than passing a promise around like this?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -106,6 +107,7 @@ final class GrpcClientSettings private (
   val serviceDiscovery: SimpleServiceDiscovery,
   val defaultPort: Int,
   val resolveTimeout: FiniteDuration,
+  val connectionAttempts: Int = 5,
```
patriknw: is default needed here, doesn't that come from apply (config)?

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
patriknw: Is that because Promise.apply isn't accessible from Java with Scala 2.11? Perhaps add a comment

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+  def close(channel: Future[ManagedChannel], done: Promise[Done])(implicit ec: ExecutionContext): Future[Done] = {
+    channel.foreach(_.shutdown())
+    done.future
+  }
+
+  def closeCS(channel: Future[ManagedChannel], done: Promise[Done])(implicit ec: ExecutionContext): CompletionStage[Done] = {
+    close(channel, done).toJava
+  }
+
+  def monitorChannel(done: Promise[Done], channel: ManagedChannel, maxConnectionAttempts: Int): Unit = {
```
patriknw: this isn't used directly from generated code so could be `private[akka]`

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    def monitor(previousState: ConnectivityState, connectionAttempts: Int): Unit = {
+      if (connectionAttempts == maxConnectionAttempts) {
+        // shutdown is idempotent in ManagedChannelImpl
+        channel.shutdown()
```
patriknw: `shutdownNow`

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+        done.failure(new RuntimeException("Unable to establish connection"))
```
patriknw: after [$connectionAttempts] attempts

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+      } else {
+        val currentState = channel.getState(false)
+        if (currentState == ConnectivityState.SHUTDOWN) {
+          done.success(Done)
```
patriknw: `trySuccess` to be on the safe side (I don't fully know the threading model of these callbacks)

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
patriknw: tryFailure to be on the safe side (I don't fully know the threading model of these callbacks)

johanandren:
Fixes #217 I think

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
chbatey: There's a path to create that doesn't go via any Config

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
chbatey: Does it make sense to use the forceful one? I would expect all queries to have failed by now anyhow but can it hurt to allow a graceful shutdown?

chbatey:
Yep it does 217

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
chbatey: It is but via the $MONDULE$ and i didn't really want that in the generated code that ppl may look at and we have to maintain without IDE support :)

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
chbatey: Yes I think it would be cleaner!

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
chbatey: Removed this by having the promise inside a channel wrapper so the clients create it

chbatey:
Pushed review feedback. Wrapped managed channel and promise in a class. Only thing not done is change shutdown to shutdownNow.

docs/src/main/paradox/client.md
```diff
@@ -165,6 +165,11 @@ avoid leaking in the latter case, you should call `.close()` on the client.
 When the connection breaks, the client will start failing requests and try reconnecting
 to the server automatically.
 
+If a connection can not be established after the configured number of attempts then
+the client closes its self. When this happens the @scala[`Future`]@java[`CompletionStage`] 
+returned by `closed()` will complete with a failure. You do not need to call `close()` in
+this case.
```
raboof: This means creating a client and not listening on the `closed()` future is basically a bug, right? Do we have examples we have to update to show that practice?

chbatey: Yeah that also makes me think this should be a much larger default to take advantage of the built in backoff etc

runtime/src/main/scala/akka/grpc/internal/InternalChannel.scala
```diff
@@ -0,0 +1,24 @@
...
+/**
+  * INTERNAL API
+  * Used from generated code so can't be private.
+  */
+class InternalChannel(val managedChannel: Future[ManagedChannel], promiseDone: Promise[Done]) {
+
+  val doneCS = promiseDone.future.toJava
+  val done = promiseDone.future
+
+}
```
johanandren: Much nicer.

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -13,6 +13,15 @@ import io.grpc.{ ConnectivityState, ManagedChannel }
...
 import scala.concurrent.{ ExecutionContext, Future, Promise }
 
+/**
+  * Used to indicate that a gRPC client can not establish a connection
+  * after the configured number of attempts.
+  *
+  * Can be caught to re-create the client if it is likely that
+  * your service discovery mechanism will resolve to different instances.
+  */
+class ClientConnectionException(msg: String) extends RuntimeException(msg)
```
johanandren: If you liked it, make it final.

runtime/src/main/scala/akka/grpc/internal/AkkaGrpcClient.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import scala.concurrent.Future
+
+/**
+ * INTERNAL API
+ *
+ * Public as is included in generated code.
+ */
+@ApiMayChange
+trait AkkaGrpcClient {
```
johanandren: `@InternalApi`

PR #287, opened 2018-07-11
minor things found when writing quickstart

patriknw:


docs/src/main/paradox/server.md
```diff
@@ -13,8 +13,14 @@ sbt
 ```scala
 // in project/plugins.sbt:
 addSbtPlugin("com.lightbend.akka.grpc" % "sbt-akka-grpc" % "$projectversion$")
+addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.4") // ALPN agent
```
raboof: It'd have been nice if they could come in as transitive dependencies of the sbt-akka-grpc plugin - but I guess they don't :)

johanandren: Just FYI: I moved this out of here and into the specific plugin pages, with a link below, to make this consistently incomplete (but maybe that was a bad idea). The reason was that the other plugins also has specific configuration for enabling the agent, and that takes more space to explain.

johanandren: The link is in the *Running* section, which used to be just below this but now has gotten offset by more server samples.

patriknw: yes, since it's just two lines in sbt I think it doesn't hurt to include it there but for maven and gradle your motivation makes sense to leave it out here and have it in the full details. I'm writing the gradle page now, by the way.

docs/src/main/paradox/server.md
```diff
@@ -13,8 +13,14 @@ sbt
...
+
 // in build.sbt:
 enablePlugins(AkkaGrpcPlugin)
+
+// ALPN agent
```
ktoso: small mention why one needs it would be okey here perhaps good here?

PR #291, opened 2018-07-12
Simpler server usage in Play

johanandren:
This allows for just one class, annotated as Singleton (or injected in some other way - no longer tied to Guice) referenced in the routes file like so:

```
@Singleton 
class MyWhateverNameILike @Inject() (i: Injector) extends AbstractGreeterServiceRouter(i) {
 ... gRPC method implementations ...
} 

in the routes file
->     /helloworld.GreeterService   whatever.package.MyWhateverNameILike
```

Originally conceived here: https://github.com/raboof/play-scala-grpc-server/pull/8/files

### DIFF
--- a/docs/src/main/paradox/play-framework.md
+++ b/docs/src/main/paradox/play-framework.md
@@ -1,10 +1,7 @@
 # Play Framework
 
-
-## Serving gRPC from a Play Framework app
-
-
-
+Akka gRPC has special support for both consuming gRPC services through an Akka gRPC client and for implementing
+your own gRPC service as a part of a Play Framework application.
 
 ## Using a gRPC client in a Play Framework app
 
@@ -12,9 +9,16 @@ Akka gRPC has special support to allow for seamless injection of generated clien
 need to first enable the gRPC plugin as described in the @ref[client docs](client.md) and then add an additional
 source generator in `build.sbt`:
 
-```
+Scala
+:  ```
 import akka.grpc.gen.scaladsl.play.PlayScalaClientCodeGenerator
 akkaGrpcExtraGenerators += PlayScalaClientCodeGenerator
+```
+
+Java
+:  ```
+import akka.grpc.gen.javadsl.play.PlayJavaClientCodeGenerator
+akkaGrpcExtraGenerators += PlayJavaClientCodeGenerator
 ```
 
 This will generate a Play module that provides all generated clients for injection. The module must be enabled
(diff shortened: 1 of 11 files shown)


johanandren:
@richdougherty your opinion on this solution would be great

johanandren:
Refs #209

johanandren:
Now updated according to the feedback from the Play team

play-interop-test-java/src/main/java/controllers/GreeterServiceImpl.java
```diff
@@ -5,18 +5,26 @@
 package controllers;
 
 import akka.NotUsed;
+import akka.stream.Materializer;
 import akka.stream.javadsl.Source;
-import example.myapp.helloworld.grpc.GreeterService;
+import com.google.inject.Inject;
```
raboof: (no longer necessary?)

play-interop-test-java/src/test/scala/akka/grpc/gen/PlayJavaRouterSpec.scala
```diff
@@ -79,21 +73,10 @@ class PlayJavaRouterSpec extends WordSpec with Matchers with BeforeAndAfterAll w
...
-      val name = "John"
-
-      val handler = router.route(playRequestFor(uri)).get().asInstanceOf[AkkaHttpHandler]
-      val request = akkaHttpRequestFor(uri, HelloRequest.newBuilder().setName(name).build())(HelloRequestSerializer)
-      val response = handler(request).futureValue
-      response.status shouldBe StatusCodes.OK
-
-      val reply = akkaHttpResponse[HelloReply](response).futureValue
-      reply.getMessage shouldBe s"Hello, $name!"
+    "not allow specifying another prefix" in {
```
raboof: Hm so we drop this feature - that's OK I guess, it was limited and explicitly discouraged from the spec anyway :).

runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  final override def documentation: Seq[(String, String, String)] = Seq.empty
+
+  /**
+   * Registering a gRPC service under a custom prefix is not widely supported and strongly discouraged by the specification
+   * so therefore not supported.
+   */
+  final override def withPrefix(prefix: String): Router =
+    throw new UnsupportedOperationException("Prefixing gRPC services is not widely supported, " +
+      "strongly discouraged by the specification and therefore not supported")
```
raboof: I'm not sure we can get away with this: I think Play will call `withPrefix` on the `Router` with the prefix specified in the `routes` file.

We could check if `prefix` is equal to `serviceName`, if so return `this`, and if not indeed throw an exception.

Might be nice to include both the provided and the expected service name in the exception text?

play-interop-test-scala/src/test/scala/akka/grpc/gen/PlayScalaRouterSpec.scala
```diff
@@ -21,24 +21,23 @@ import akka.stream.scaladsl.{ Sink, Source }
...
 
-class PlayRouterSpec extends WordSpec with Matchers with BeforeAndAfterAll with ScalaFutures {
+class PlayScalaRouterSpec extends WordSpec with Matchers with BeforeAndAfterAll with ScalaFutures {
   implicit val sys = ActorSystem()
   implicit val mat = ActorMaterializer()
   implicit val ec = sys.dispatcher
   implicit val patience = PatienceConfig(timeout = 3.seconds, interval = 15.milliseconds)
 
-  "The generated Play Router" should {
+  val router = new GreeterServiceImpl
```
raboof: :+1: nice

runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
johanandren: I think i tested this in an actual Play application, but I'll double check

johanandren: Would actually be nice if the play interop testers could be play projects, but I guess you guys tested that and realized it was problematic?

johanandren: You are perfectly right, how else would it work. 🤦‍♂️

johanandren:
Verified that it works (with the latest commits) in a standalone play project called from a separate Akka gRPC client project.

Still missing docs though, so I'd rather merge the client work first.

raboof:
LGTM but let's review #293 first as requested then :+1:

ktoso:
Please review whenever you'd have time @richdougherty :)

docs/src/main/paradox/play-framework.md
```diff
@@ -50,3 +54,58 @@ Scala
...
+    ```scala
+    // in project/plugins.sbt:
+    addSbtPlugin("com.lightbend.akka.grpc" % "sbt-akka-grpc" % "$projectversion$")
+    ```
+    @@@
+
+
+Then you need to enable the Play server side code generator in `build.sbt`:
+
+Scala
```
ktoso: This should eventually be sbt/maven etc, not by lang

johanandren: No, because Play with sbt is the default and maven is semi-supported, and these are different configurations to trigger generation of Scala or Java client/interfaces for the service.

johanandren: However, I see now that I made a copy-pasta-miss here: importing the server and using the client generator. D'oh!

johanandren: Fixed in #304

ktoso: (tho here is no maven here, i know, so just sbt)

docs/src/main/paradox/play-framework.md
```diff
@@ -50,3 +54,58 @@ Scala
...
+```
+
+Java
+:   ```scala
+enablePlugins(AkkaGrpcPlugin)
+import akka.grpc.gen.javadsl.play.PlayJavaServerCodeGenerator
+akkaGrpcExtraGenerators += PlayJavaClientServerGenerator
+```
+
+The plugin will look for `.proto` service descriptors in `src/main/protobuf` and output an abstract class per service
```
richdougherty: Might be in `app/protobuf` instead of `src/main/protobuf`, e.g. https://github.com/raboof/play-scala-grpc-server/tree/master/app/protobuf, https://www.playframework.com/documentation/2.7.0-M2/Anatomy#Default-SBT-layout.

project/Dependencies.scala
```diff
@@ -94,7 +94,10 @@ object Dependencies {
     Compile.akkaHttpCore,
     Compile.akkaHttp,
     Compile.akkaHttp2Support,
-    Compile.akkaDiscovery
+    Compile.akkaDiscovery,
+    // these two are available when used through Play, which is also the only case when they are needed
+    Compile.play % "provided",
+    Compile.playAkkaHttpServer % "provided",
```
richdougherty: :+1:

runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
```diff
@@ -0,0 +1,74 @@
...
+      .map(javaResp => javaResp.asInstanceOf[akka.http.scaladsl.model.HttpResponse])(ExecutionContexts.sameThreadExecutionContext))
+}
+
+/**
+ * Boiler plate needed for the generated Play routers allowing for adding a service implementation in a Play app,
+ * inherited by the generated abstract service router (both Java and Scala) which is then implemented by the user.
+ *
+ * INTERNAL API
+ */
+@InternalApi abstract class PlayRouter(mat: Materializer, serviceName: String) extends play.api.routing.Router {
```
richdougherty: :+1:

docs/src/main/paradox/play-framework.md
ignasi35: I think Play docs (or migration guide) should encapsulate this particular info (not akka-grpc's).

johanandren: One idea we have discussed is to have the docs in the play docs. But since we have postponed doing that (and deciding to do that perhaps). Tracked in #300

Issue #273, opened 2018-07-04
All generated sources should contain a comment declaring this


ktoso:
For our sanity perhaps it would be also good to include the version of tools they were generated with?

Issue #208, opened 2018-05-16
Use a gRPC client in a Play application


raboof:
Would it suffice to describe the client in configuration (#151) - can we then make the actual client available through dependency injection as is 'native' in Play?

Progress:
* [x] Scala codegen (#272)
* [x] Java codegen
* [ ] Documentation (in Play, see omnidoc)
* [ ] Java example project (perhaps based on https://github.com/richdougherty/play-grpc-experiment)
* [ ] Scala example project

johanandren:
If we can figure out a way to hook in logic automagically we would still need the user to specify what host+port or service name to use, I'm guessing we cannot assume that there will always be a 1:1 between service type and host you'd want to connect to either.

I wonder if it is maybe best to just provide a sample of how to bind it for injection, and one sample for how to create the client ad hoc in a controller instead of adding API for this?

raboof:
#272 already reads that information from the configuration to make the client available for injection via DI without any further code.

Indeed there could be situations where it is not possible to put host+ip/name in the config, in that case you could create the client 'manually' (like with normal Akka gRPC).

Seems like a reasonable compromise to me?

johanandren:
Aha, I missed out on the `ClientProvider` there. I see, clever. And that module will be usable from both Scala and Java so no "Java codegen" needed, only docs?

johanandren:
If we could have a pre-made `Module` that picks up the generated list of clients from somewhere else, rather than generating the `Module` (which play autoloaders but would potentially conflict with the manual `Module` created by the user) that could make it possible to not have to add any extra config to hook the clients in.

raboof:
I'm not *sure* the `ClientProvider` is usable for Java, finding out and adding Java codegen if needed is one of the open tasks for this issue.

raboof:
I didn't find a way to make the generated clients discoverable so we can load them all from one `Module`, so far I've decided to live with the fact that the user needs to manually enable each generated client `Module` in the `application.conf`. Would definitely be neat if we can avoid it though ;).

johanandren:
Just saw here now that the plan was to add docs in the Play documentation, I just added such docs on the Akka gRPC side. 

Hard to do a PR to play with docs before we have a release of akka-grpc though, unless we are ok with non-validated/compiled docs code ofc.

johanandren:
Created separate tickets to track the in-play docs and the example projects.

PR #293, opened 2018-07-12
Easier life for play apps consuming a service

johanandren:
A single module to register will hook all clients into dependency injection. The per-client settings are looked up from config via serviceNames derived from the generated service name.

The user will then only have to register that single module in the play config to get things wired up.

Mostly poking around and checking what is possible so far.


codegen/src/main/scala/akka/grpc/gen/scaladsl/play/PlayScalaClientCodeGenerator.scala
```diff
@@ -20,10 +20,13 @@ object PlayScalaClientCodeGenerator extends ScalaCodeGenerator {
...
-  private val generateModule: Service => CodeGeneratorResponse.File = service => {
-    val b = CodeGeneratorResponse.File.newBuilder()
-    b.setContent(Module(service).body)
-    b.setName(s"${service.packageName.replace('.', '/')}/${service.name}Module.scala")
-    b.build
+  override def staticContent(allServices: Seq[Service]): Set[CodeGeneratorResponse.File] = {
+    if (allServices.nonEmpty) {
+      val b = CodeGeneratorResponse.File.newBuilder()
+      b.setContent(AkkaGrpcClientModule(allServices).body)
+      b.setName(s"${allServices.head.packageName.replace('.', '/')}/AkkaGrpcClientModule.scala")
```
johanandren: choosing package from the first client is maybe not ideal, another option would perhaps be least common sub package (trixy and maybe a bit surprising) and a default location (`akka.grpc.play.AkkaGrpcClientModule`) but that would make it impossible to re-use a pre-packaged client jar as the modules would have the same name.

raboof: We should discourage 'client libraries', but perhaps still allow them if the proto codegen might contribute a lot to build time/complexity. Otherwise `akka.grpc.play.AkkaGrpcClientModule` would have been nice :)

johanandren: For example a library that doesn't really contain the gRPC as user API but rather has a separate API that somewhere inside is using Akka gRPC to talk to a service.

johanandren:
Refs #208

johanandren:
I guess the docs may end up on the play side sooner or later, but at least this is a start.

codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaClientCodeGenerator.scala
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  override def staticContent(allServices: Seq[Service]): Set[CodeGeneratorResponse.File] = {
+    if (allServices.nonEmpty) {
+      val packageName = packageForSharedModuleFile(allServices)
+      val b = CodeGeneratorResponse.File.newBuilder()
+      b.setContent(AkkaGrpcClientModule(packageName, allServices).body)
+      b.setName(s"${packageName.replace('.', '/')}/AkkaGrpcClientModule.java")
+      val set = Set(b.build)
+      println(s"Generated [${packageName}.AkkaGrpcClientModule] add it to play.modules.enabled and a section " +
```
ktoso: extract this name as val AkkaGrpcClientModule, it's used in more places, in case we rename it

codegen/src/main/twirl/templates/PlayJava/AkkaGrpcClientModule.scala.txt
```diff
@@ -0,0 +1,31 @@
+@*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ *@
```
ktoso: rather I would say such generated files should contain info like "generated ...."?

ktoso: Whoops my bad, that's just comment here...

How about adding the // generated... comment though anyway or should we do this together in all files later?

johanandren: Yeah, lets' do that separately, I think we have a ticket for it already

ktoso: (I mean a huge "GENERATED DO NOT EDIT" at the top of the file, not as scaladoc on the classes, same as protobuf does)

docs/src/main/paradox/play-framework.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## Serving gRPC from a Play Framework app
+
+
+
+
+## Using a gRPC client in a Play Framework app
+
+Akka gRPC has special support to allow for minimal fuzz injection of generated clients in Play. To enable this you
```
ktoso: minimal fuzz sounds weird. Maybe "seamless injection of ...."

codegen/src/main/twirl/templates/PlayJava/AkkaGrpcClientModule.scala.txt
johanandren: #273

johanandren:
I think given how short we are on people maybe we just merge this, and if things aren't 100% we work on that later.

Issue #276, opened 2018-07-05
Configure additional code generators


raboof:
To enable play support under gradle, you should append the play codegen to the list of code generators. This is currently possible in sbt but not yet in the gradle plugin

raboof:
The akka-grpc work for this has been done in #302, but as Gradle doesn't support Play 2.7 at all yet cannot be tested in practice (#342)

PR #302, opened 2018-07-24
Play code generation from gradle

johanandren:
Refs #276

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/Main.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/Main.scala
@@ -10,9 +10,11 @@ import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import akka.grpc.gen.javadsl.JavaBothCodeGenerator
 import akka.grpc.gen.javadsl.JavaClientCodeGenerator
 import akka.grpc.gen.javadsl.JavaServerCodeGenerator
+import akka.grpc.gen.javadsl.play.{PlayJavaBothCodeGenerator, PlayJavaClientCodeGenerator, PlayJavaServerCodeGenerator}
 import akka.grpc.gen.scaladsl.ScalaBothCodeGenerator
 import akka.grpc.gen.scaladsl.ScalaClientCodeGenerator
 import akka.grpc.gen.scaladsl.ScalaServerCodeGenerator
+import akka.grpc.gen.scaladsl.play.{PlayScalaClientCodeGenerator, PlayScalaServerCodeGenerator}
 
 // This is the protoc plugin that the gradle plugin uses
 object Main extends App {
@@ -38,6 +40,8 @@ object Main extends App {
 
   private val generateServer: Boolean = !reqLowerCase.contains("generate_server=false")
 
+  private val generatePlay: Boolean = reqLowerCase.contains("generate_play=true")
+
   val LogFileRegex = """(?:.*,)logfile=([^,]+)(?:,.*)?""".r
   private val logger = req.getParameter match {
     case LogFileRegex(path) => new FileLogger(path)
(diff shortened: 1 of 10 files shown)


chbatey:
Ping @ktoso

codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -46,18 +50,34 @@ object Main extends App {
 
   val out = {
     val codeGenerator =
-      if (languageScala) {
-        // Scala
-        if (generateClient && generateServer) ScalaBothCodeGenerator
-        else if (generateClient) ScalaClientCodeGenerator
-        else if (generateServer) ScalaServerCodeGenerator
-        else throw new IllegalArgumentException("At least one of generateClient or generateServer must be enabled")
+      if (!generatePlay) {
```
jrudolph: Let's hope we don't get any more dimensions ;)

codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaBothCodeGenerator.scala
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.gen.javadsl.play
+
+object PlayJavaBothCodeGenerator extends PlayJavaServerCodeGenerator with PlayJavaClientCodeGenerator {
+
```
jrudolph: Shouldn't we override the name here as well?

codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -46,18 +50,34 @@ object Main extends App {
...
+        if (languageScala) {
+          // Scala
+          if (generateClient && generateServer) PlayScalaClientCodeGenerator
+          else if (generateClient) PlayScalaClientCodeGenerator
+          else if (generateServer) PlayScalaServerCodeGenerator
+          else throw new IllegalArgumentException("At least one of generateClient or generateServer must be enabled")
+        } else {
+          // Java
+          if (generateClient && generateServer) PlayJavaBothCodeGenerator
+          else if (generateClient) PlayJavaClientCodeGenerator
```
raboof: If I'm reading this right, we're running *either* `JavaClientCodeGenerator` *or* `PlayJavaClientCodeGenerator` here. For sbt, I think we're running `JavaClientCodeGenerator` and *optionally also* `PlayJavaClientCodeGenerator`: the code generated by `PlayJavaClientCodeGenerator` depends on the code generated by `JavaClientCodeGenerator`, so running only `PlayJavaClientCodeGenerator` might not be sufficient.

I think we have 2 ways to approach this: 
* invoking this `Main` twice from the Gradle plugin, once with `generate_play` and once without
* invoking this `Main` once from the Gradle plugin, and generate both the 'plain' client and the play-specific code at once.

I think the second approach is ultimately the most appealing one, as it allows eventually splitting off the play-specific portions.

For MVP, however, the first approach is likely the most easy/practical. It might even be as simple as introducing a `object CombinedPlayJavaClientCodeGenerator extends PlayJavaClientCodeGenerator with JavaClientCodeGenerator`. We should make sure this is a private generator though, to avoid people accidentally using it from sbt.

chbatey:
Is this ready for re-review @raboof ?

raboof:
Sorry, no, I only rebased it, my review comments still need to be addressed.

codegen/src/main/scala/akka/grpc/gen/Main.scala
ktoso: Let's go with the good enough one, please ticketify the other style

codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaBothCodeGenerator.scala
ktoso: I'll do that

codegen/src/main/scala/akka/grpc/gen/Main.scala
ktoso: Not entirely sure how to do this in the plugin to be honest... Let's chat and I can PR and merge it later on on top of this. @raboof

raboof:
I tested this with the example project from https://github.com/playframework/akka-grpc-play-quickstart-java . For this to work I had to install Gradle from source with `./gradlew -Djava9Home=/usr/lib/jvm/java-9-openjdk-amd64/ install -Pgradle_installPath=/home/aengelen/dev/mygradle` from https://github.com/gradle/gradle/pull/6374 . This only worked with JDK8, not JDK9, because of the way we propagate the project version (https://stackoverflow.com/questions/21693547/get-a-custom-gradle-plugins-version-from-its-own-code). It did not seem to trigger generating code at all, so looking into that.

codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -85,4 +85,16 @@ object Main extends App {
...
+   * In other build tools (Maven and sbt) these are passed to protoc separately, but since Gradle does the
+   * akka-grpc code generation in a single protoc invocation we need to combine those here:
+   */
+  object CombinedPlayJavaClientCodeGenerator extends PlayJavaClientCodeGenerator with JavaClientCodeGenerator
+  object CombinedPlayJavaServerCodeGenerator extends PlayJavaServerCodeGenerator with JavaServerCodeGenerator
+  object CombinedPlayJavaBothCodeGenerator extends PlayJavaClientCodeGenerator with JavaClientCodeGenerator with PlayJavaServerCodeGenerator with JavaServerCodeGenerator
+  object CombinedPlayScalaClientCodeGenerator extends PlayScalaClientCodeGenerator with ScalaClientCodeGenerator
+  object CombinedPlayScalaServerCodeGenerator extends PlayScalaServerCodeGenerator with ScalaServerCodeGenerator
+  object CombinedPlayScalaBothCodeGenerator extends PlayScalaClientCodeGenerator with ScalaClientCodeGenerator with PlayScalaServerCodeGenerator with ScalaServerCodeGenerator
 }
```
ktoso: should all those `override def name`?

raboof: I'm not sure, as only one is executed here anyway - but can't hurt to be sure

PR #303, opened 2018-07-24
Lookup service discovery mechanism by name

### LINKED ISSUE #281: ServiceDiscovery method in GrpcClientSettings
`def apply(config: Config)` should load ServiceDiscovery instance from configured method.
See https://github.com/akka/akka-management/pull/230

After https://github.com/akka/akka-grpc/pull/260

chbatey:
* Documentation for GrpcClientSettings
* Clarify between a "clientName" for looking up a particular client's
configuration and "serviceName" for the name to look up in service
discovery for a service's host/port

Closes #281 

Note I made it illegal to create a client form configuration that doesn't exist. I don't think that makes sense with the distinction between clientName and serviceName WDYT?

### DIFF
--- a/codegen/src/main/twirl/templates/PlayJava/ClientProvider.scala.txt
+++ b/codegen/src/main/twirl/templates/PlayJava/ClientProvider.scala.txt
@@ -33,6 +33,6 @@ public class @{service.name}ClientProvider implements Provider<@{service.name}Cl
   }
 
   public @{service.name}Client get() {
-    return @{service.name}Client.create(GrpcClientSettings.create(@{service.name}.name, sys), mat, sys.dispatcher());
+    return @{service.name}Client.create(GrpcClientSettings.fromConfig(@{service.name}.name, sys), mat, sys.dispatcher());
   }
 }
--- a/docs/src/main/paradox/client.md
+++ b/docs/src/main/paradox/client.md
@@ -156,7 +156,82 @@ Maven
 mvn akka-grpc:generate compile exec:java -Dexec.mainClass=io.grpc.examples.helloworld.GreeterClient
 ```
 
+### Configuration 
 
+A gRPC client is configured with a `GrpcClientSettings` instance. There are a number of ways of creating one and the API
+docs are the best reference. 
+
+The simplest way to create a client is to provide a static host and port.
+
+Scala
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly.scala) { #simple }
+
+Java
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #simple }
+
+Further settings can be added via the `with` methods
+
+Scala
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly.scala) { #simple-programmatic }
+
+Java
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #simple-programmatic }
+
+Instead a client can be defined in configuration. All client configurations need to be under `akka.grpc.client`
+
+Scala
+:  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #client-config }
+
+Java
+:  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #client-config }
+
+Clients defined in configuration pick up defaults from `reference.conf`:
+
+Scala
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+Java
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+#### Service discovery
+
+The examples above all use a hard coded host and port for the location of the gRPC service. Alternatively 
+[Akka Service Discovery](https://developer.lightbend.com/docs/akka-management/current/discovery.html) can be used.
+This allows a gRPC client to switch between discovering services via DNS, config, Kubernetes and Consul by just changing
+the configuration.
+
+To see how to config a particular service discovery mechanism see the [Akka Service Discovery docs](https://developer.lightbend.com/docs/akka-management/current/discovery.html).
+Once it is configured a service discovery instance can either be passed into settings or the name of the mechanism
+to lookup via an ActorSystem's configuration.
+
+Scala
+:  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #config-service-discovery }
+
+Java
+:  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #config-service-discovery }
+
+The above example configures the client `project.WithConfigServiceDiscovery` to use `config` based service discovery.
+
+Then to create the `GrpcClientSettings`:
+
+Scala
+:  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #sd-settings }
+
+Java
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #sd-settings }
+
+Alternatively if a `SimpleServiceDiscovery` is available else where in your system is can be passed in:
+
+Scala
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly.scala) { #provide-sd }
+
+Java
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #provide-sd }
+
+ 
+Currently service discovery is only queried on creation of the client. A client can be automatically re-created, and go via service discovery again,
+ if a connection can't established, see the lifecycle section.
+ 
 ### Lifecycle
 
 Instances of the generated client may be long-lived and can be used concurrently.
--- a/docs/src/main/paradox/client.md
+++ b/docs/src/main/paradox/client.md
@@ -159,7 +159,7 @@ mvn akka-grpc:generate compile exec:java -Dexec.mainClass=io.grpc.examples.hello
 ### Configuration 
 
 A gRPC client is configured with a `GrpcClientSettings` instance. There are a number of ways of creating one and the API
-docs are the best reference. 
+docs are the best reference. An `ActorSystem` is always required as it is used for default configuration and service discovery.
 
 The simplest way to create a client is to provide a static host and port.
 
@@ -195,14 +195,13 @@ Java
 
 #### Service discovery
 
-The examples above all use a hard coded host and port for the location of the gRPC service. Alternatively 
-[Akka Service Discovery](https://developer.lightbend.com/docs/akka-management/current/discovery.html) can be used.
+The examples above all use a hard coded host and port for the location of the gRPC service which is the default if you do not configure a `service-discovery-mechanism`.
+Alternatively [Akka Service Discovery](https://developer.lightbend.com/docs/akka-management/current/discovery.html) can be used.
 This allows a gRPC client to switch between discovering services via DNS, config, Kubernetes and Consul by just changing
 the configuration.
 
 To see how to config a particular service discovery mechanism see the [Akka Service Discovery docs](https://developer.lightbend.com/docs/akka-management/current/discovery.html).
-Once it is configured a service discovery instance can either be passed into settings or the name of the mechanism
-to lookup via an ActorSystem's configuration.
+Once it is configured a service discovery mechanism name can either be passed into settings or put in the client's configuration.
 
 Scala
 :  @@snip [GrpcClientSettingsSpec]($root$/../runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala) { #config-service-discovery }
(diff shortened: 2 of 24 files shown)


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -21,16 +21,25 @@ import scala.concurrent.duration.{ Duration, _ }
...
 object GrpcClientSettings {
 
-  /** Scala API */
+  /**
+   * Scala API
+   *
+   * Create a client hat uses a static host and port. Does not load
+   * any default configuration from reference.conf
+   */
   def apply(host: String, port: Int): GrpcClientSettings =
```
johanandren: I wonder if we should really keep this, maybe we should always require either an ActorSystem or a config so that we can load defaults. I think a few of the settings in Akka and Akka HTTP work like that.

docs/src/main/paradox/client.md
```diff
@@ -156,7 +156,82 @@ Maven
...
+
+Scala
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+Java
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+#### Service discovery
+
+The examples above all use a hard coded host and port for the location of the gRPC service. Alternatively
```
johanandren: " which is the default if you do not configure a `service-discovery-mechanism`." or something like that?

docs/src/main/paradox/client.md
```diff
@@ -156,7 +156,82 @@ Maven
...
+
+Scala
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly.scala) { #provide-sd }
+
+Java
+:  @@snip [GrpcClientSettingsCompileOnly]($root$/../runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #provide-sd }
+
+ 
+Currently service discovery is only queried on creation of the client. A client can be automatically re-created, and go via service discovery again,
+ if a connection can't established, see the lifecycle section.
```
johanandren: "can't be established"

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
chbatey: Should we make it implicit?

chbatey:
Updated to require an ActorSystem to create a settings

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -21,50 +21,82 @@ import scala.concurrent.duration.{ Duration, _ }
...
+   * Scala API
+   *
+   * Create a client hat uses a static host and port. Default configuration
+   * is loaded from reference.conf
+   */
+  def apply(host: String, port: Int)(implicit actorSystem: ActorSystem): GrpcClientSettings = {
+    // default is hardcoded
+    val defaultServiceConfig = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
+      .withValue("service-name", ConfigValueFactory.fromAnyRef(host))
+      .withValue("port", ConfigValueFactory.fromAnyRef(port))
```
johanandren: Shouldn't we look for that akka.grpc.client."host" first and only use '*' if that doesn't exist?

chbatey: We could. I don't know if it is confusing to treat a `host` as a `clientName`. Atm we only look for a specific config section for the constructor that takes in a `clientName`

johanandren: Ok, let's keep it like that then.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -85,29 +117,34 @@ object GrpcClientSettings {
...
+   *
+   * Configure client via the provided Config. See reference.conf for configuration properties.
    */
-  def create(config: Config): GrpcClientSettings =
-    GrpcClientSettings(config)
+  def create(config: Config, sys: ActorSystem): GrpcClientSettings = apply(config, sys)
 
   /** Java API */
-  def create(host: String, port: Int): GrpcClientSettings = apply(host, port)
+  def create(host: String, port: Int)(implicit actorSystem: ActorSystem): GrpcClientSettings = apply(host, port)
```
johanandren: No implicit in java land?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -85,29 +117,34 @@ object GrpcClientSettings {
...
-   * @param serviceName Name of the service to look up in the ServiceDiscovery
-   * @param defaultPort Port to use if service discovery only return a host
-   * @param resolveTimeout Passed to service discovery resolve
+   * @param serviceName    Name of the service to look up in the ServiceDiscovery
+   * @param defaultPort    Port to use if service discovery only return a host
+   * @param serviceDiscoveryMechanism Service discovery mechanism to use. Must be correctly configured in the ActorSystem's config
    */
-  def create(serviceName: String, defaultPort: Int, serviceDiscovery: SimpleServiceDiscovery, resolveTimeout: java.time.Duration) =
-    apply(serviceName, defaultPort, serviceDiscovery, resolveTimeout.asScala)
+  def create(serviceName: String, defaultPort: Int, serviceDiscoveryMechanism: String)(implicit actorSystem: ActorSystem) =
```
johanandren: and again

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
chbatey: Ok

chbatey:
Ping @ktoso

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -21,50 +21,82 @@ import scala.concurrent.duration.{ Duration, _ }
...
-  def apply(host: String, port: Int): GrpcClientSettings =
-    // hardcoded doesn't use the resolve timeout
-    new GrpcClientSettings(host, hardcodedServiceDiscovery(host, port), port, 1.second)
+  /**
+   * Scala API
+   *
+   * Create a client hat uses a static host and port. Default configuration
+   * is loaded from reference.conf
+   */
+  def apply(host: String, port: Int)(implicit actorSystem: ActorSystem): GrpcClientSettings = {
```
jrudolph: I wonder if it would make sense not to overload `apply`/`create` but create more descriptive names here. IMO that would make the API better readable and also help with discovering the different alternatives for connecting to services.

E.g. `GrpcClientSettings.connectToServiceAt(host, port)`, `GrpcClientSettings.discoverService(name)`

WDYT?

chbatey: I was thinking about this as well, I prefer it. Will update.

docs/src/main/paradox/client.md
```diff
@@ -156,7 +156,81 @@ Maven
...
+Scala
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+Java
+:  @@snip [reference]($root$/../runtime/src/main/resources/reference.conf) { #defaults }
+
+#### Service discovery
+
+The examples above all use a hard coded host and port for the location of the gRPC service which is the default if you do not configure a `service-discovery-mechanism`.
+Alternatively [Akka Service Discovery](https://developer.lightbend.com/docs/akka-management/current/discovery.html) can be used.
```
ktoso: hm technically we call it Akka Discovery, perhaps good to keep that name for now, in case we grep rename later on?

docs/src/main/paradox/client.md
```diff
@@ -156,7 +156,81 @@ Maven
...
+This allows a gRPC client to switch between discovering services via DNS, config, Kubernetes and Consul by just changing
```
ktoso: maybe some "and others"?

interop-tests/src/test/scala/io/grpc/testing/integration/test/AkkaGrpcScalaClientTester.scala
```diff
@@ -26,10 +27,11 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
 import scala.util.Failure
 import scala.util.control.NoStackTrace
 
-class AkkaGrpcScalaClientTester(val settings: Settings)(implicit mat: Materializer, ex: ExecutionContext) extends ClientTester {
+class AkkaGrpcScalaClientTester(val settings: Settings)(implicit mat: Materializer, as: ActorSystem) extends ClientTester {
```
ktoso: nitpick: sys or system would be nicer everywhere... `as` always is a bit weird since `ass` :P but maybe that's just me...

project/Dependencies.scala
```diff
@@ -8,7 +8,7 @@ object Dependencies {
   object Versions {
     val akka = "2.5.14"
     val akkaHttp = "10.1.3"
-    val akkaDiscovery = "0.15.0"
+    val akkaDiscovery = "0.17.0"
```
ktoso: :)

runtime/src/main/resources/reference.conf
```diff
@@ -1,13 +1,29 @@
...
   ssl-config = ${ssl-config}
+
   # TODO: Enforce HTTP/2 TLS restrictions: https://tools.ietf.org/html/draft-ietf-httpbis-http2-17#section-9.2
   connection-attempts = -1
+
+  # Service discovery mechamism to use. The default is to use a hard coded host
+  # and port that will be resolved via DNS.
+  # Any of the mechanisms described [here](https://developer.lightbend.com/docs/akka-management/current/discovery.html) can be used
+  # including Kubernetes, Consul, AWS API
+  service-discovery-mechanism = "hardcoded"
```
ktoso: good :)

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
ktoso: sounds nice to me, leaving final call to @chbatey

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
ktoso: Keeping is fine

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -21,50 +21,82 @@ import scala.concurrent.duration.{ Duration, _ }
...
 
-    GrpcClientSettings(serviceConfig)
+    GrpcClientSettings(clientConfig)
   }
 
-  /** Scala API */
-  def apply(config: Config): GrpcClientSettings = {
-    GrpcClientSettings(config getString "host", config getInt "port")
+  /**
+   * Scala API
```
ktoso: Please avoid single line "... API" lines as first line since this shows up as "highlighted" line in scaladoc, so its not very useful; `Scala API: Configure client via the provided Config. See reference.conf for configuration properties.` is nicer

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -140,8 +177,12 @@ object GrpcClientSettings {
 
 }
 
+/**
+ * @param serviceName      Service name to lookup in serviceDiscovery.
```
ktoso: no docs other then the param?

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
```diff
@@ -0,0 +1,44 @@
...
+import akka.grpc.GrpcClientSettings;
+
+import java.time.Duration;
+
+public class GrpcClientSettingsCompileOnly {
+    public static void sd() {
+
+        ActorSystem actorSystem = ActorSystem.create();
+        //#simple
+        GrpcClientSettings.create("localhost", 443, actorSystem);
```
ktoso: yeah i think the nicer names could be nice now that i look at it hm...

runtime/src/main/resources/reference.conf
```diff
@@ -1,13 +1,29 @@
+//#defaults
 akka.grpc.client."*" {
-  host = ""
+  # host to use if service-discovery-mechaism is hardcoded
```
jrudolph: "mechanism"

runtime/src/main/resources/reference.conf
jrudolph: Could we make it work like this:

```
service-discovery = {
  method = "static"
  host = "xyz"
  port = 0 # or 8080
}
```

or 

```
service-discovery = {
  method = "akka-dns" # or even class name of service discovery
  service-name = "xyz"
}
```

jrudolph:
Does this one include #311 or parts thereof?

docs/src/main/paradox/client.md
chbatey: Hrmm we seem to use Service Discovery and Akka Discovery in equal measure e.g. https://developer.lightbend.com/docs/akka-management/current/discovery.html

I'll go with Akka Discovery and make note to update Akka Management docs

ktoso: Okey, sounds good :) thanks for checking!

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
chbatey: It is a private constructor

runtime/src/main/resources/reference.conf
chbatey: static is a much nicer name :+1:

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
ktoso: ah sorry

runtime/src/main/resources/reference.conf
ktoso: nice one, agreed!

chbatey:
311 was branched from this, will rebase it once this is merged

runtime/src/main/resources/reference.conf
chbatey: Also updated to look for host if it is static otherwise service name :+1:

codegen/src/main/twirl/templates/PlayJava/ClientProvider.scala.txt
```diff
@@ -33,6 +33,6 @@ public class @{service.name}ClientProvider implements Provider<@{service.name}Cl
   }
 
   public @{service.name}Client get() {
-    return @{service.name}Client.create(GrpcClientSettings.create(@{service.name}.name, sys), mat, sys.dispatcher());
+    return @{service.name}Client.create(GrpcClientSettings.fromConfig(@{service.name}.name, sys), mat, sys.dispatcher());
```
ktoso: naisu

PR #304, opened 2018-07-25
Copy-pasta-miss with the extra generators for Play serverside

johanandren:


johanandren:
Yeah, couldn't figure out how to do that though since we don't have an actual play project for interop testing (and that I couldn't figure out how to do earlier while doing play work)

PR #305, opened 2018-07-26
WIP: Change Play router prefix behavior

richdougherty:
I started writing some Play tests in order to test SSL but I ran into some problems with the way we're generating Play routers at the moment. The main issue was the lack of a slash for the router's prefix, but that led to a few other changes.

1. Change the generated router to default to "/" + the gRPC service name instead of just the gRPC service name. Play routers need to have a prefix starting with slash to work properly. Usually a router defaults to an empty "/" prefix, but I think we can let this slide so that users can include gRPC routers without needing to give a prefix, i.e.

    ```
    # No need to give the service name in the routes file if the router knows its own prefix
    -> / my.grpc.Router
    ```

2. Change the withPrefix method to properly prepend the new prefix to the existing prefix. While it's not recommended to change the prefix to be anything other than the gRPC service name, if we are going to support the `withPrefix` method we should do it consistently with other parts of Play.  **STILL TO DO**

3. Change the generated handler partial method to support full path instead of a prefix of just a single path segment. This is needed because of change 2.

These changes will make the router behave more like normal Play routers with the exception of point 1. The reason for point 1, as mentioned before, is to make it easier to include gRPC routers inside other routers. It saves the user typing out the prefix.

Thoughts or comments?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -167,6 +167,7 @@ lazy val playInteropTestScala = Project(
   .settings(commonSettings)
   .settings(Seq(
     ReflectiveCodeGen.reflectiveGeneratorClass := "akka.grpc.sbt.test.PlayScalaCompositeCodeGenerator",
+    fork in Test := true,
   ))
   .enablePlugins(akka.grpc.NoPublish)
   .enablePlugins(akka.grpc.ReflectiveCodeGen)
@@ -183,6 +184,7 @@ lazy val playInteropTestJava = Project(
   .settings(commonSettings)
   .settings(Seq(
     ReflectiveCodeGen.reflectiveGeneratorClass := "akka.grpc.sbt.test.PlayJavaCompositeCodeGenerator",
+    fork in Test := true,
   ))
   .enablePlugins(akka.grpc.NoPublish)
   .enablePlugins(akka.grpc.ReflectiveCodeGen)
--- a/codegen/src/main/twirl/templates/PlayScala/Router.scala.txt
+++ b/codegen/src/main/twirl/templates/PlayScala/Router.scala.txt
@@ -15,13 +15,17 @@ import akka.stream.Materializer
 
 import play.api.mvc._
 import play.api.mvc.akkahttp.AkkaHttpHandler
-import play.api.routing.Router
+import play.api.routing.{Router, SimpleRouter}
 import play.api.routing.Router.Routes
 
 @@Singleton
-class @{service.name}Router @@Inject()(impl: controllers.@{service.name}Impl)(implicit mat: Materializer)
-  extends Router {
-  protected val prefix = @{service.name}.name
+class @{service.name}Router(impl: controllers.@{service.name}Impl, prefix: String)(implicit mat: Materializer)
+  extends SimpleRouter {
+  require(prefix != null && !prefix.isEmpty && prefix.charAt(0) == '/')
+
+  @@Inject()
+  def this(impl: controllers.@{service.name}Impl)(implicit mat: Materializer) =
+    this(impl, "/" + @{service.name}.name)
 
   protected val handler = new AkkaHttpHandler {
     val h = @{service.name}Handler(impl, prefix)
@@ -37,20 +41,7 @@ class @{service.name}Router @@Inject()(impl: controllers.@{service.name}Impl)(im
 
   override def withPrefix(newPrefix: String): Router = {
     if (newPrefix == "/") this
-    else new @{service.name}Router(impl) {
-      override val prefix = newPrefix
-
-      // Needs to be overridden to point to new prefix due to initialization order
-      override val handler = new AkkaHttpHandler {
-        val h = @{service.name}Handler(impl, newPrefix)
-        override def apply(request: HttpRequest): Future[HttpResponse] = h(request)
-      }
-
-      override def withPrefix(additionalPrefix: String): Router = {
-        if (prefix == "/") this
-        else throw new IllegalStateException("Only one level of prefixes is supported for gRPC")
-      }
-    }
+    else new @{service.name}Router(impl, newPrefix)
   }
 
 }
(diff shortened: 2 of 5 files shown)


build.sbt
```diff
@@ -167,6 +167,7 @@ lazy val playInteropTestScala = Project(
   .settings(commonSettings)
   .settings(Seq(
     ReflectiveCodeGen.reflectiveGeneratorClass := "akka.grpc.sbt.test.PlayScalaCompositeCodeGenerator",
+    fork in Test := true,
```
richdougherty: Too many OOMEs!

codegen/src/main/twirl/templates/PlayScala/Router.scala.txt
```diff
@@ -15,13 +15,17 @@ import akka.stream.Materializer
...
+import play.api.routing.{Router, SimpleRouter}
 import play.api.routing.Router.Routes
 
 @@Singleton
-class @{service.name}Router @@Inject()(impl: controllers.@{service.name}Impl)(implicit mat: Materializer)
-  extends Router {
-  protected val prefix = @{service.name}.name
+class @{service.name}Router(impl: controllers.@{service.name}Impl, prefix: String)(implicit mat: Materializer)
+  extends SimpleRouter {
+  require(prefix != null && !prefix.isEmpty && prefix.charAt(0) == '/')
```
richdougherty: Being careful here. We should probably have these checks in Play routers too and mention them in the Play scaladocs.

codegen/src/main/twirl/templates/PlayScala/Router.scala.txt
```diff
@@ -15,13 +15,17 @@ import akka.stream.Materializer
...
+
+  @@Inject()
+  def this(impl: controllers.@{service.name}Impl)(implicit mat: Materializer) =
+    this(impl, "/" + @{service.name}.name)
```
richdougherty: The injected constructor provides a default prefix.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -76,11 +77,36 @@ object @{service.name}Handler {
...
+          handleRec(prefix1, path1)
+        case (_: SlashOrEmpty, Slash(Segment(method, Empty))) =>
+          println(s"$prefix0 is a prefix of $path0 - returning handler for $method")
+          Some(handle(req, method).recoverWith(GrpcExceptionHandler.default))
+        case _ =>
+          println(s"$prefix0 is NOT a prefix of $path0")
+          None
+      }
+      handleRec(prefixPath, req.uri.path)
+    }
```
richdougherty: If we agree this is the way to go then we can factor some of this into a reusable method.

project/Dependencies.scala
```diff
@@ -61,6 +62,8 @@ object Dependencies {
     val scalaTest = "org.scalatest" %% "scalatest" % Versions.scalaTest % "test" // ApacheV2
     val scalaJava8Compat = "org.scala-lang.modules" %% "scala-java8-compat" % Versions.scalaJava8Compat % "test" // BSD 3-clause
     val junit = "junit" % "junit" % "4.12" % "test" // Common Public License 1.0
+    val play = "com.typesafe.play" %% "play-test" % Versions.play // Apache M2
+    val scalaTestPlusPlay = "org.scalatestplus.play" %% "scalatestplus-play" % Versions.scalaTestPlusPlay % "test"
```
richdougherty: This is what we need to write realistic Play tests within the akka-grpc project.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -76,11 +77,36 @@ object @{service.name}Handler {
...
-    })
+    import Path._
+
+    val prefixPath: Path = {
+      val unslashed: Path = Path(prefix)
+      // Add slash automatically in the same way as Akka HTTP's `path` routing directive does.
+      unslashed match {
+        case p: Slash => p
+        case p => Slash(p)
+      }
```
richdougherty: Should we add a slash automatically as shown here? I'd prefer to make a call that the prefix either always or never has a leading slash, but I wasn't sure which way to go.

Always having a leading slash would be consistent with Play but the existing akka-grpc code is written as if the prefix is always the service name, without any leading slash.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
richdougherty: Also is there some clever pattern match I can do to match a path prefix this instead of writing a custom function?

ktoso:
OK TO TEST

ktoso:
PLS WHITELIST

build.sbt
ktoso: lol good to know!

johanandren:
Hey @richdougherty 

I think this was done against a pre play-router-merge because the slash I already had to handle. See here: https://github.com/akka/akka-grpc/blob/master/runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala#L43

I disagree about the allowing prefixing for consistency since no client supports it. Too easy for users to think they can put the service on whatever path they want and then not understand why it doesn't work. That's why I made it throw with a clear error message, here: https://github.com/akka/akka-grpc/blob/master/runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala#L67

For 3 we would need to somehow list all the methods of the service since they are mapped into the path, since it doesn't really make sense for anything else to map itself into the namespace of a specific service I don't think we need to support this either.

johanandren:
I think since we already give it the slash + service name prefix, if just `/` doesn't invoke `withPrefix` then it should work. I'll try this out a bit later today.

richdougherty:
> a pre play-router-merge

I deserve that for forgetting to review the PR yesterday!

I'm ok with the prefix being fixed as `s"/$serviceName"` if other prefixes are basically errors. I think only allowing the identity *withPrefix("/")* operation and disallowing all other *withPrefix* values will probably work.

johanandren:
Closing this in favor of #306

PR #306, opened 2018-07-26
Additional improvements for play server side

johanandren:
@richdougherty ping

### DIFF
--- a/docs/src/main/paradox/play-framework.md
+++ b/docs/src/main/paradox/play-framework.md
@@ -86,7 +86,7 @@ import akka.grpc.gen.javadsl.play.PlayJavaServerCodeGenerator
 akkaGrpcExtraGenerators += PlayJavaServerCodeGenerator
 ```
 
-The plugin will look for `.proto` service descriptors in `src/main/protobuf` and output an abstract class per service
+The plugin will look for `.proto` service descriptors in `app/protobuf` and output an abstract class per service
 that you then implement, so for example for the following protobuf descriptor:
 
 @@snip[helloworld.proto]($root$/../play-interop-test-scala/src/main/protobuf/helloworld.proto) { #protoSources }
--- a/docs/src/main/paradox/play-framework.md
+++ b/docs/src/main/paradox/play-framework.md
@@ -101,11 +101,13 @@ Scala
 Java
 :   @@snip[GreeterServiceImpl.java]($root$/../play-interop-test-java/src/main/java/controllers/GreeterServiceImpl.java) { #service-impl }
 
-And then add this to your Play `conf/routes` file, note that the path here depends on the package and service name from the
-`.proto` service descriptor and cannot be set to an arbitrary value (if you try to do so an exception will be thrown when the router is started).
+And then add the router to your Play `conf/routes` file. Note that the router alerady knows its own path since it is
+based on the package name and service name of the service and therefore the path `/` is enough to get it to end up in the right place
+(in this example the path will be `/helloworld.GreeterService`).
+It cannot be added at an arbitrary path (if you try to do so an exception will be thrown when the router is started).
 
 ```
-->     /helloworld.GreeterService   controllers.GreeterServiceController
+->     /   controllers.GreeterServiceController
 ```
 
 A gRPC client can now connect to the server and call the provided services.
--- a/docs/src/main/paradox/play-framework.md
+++ b/docs/src/main/paradox/play-framework.md
@@ -101,7 +101,7 @@ Scala
 Java
 :   @@snip[GreeterServiceImpl.java]($root$/../play-interop-test-java/src/main/java/controllers/GreeterServiceImpl.java) { #service-impl }
 
-And then add the router to your Play `conf/routes` file. Note that the router alerady knows its own path since it is
+And then add the router to your Play `conf/routes` file. Note that the router already knows its own path since it is
 based on the package name and service name of the service and therefore the path `/` is enough to get it to end up in the right place
 (in this example the path will be `/helloworld.GreeterService`).
 It cannot be added at an arbitrary path (if you try to do so an exception will be thrown when the router is started).
--- a/runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
@@ -65,7 +65,9 @@ import scala.compat.java8.OptionConverters._
    * so therefore not supported.
    */
   final override def withPrefix(prefix: String): Router =
-    if (prefix == this.prefix) this
+    // explicitly stating the full path is fine, we also treat / as identity for convenience so that users do not
+    // _have_ to repeat the service name but can just have a / in the routes file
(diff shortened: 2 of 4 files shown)


docs/src/main/paradox/play-framework.md
```diff
@@ -101,11 +101,13 @@ Scala
 Java
 :   @@snip[GreeterServiceImpl.java]($root$/../play-interop-test-java/src/main/java/controllers/GreeterServiceImpl.java) { #service-impl }
 
-And then add this to your Play `conf/routes` file, note that the path here depends on the package and service name from the
-`.proto` service descriptor and cannot be set to an arbitrary value (if you try to do so an exception will be thrown when the router is started).
+And then add the router to your Play `conf/routes` file. Note that the router alerady knows its own path since it is
```
jrudolph: "already"

jrudolph: "And then add the router" - is it a "router"? Isn't it the "controller"?

docs/src/main/paradox/play-framework.md
```diff
@@ -101,11 +101,13 @@ Scala
...
+based on the package name and service name of the service and therefore the path `/` is enough to get it to end up in the right place
+(in this example the path will be `/helloworld.GreeterService`).
+It cannot be added at an arbitrary path (if you try to do so an exception will be thrown when the router is started).
 
 ```
-->     /helloworld.GreeterService   controllers.GreeterServiceController
+->     /   controllers.GreeterServiceController
```
jrudolph: Without reading the sources I find that explanation (and how the routes file look) weird. Could the automatic behavior made more explicit instead, e.g. as `/[auto]` or something like that?

Maybe

"Then add the controller to your Play `conf/routes` file. The path to the service is fixed by the gRPC specification based on package and service name, so the route entry is a dummy that tells Play about the existence of a service implementation. You can either use the fixed service path in the routes file like this:


```
->     /helloworld.GreeterService   controllers.GreeterServiceController
```

or tell Play to figure out the path from the metadata by using `/[auto]` as the path:

```
->     /[auto]   controllers.GreeterServiceController
```

Providing a different path to access the service is not possible (because that's how gRPC works)."

WDYT?

ktoso:
Johan is now away for a few weeks so I'll take these comments on tomorrow

docs/src/main/paradox/play-framework.md
ktoso: Hmm I think it is a Router... not sure `public class GreeterServiceImpl extends AbstractGreeterServiceRouter {` in Play wording I think Router is a lower level thing simply hm

docs/src/main/paradox/play-framework.md
ktoso: Maybe indeed /[grpc] would be nice hm, I would like to avoid the plain `[auto]` though. May want opinion of @richdougherty here?

richdougherty: If using `-> /` doesn't look right we could just use `-> /helloworld.GreeterService` and have `controllers.GreeterServiceController` be a normal Play router that picks up its path from the routes file, instead of having its own path hardcoded. This is the normal way you'd do it in Play. It's a bit unusual how it's done at the moment.

Regarding `/[auto]` or `/[grpc]` I'm not sure if we want to add new routing syntax unless we really need to, since this is an MVP.

Welcome back @jrudolph. :)

richdougherty: The current solution works OK, but it feels a bit strange to have `router.withPrefix("/helloworld.GreeterService")` leave the router unchanged. The `withPrefix` method is supposed to prepend a prefix.

runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
```diff
@@ -65,7 +65,9 @@ import scala.compat.java8.OptionConverters._
    * so therefore not supported.
    */
   final override def withPrefix(prefix: String): Router =
-    if (prefix == this.prefix) this
+    // explicitly stating the full path is fine, we also treat / as identity for convenience so that users do not
+    // _have_ to repeat the service name but can just have a / in the routes file
+    if (prefix == this.prefix || prefix == "/") this
```
ktoso: Discussed with Rich and we're going to make it feel more "play", the `prefix == this.prefix` skipping is something that Rich was not happy with but the `prefix == /` is ok for this logic

docs/src/main/paradox/play-framework.md
ktoso: Discussed with Rich about what would be best for Play users etc, we're not going for the special syntax, at least not now

ktoso: Mentioned below

ktoso:
I think this is good now... please check

docs/src/main/paradox/play-framework.md
jrudolph: > Welcome back @jrudolph. :)

Thanks :)

I haven't completely understood the rationale for going with `/` instead of `-> /helloworld.GreeterService`. If we just have `-> / controllers.GreeterServiceController`, you cannot determine which path is actually handled (or which service is offered) just by looking at the route file. Isn't that what the route file is supposed to do? But maybe I'm missing some bit about the expectations of Play users :)

ktoso: @richdougherty was more worried about "magical parameter value" here really which would cause the confusion; since we do not allow mapping it over to a different URL, since grpc clients expect the service on the "known" path. The `<grpc>` would be good but we don't want to go there now.

The `/` is the usual way of allowing "if the router has a default path" so in that sense it is more expected than suprising

jrudolph: > The / is the usual way of allowing "if the router has a default path"

If that's the idiomatic way, I agree, let's go for it :)

renatocaval: I agree that we could leave that as is for now. If we decide to introduce a new syntax, I would prefer something that make it clear that the `Router` being used is already prefixed and therefore should not get any new prefix, unless explicitly required by the user. For instance:

```
// to be used for routers that have already a prefix
// (eg: gRPC and cluster http management)
-> [noprefix] controllers.GreeterServiceRouter
-> [noprefix] controllers.AkkaClusterMgmtRouter


// to be used for any we want to prepend a prefix
// that's weird in the context of gRPC, but at least we keep consistency. 
-> /abc controllers.GreeterServiceRouter
```

ps: we should really NOT call it a Controller. We had that discussion before and I think this wasn't changed yet. We don't generate a Controller, it's a Router. And we are using the router operator `->`, so we should generate it as `GreeterServiceRouter`.

ktoso: Yeah they are called routers. I would like to merge this PR, objections?

jrudolph: Will we do the Controller -> Router renaming in a follow-up?

ktoso: Ah the name here is controller... will fix that. The generated code is all Router AFAIK tho

ktoso:
Was Rich Okeyed :P

PR #311, opened 2018-07-27
Fail fast if service-name not set

chbatey:
This is just the last commit, depends on my other PR


chbatey:
Rebased and ready for review

PR #312, opened 2018-08-01
Add comment to all generated files

### LINKED ISSUE #273: All generated sources should contain a comment declaring this

chbatey:
Closes #273


codegen/src/main/scala/akka/grpc/gen/Constants.scala
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.gen
+
+object Constants {
+  val DoNotEditComment = "// Generated by Akka gRPC. DO NOT EDIT."
```
jrudolph: :+1: 

As the next step could we add the name of the proto file this was generated from (to edit instead)?

codegen/src/main/twirl/templates/PlayJava/ClientProvider.scala.txt
```diff
@@ -4,6 +4,7 @@
 
 @(service: akka.grpc.gen.javadsl.Service)
 
+@akka.grpc.gen.Constants.DoNotEditComment
```
ktoso: :)

PR #314, opened 2018-08-02
Make plugin-tester projects available from main project

jrudolph:
I revamped the ReflectiveCodeGen a bit to make its configuration more like the actual
AkkaGrpcPlugin. I could then get rid of the stubs in the sbt-plugin code.

Then, I created subprojects for `plugin-tester-{java,scala}` in the main project,
so we can directly test and compile them there.

~I created copies of the old ones in `plugin-tester-{java,scala}-standalone` which
contain the old directories to be used outside of the main project. The question is
if this is really necessary, the only reason it is, is because those projects contain their own sbt project files. If we agree that those are not necessary, we can move everything over and remove the `-standalone` versions. (The only reason
I see could be for example purposes to the actual sbt-plugin usage, however, that
doesn't quite work anyway because the code contains the source links to the plugin).~


jrudolph:
(Jenkins will fail in any case, because it tries to build the plugin-tester projects from their own directories, which isn't necessary any more)

build.sbt
```diff
@@ -214,6 +200,34 @@ lazy val docs = Project(
...
 
+lazy val pluginTesterScala = Project(
+  id = "akka-grpc-plugin-tester-scala",
+  base = file("plugin-tester-scala")
+)
+  .settings(Dependencies.pluginTester)
+  .settings(commonSettings)
+  .enablePlugins(akka.grpc.NoPublish)
+  .settings(
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package")
```
raboof: Nice ;)

project/ReflectiveCodeGen.scala
```diff
@@ -40,61 +42,72 @@ object ReflectiveCodeGen extends AutoPlugin {
...
+          |val logger = akka.grpc.gen.StdoutLogger
+          |
+          |(targetPath: java.io.File, settings: Seq[String]) => {
+          |  val generators =
+          |    AkkaGrpcPlugin.generatorsFor(sources, languages, logger) ++
+          |    Seq($extraGenerators).map(gen => AkkaGrpcPlugin.toGenerator(gen, akka.grpc.gen.StdoutLogger))
+          |  AkkaGrpcPlugin.targetsFor(targetPath, settings, generators)
+          |}
+        """.stripMargin
+    val generatorsF = tb.eval(tb.parse(source)).asInstanceOf[(File, Seq[String]) => Seq[Target]]
```
raboof: :open_mouth: cool

build.sbt
```diff
@@ -214,6 +200,34 @@ lazy val docs = Project(
...
+  )
+  .pluginTestingSettings
```
raboof: I'm not sure we really want to use the 'reflective' codegen for the `plugin-tester` projects: I found it quite nice to be able to test against a 'real' sbt project using the 'real' plugin, both on CI and during development.

jrudolph: I came to work on this, because I was reviewing documentation which takes code snippets from `plugin-tester-scala` but I couldn't readily edit those files because IDEA didn't load them correctly because of the project structure (though someone said, that it would work when opening IDEA from the subdirectory).

As you can see in the history, I tried to preserve the old directory structure next to the new one by creating new subdirectories, linking to the same sources but without the standalone sbt build files. 

> found it quite nice to be able to test against a 'real' sbt project using the 'real' plugin, both on CI and during development.

I came to the conclusion that it isn't worth it to keep the extra copy because the plugin-tester projects are no real sbt projects because they still reference the main project for loading the plugin instead of publishing it first.

> On the other hand, during development we can now/soon use the example/quickstart/demo projects for this, so it might not be a big deal anymore.

Exactly.

raboof: > couldn't readily edit those files because IDEA didn't load them correctly

Yes that is annoying

> still reference the main project for loading the plugin instead of publishing it first

You're right - I thought we had to `publishLocal`/`publishM2` and then use `-Dplugin.version`, but that goes for scripted tests and non-sbt builds, not for these projects.

>> On the other hand, during development we can now/soon use the example/quickstart/demo projects for this, so it might not be a big deal anymore.
> Exactly.

Agreed then!

jrudolph:
Did another rebase and squashed into reasonably sized standalone commits.

jrudolph:
Fix a conflict wrt the recently added `proto` directory.

build.sbt
```diff
@@ -116,19 +118,11 @@ lazy val interopTests = Project(
...
-    // needed explicitly as we don't directly depend on the codegen project
-    watchSources ++= (watchSources in codegen).value,
-    // yeah ugly, but otherwise, there's a circular dependency between the project values
-    watchSources ++= (watchSources in ProjectRef(file("."), "sbt-akka-grpc")).value,
-  )
-  .dependsOn(runtime)
-  .enablePlugins(akka.grpc.ReflectiveCodeGen)
-  // needed to be able to override the PB.generate task reliably
-  .disablePlugins(ProtocPlugin)
+    ReflectiveCodeGen.generatedLanguages := "Scala, Java",
```
ktoso: this seems a bit weird that its a string like that...? Why not a specific type that I think we have already?

jrudolph: It's because of the setup where we cannot directly reference settings from the sbt plugin but have to mimic it as far as possible.

jrudolph: So the idea is to keep this mimic'd structure as light as possible being just enough to support running the test project inside the same project.

project/ReflectiveCodeGen.scala
ktoso: lol that's some old school stuff there lol, have not seen this style of reflection since ages :)

build.sbt
ktoso: Ah I see... okey!

jrudolph:
Next rebase ;)

It's expected for the Jenkins PR validation to fail because it still tries to build plugin-tester-scala and plugin-tester-java manually.

jrudolph:
I changed akka-grpc-pr-validator not to run sbt in the plugin-tester-* projects any more. This is now done by aggregation in the main project.

PR #315, opened 2018-08-06
docs improvements while working on play sample apps

ktoso:


docs/src/main/paradox/play-framework.md
```diff
@@ -28,15 +28,21 @@ With the following `helloworld.proto` file:
 
 @@snip[helloworld.proto]($root$/../play-interop-test-scala/src/main/protobuf/helloworld.proto) { #protoSources }
 
-The module file is generated in `example.myapp.helloworld.grpc.helloworld.AkkaGrpcClientModule`.
+The module file is generated in @scala[`example.myapp.helloworld.grpc.helloworld.AkkaGrpcClientModule` by default for Scala]
+@java[`example.myapp.helloworld.grpc.AkkaGrpcClientModule` by default for Java], which corresponds to the default value
```
ktoso: the package is different since the flat_package setting is different for the langs

play-interop-test-java/src/main/java/controllers/MyController.java
```diff
@@ -2,6 +2,7 @@
  * Copyright (C) 2009-2018 Lightbend Inc. <https://www.lightbend.com>
  */
 
+// #using-client
```
ktoso: include those in samples so it's less confusing where imports come from

docs/src/main/paradox/play-framework.md
ktoso: I think we may have messed it up... digging... The package was definitely different and not the same on default settings and docs which i was following

raboof: for `play-interop-test-scala` the package is different because there we don't use the plugin, but we use `ReflectiveCodeGen`, which sets `flat_package` back to `false`.

Perhaps we should avoid being so subtle and just split up our interop-test subproject into separate scala/java/common projects... more subprojects, but less complicated ones.

ktoso: Would be nice I agree (the splitting up into packages)

ktoso:
I'll merge for now and will likely follow up with a smaller one today later on once done with examples

PR #316, opened 2018-08-06
Allow TLS to be turned off in config

chbatey:


runtime/src/main/resources/reference.conf
```diff
@@ -19,6 +19,7 @@ akka.grpc.client."*" {
   user-agent = ""
   # Pulls default configuration from ssl-config-core's reference.conf
   ssl-config = ${ssl-config}
+  use-tls = true
```
ktoso: quickly add some mini docs? :)

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -91,7 +91,8 @@ object GrpcClientSettings {
         deadline = getPotentiallyInfiniteDuration(clientConfiguration, "deadline"),
         userAgent = getOptionalString(clientConfiguration, "user-agent"),
         sslContext = getOptionalSSLContext(clientConfiguration, "ssl-config"),
-        connectionAttempts = getOptionalInt(clientConfiguration, "connection-attempts"))
+        connectionAttempts = getOptionalInt(clientConfiguration, "connection-attempts"),
+        useTls = clientConfiguration.getBoolean("use-tls"))
```
ktoso: Ah okey since it defaulted to true; Thanks!

ktoso:
Want to try this locally, merging; Following up on docs would be nice some day

Issue #308, opened 2018-07-26
sbt docs for Java shows wrong setting  


johanandren:
Shows `akkaGrpcTargetLanguages` but is now `akkaGrpcGeneratedLanguages`

ennru:
Fixed with #317.

PR #317, opened 2018-08-06
Use correct sbt configuration key in docs

raboof:
akkaGrpcGeneratedLanguages rather than akkaGrpcTargetLanguages

Refs #308


docs/src/main/paradox/sbt.md
```diff
@@ -20,15 +20,15 @@ akkaGrpcGeneratedSources := Seq(AkkaGrpc.Server)
 What language to generate stubs for is also configurable:
 ```
 // default is Scala only
-akkaGrpcTargetLanguages := Seq(AkkaGrpc.Scala)
```
jrudolph: Can we take those snippets from `scripted` tests or from a simple file in `plugin/src/test` that is compiled?

ktoso: Would be nice indeed :)

PR #325, opened 2018-08-07
Fix gradle build problems on Travis

### LINKED ISSUE #322: Gradle on travis fails
https://travis-ci.org/akka/akka-grpc/jobs/412688002:
```
FAILURE: Build failed with an exception.
* Where:
Build file '/home/travis/build/akka/akka-grpc/gradle-plugin/build.gradle' line: 22
* What went wrong:
A problem occurred evaluating project ':akka-grpc-gradle-plugin'.
> String index out of range: -1
```
It would seem like `"git describe --tags".execute().text` is returning an empty string

jrudolph:
Fixes #322.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,7 @@ cache:
   directories:
     - $HOME/.ivy2/cache
     - $HOME/.sbt/boot
+    - $HOME/.gradle
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ language: scala
 scala:
   - 2.12.6
 
+# make sure to fetch enough commits, so that git describe still works
+git:
+  depth: false
+
 jobs:
   include:
     - stage: tests


raboof:
Is that to diagnose #322? I'm pretty sure it's the git command producing an empty string... curious why though

jrudolph:
> Is that to diagnose #322? I'm pretty sure it's the git command producing an empty string... curious why though

No idea, I also don't know enough about gradle to add real diagnostic output.

The stacktrace is this:

```
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java_lang_String$substring.call(Unknown Source)
	at build_1vs261n3s21r3usm2obayf5hm.run(/home/travis/build/akka/akka-grpc/gradle-plugin/build.gradle:22)
	at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory.java:90)
	... 98 more
```

raboof:
Yeah and it mentions:
```
* Where:
Build file '/home/travis/build/akka/akka-grpc/gradle-plugin/build.gradle' line: 22
```

jrudolph:
I added `git describe --tags` to the travis build script and now we get

```
Before describe
fatal: No names found, cannot describe anything.
```

So maybe tags are not fetched?

jrudolph:
Ah....

travis clones with `git clone --depth=50` and v0.1 just moved outside of that depth...

jrudolph:
When you clone with `--depth=50` there will an entry in `.git/shallow` which cuts off history add that point, so even if you look at history with `tig` or other tools, the older commits are suddenly missing even if you fetched them later on and they actually exist in your repo.

raboof:
wow that explains :D. Damn.

jrudolph:
Pushed a fix, let's see if it works ;)

raboof:
We should just release more often, right? :laughing:

jrudolph:
That would be the boring resolution ;)

jrudolph:
Tests green, please review.

PR #348, opened 2018-08-15
fix #285 parallelism must be set explicitly

### LINKED ISSUE #285: Opposite streams block each other
I don't know if the following is a bug or what.

I have a situation where a long-lived stream is used to stream calculated data from client to server.
With the same client I have a second stream, which is used to retrieve data from the server.

The problems arises when the first stream is open and is not sending anything, it blocks the second stream from retrieving data.

Sincerely,
Peter

ktoso:
https://github.com/akka/akka-grpc/issues/285


chbatey:
Hrmm not sure about this guy, no CLA

ktoso:
https://www.lightbend.com/contribute/cla/check/ktoso 😉

PR #360, opened 2018-08-21
Upgrade to akka http 10.1.4

chbatey:
Will merge the TLS PR after this then do the release

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -7,7 +7,7 @@ object Dependencies {
 
   object Versions {
     val akka = "2.5.14"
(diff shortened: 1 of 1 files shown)


raboof:
Can/should we also revert the changes in https://github.com/akka/akka-grpc/pull/354/files with this?

chbatey:
Yep, i'll add that

chbatey:
@raboof i've taken out the reverting you asked for due to https://github.com/akka/akka-http/pull/2165

10.1.4 still contains the java dsl fixes so shall we merge this and do a release with the TLS removed from samples PRs as well?

PR #366, opened 2018-08-23
Generate code before running paradox

raboof:
Because we refer to the generated code from snippets


ktoso:
Sounds good 👍

jrudolph:
Ah, I suspected that it could fail but didn't really check thoroughly. Thanks for fixing!

PR #369, opened 2018-08-23
Move AkkaGrpcClient out of 'internal'

raboof:


dwijnand:
What's the backstory with this PR?

raboof:
Indeed this could have used some clarification ;).

In the context of the play-testkit, it turned out that this common supertype for all generated clients was helpful, e.g. in https://github.com/akka/akka-grpc/blob/master/play-testkit/src/main/scala/akka/grpc/play/AkkaGrpcClientHelpers.scala - so that suggested it might make sense to take it out of 'internal' and make it part of the non-'internal' API.

I'm less sure about the `AkkaGrpcClientFactory` - perhaps it would make sense to move that one into play-testkit actually.

dwijnand:
Superseded by #406.

PR #378, opened 2018-08-30
WIP: Play routers and services

### LINKED ISSUE #364: Seperate out service impl and Play router
Current code generation creates an abstract class that is both a Play router and a service class. This makes it easier to wire together the application (no need to configure a binding for the service implementation) but it's not the usual way that Play apps are written. Usually the router is one class and the controller (with action logic) is another. I think it would make sense if we followed this pattern for gRPC code generation too - have the router as one class and the service implementation as another.

Example code snippets taken from codegen as at 21 August 2018:

```scala
abstract class AbstractGreeterServiceRouter(mat: Materializer) extends PlayRouter(mat, GreeterService.name) with GreeterService {
  final override def createHandler(serviceName: String, mat: Materializer): HttpRequest => Future[HttpResponse] =
    GreeterServiceHandler(this, serviceName)(mat)
}
```

```scala
@Singleton
class GreeterServiceImpl @Inject() (implicit mat: Materializer) extends AbstractGreeterServiceRouter(mat) {
  override def sayHello(in: HelloRequest): Future[HelloReply] = Future.successful(HelloReply(s"Hello, ${in.name}!"))
}
```

richdougherty:
Fixes #364.

This changes the way that Play gRPC services are generated so that the service implementation and router are in separate classes. I feel this is more idiomatic for Play. The service implementation isn't generated - the user has to do that - so I've worked to make the log and error messages informative and helpful.

At the moment this PR is in a rough state. I've made the main changes and got the tests passing. I still need to go through the code and remove duplication, moving files around and add scaladoc and javadoc.

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
@@ -4,9 +4,14 @@
 
 package akka.grpc.gen
 
+import akka.grpc.gen.javadsl.Service
+import akka.grpc.gen.scaladsl.Service
+import akka.grpc.gen.scaladsl.play.PlayScalaClientCodeGenerator
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse
 import protocbridge.Artifact
+
+import scala.annotation.tailrec
 
 /**
  * Code generator trait that is not directly bound to scala-pb or protoc (other than the types).
@@ -22,4 +27,29 @@ trait CodeGenerator {
 
   final def run(request: Array[Byte], logger: Logger): Array[Byte] =
     run(CodeGeneratorRequest.parseFrom(request), logger: Logger).toByteArray
+}
+
+private[gen] object CodeGenerator {
+
+  /** Extract the longest common package prefix for a list of packages. */
+  private[gen] def commonPackage(packages: Seq[String]): String =
+    packages.reduce(commonPackage(_, _))
+
+  /** Extract the longest common package prefix for two packages. */
+  private[gen] def commonPackage(a: String, b: String): String = {
+    if (a == b) a else {
+      val aPackages = a.split('.')
+      val bPackages = b.split('.')
+      @tailrec
+      def countIdenticalPackage(pos: Int): Int = {
+        if (aPackages(pos) == bPackages(pos)) countIdenticalPackage(pos + 1)
+        else pos
+      }
+
+      val prefixLength = countIdenticalPackage(0)
+      if (prefixLength == 0) "" // no common, use root package
+      else aPackages.take(prefixLength).mkString(".")
+    }
+  }
+
 }
--- a/play-interop-test-java/src/test/scala/akka/grpc/gen/PlayJavaRouterSpec.scala
+++ b/play-interop-test-java/src/test/scala/akka/grpc/gen/PlayJavaRouterSpec.scala
@@ -16,7 +16,7 @@ import akka.stream.scaladsl.{ Sink, Source }
 import akka.stream.{ ActorMaterializer, Materializer }
 import akka.util.ByteString
 import controllers.GreeterServiceImpl
-import example.myapp.helloworld.grpc.{ GreeterService, HelloReply, HelloRequest }
+import example.myapp.helloworld.grpc.{ GreeterService, GreeterServiceRouter, HelloReply, HelloRequest }
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.{ BeforeAndAfterAll, Matchers, WordSpec }
 import play.api.inject.SimpleInjector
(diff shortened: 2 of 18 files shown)


codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
```diff
@@ -23,3 +28,28 @@ trait CodeGenerator {
   final def run(request: Array[Byte], logger: Logger): Array[Byte] =
     run(CodeGeneratorRequest.parseFrom(request), logger: Logger).toByteArray
 }
+
+private[gen] object CodeGenerator {
+
+  /** Extract the longest common package prefix for a list of packages. */
+  private[gen] def commonPackage(packages: Seq[String]): String =
+    packages.reduce(commonPackage(_, _))
```
richdougherty: The old code had an optimisation to check everything for equality. I've moved that optimisation out of this method and into the `commonPackage(String,String)` method. That means we can use `reduce` instead of writing our own fold.

play-interop-test-java/src/test/scala/akka/grpc/gen/PlayJavaRouterSpec.scala
```diff
@@ -50,7 +50,7 @@ class PlayJavaRouterSpec extends WordSpec with Matchers with BeforeAndAfterAll w
   implicit def fromSourceMarshaller[T](implicit serializer: ProtobufSerializer[T], mat: Materializer, codec: Codec): ToResponseMarshaller[Source[T, NotUsed]] =
     Marshaller.opaque((response: Source[T, NotUsed]) ⇒ GrpcMarshalling.marshalStream(response)(serializer, mat, codec))
 
-  val router = new GreeterServiceImpl(mat)
+  val router = new GreeterServiceRouter(new GreeterServiceImpl(), mat)
```
richdougherty: Here you can see how a new Java router is composed.

play-interop-test-scala/src/test/scala/akka/grpc/gen/PlayScalaRouterSpec.scala
```diff
@@ -29,7 +30,7 @@ class PlayScalaRouterSpec extends WordSpec with Matchers with BeforeAndAfterAll
   implicit val ec = sys.dispatcher
   implicit val patience = PatienceConfig(timeout = 3.seconds, interval = 15.milliseconds)
 
-  val router = new GreeterServiceImpl
+  val router = new GreeterServiceRouter(new GreeterServiceImpl)
```
richdougherty: Here you can see how a new Scala router and service are composed together.

runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala
```diff
@@ -38,19 +32,10 @@ import scala.compat.java8.OptionConverters._
...
-@InternalApi abstract class PlayRouter(mat: Materializer, serviceName: String) extends play.api.routing.Router {
-
-  private val prefix = s"/$serviceName"
-
-  /**
-   * INTERNAL API
-   */
-  @InternalApi
-  protected def createHandler(serviceName: String, mat: Materializer): HttpRequest => Future[HttpResponse]
+@InternalApi abstract class PlayRouter(prefix: String, underlyingHandler: HttpRequest => Future[HttpResponse]) extends play.api.routing.Router {
```
richdougherty: I removed a lot of the logic from this class and just pass things in as constructor arguments.

runtime/src/main/scala/akka/grpc/scaladsl/play/AbstractAkkaGrpcServiceBindings.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  override def bindings(environment: Environment, configuration: Configuration): Seq[Binding[_]] =
+    services.flatMap(bindingsForService(_, environment, configuration))
+
+  protected def services: Seq[Class[_]]
+
+  @varargs // Helper for Java subclasses
+  final protected def classSeq(classes: Class[_]*): Seq[Class[_]] = classes
+
+  protected def bindingsForService[T](serviceClass: Class[T], environment: Environment, configuration: Configuration): Seq[Binding[_]] = {
```
richdougherty: This is how a service is bound. I try to be helpful in my error messages.

TODO: Write tests to check this works.

play-interop-test-scala/src/test/scala/akka/grpc/gen/PlayScalaRouterSpec.scala
ignasi35: Ah, this looks nice than `val router = new GreeterServiceImpl`! 👍🏼

richdougherty:
Apologies that I didn't get a chance to finish this. Here are some changes that are still needed.

TODO:
- eliminate duplication between Play/Scala and Service/Client codegen
- make Play/Scala codegen Twirl directory layouts and file names the same as each other
- test some of the error conditions with loading classes

dwijnand:
@raboof / @johanandren, given your comments in #364, WDYT about this proposal?

ignasi35:
@richdougherty I can't push to this PR. I'm looking into it more carefully and wanted to do a couple of small changes (unregarding the outcome of the discussion in #364)

dwijnand:
Discussed internally, and the vote was to reject it. Some notes:

- the advantage (arguably 'more familiar' approach, even though 'custom routes' seem like a fairly niche feature to begin with) doesn't seem to weigh up against the disadvantage (more things to explain or go wrong especially for new users)
- could be swayed if it turns out the 'separated' approach helps in customizing the route, like in https://github.com/akka/akka-grpc/issues/382 or https://github.com/akka/akka-grpc/issues/393
- keep the current implementation and adjust later if necessary
- add the `@ApiMayChange` on the play-specific generators (if not there already) so generated code directly used on userland code is marked. (e.g. in https://github.com/akka/akka-grpc/blob/master/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt#L24 )
- PR introduces some discussion worth revisiting in the future

PR #388, opened 2018-09-11
Upgrade to Akka HTTP 10.1.5

dwijnand:


dwijnand:
Oops

ktoso:
Nope, you’d break the lagom interop story

raboof:
@ktoso how so?

raboof:
for akka-grpc we're already targeting lagom 1.5 which is on akka-http 10.1 - perhaps you were thinking of akka-management?

ktoso:
Ah I see... sorry, I lumped them into the same bucket. But as this is a new feature we can target 1.5...

PR #383, opened 2018-09-06
Depend on Akka HTTP 10.1.5

raboof:

PR #389, opened 2018-09-11
Drop unnecessary Akka HTTP core dependency declaration

dwijnand:
These were only there to force the version:

c2c0e491f1da52a303ec063012d0e9016618adbe
40af58218270a9ead94cc8c8d4c23903168435f4


dwijnand:
Hehehe this failed for reasons unrelated to this PR: a semantic conflict from the merging #383 and #387. So master is broken.

PR #391, opened 2018-09-11
Add Play Java gRPC tests

### LINKED ISSUE #361: Play Java gRPC tests
PR #356 introduces Play Scala tests for ScalaTest and specs2, but doesn't include any Play Java tests. We need to add these tests. Before we can do this we will need the Play Scala tests to be split into a separate project so that the Java test code can use some of the common Scala test code. See #359.

dwijnand:
Closes #361

### DIFF
--- a/play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
+++ b/play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
@@ -33,7 +33,7 @@ public final class JavaAkkaGrpcClientHelpers {
    */
   public static ServerEndpoint unsafeGetHttp2Endpoint(final ServerEndpoints serverEndpoints) {
     final scala.collection.Traversable<ServerEndpoint> possibleEndpoints =
-        serverEndpoints.endpoints().filter(e->e.scheme().equals("https") && e.httpVersions().contains("2"));
+        serverEndpoints.endpoints().filter(e->e.httpVersions().contains("2"));
     if (possibleEndpoints.size() != 1) {
       throw new IllegalArgumentException(String.format(
           "gRPC client can't automatically find HTTP/2 connection: " +
--- a/play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
+++ b/play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2018 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.play;
+
+import akka.actor.ActorSystem;
+import akka.grpc.GrpcClientSettings;
+import akka.grpc.GrpcClientSettings$;
+
+import play.api.test.NewTestServer;
+import play.api.test.ServerEndpoint;
+import play.api.test.ServerEndpoints;
+
+import javax.net.ssl.SSLContext;
+
+public final class JavaAkkaGrpcClientHelpers {
+  private static final GrpcClientSettings$ GrpcClientSettings = GrpcClientSettings$.MODULE$;
+
+  private JavaAkkaGrpcClientHelpers() {}
+
+  public static GrpcClientSettings grpcClientSettings(final NewTestServer testServer) {
+    final ServerEndpoint http2Endpoint = unsafeGetHttp2Endpoint(testServer.endpoints());
+    return grpcClientSettings(http2Endpoint, testServer.testServer().application().actorSystem());
+  }
+
+  public static ServerEndpoint unsafeGetHttp2Endpoint(final ServerEndpoints serverEndpoints) {
+    final scala.collection.Traversable<ServerEndpoint> possibleEndpoints =
+        serverEndpoints.endpoints().filter(e->e.scheme().equals("https") && e.httpVersions().contains("2"));
+    if (possibleEndpoints.size() != 1) {
+      throw new IllegalArgumentException(String.format(
+          "gRPC client can't automatically find HTTP/2 connection: " +
+              "%s valid endpoints available: %s",
+          possibleEndpoints.size(),
+          serverEndpoints
+      ));
+    }
+    return possibleEndpoints.head();
+  }
+
+  public static GrpcClientSettings grpcClientSettings(
+      final ServerEndpoint http2Endpoint,
+      final ActorSystem actorSystem
+  ) {
+    final SSLContext sslContext = http2Endpoint.ssl().get().sslContext();
+    return GrpcClientSettings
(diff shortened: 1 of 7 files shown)


play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
```diff
@@ -0,0 +1,95 @@
+package akka.grpc.play.test;
```
marcospereira: Missing headers?

play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+  @After
+  public void stopServer() throws Exception {
+    if (testServer != null) {
+      testServer.stopServer().close();
+      testServer = null;
+      app = null;
+    }
+    if (greeterServiceClient != null) {
+      FutureConverters.toJava(greeterServiceClient.close()).toCompletableFuture().get();
```
marcospereira: Have a timeout here? Or does `greeterServiceClient.close()` future already has a timeout?

play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @After
+  public void stopServer() throws Exception {
+    if (testServer != null) {
+      testServer.stopServer().close();
+      testServer = null;
+      app = null;
+    }
+    if (greeterServiceClient != null) {
+      FutureConverters.toJava(greeterServiceClient.close())
+          .toCompletableFuture().get(30, TimeUnit.SECONDS);
```
ktoso: 30 is intense, sure that 3 would not be enough?

play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
```diff
@@ -0,0 +1,102 @@
...
+import play.libs.ws.*;
+import play.test.*;
+
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
+import static play.inject.Bindings.*;
+
+public final class PlayJavaFunctionalTest implements AkkaGrpcClientHelpers {
+  private static final GreeterService$ GreeterService = GreeterService$.MODULE$;
```
ktoso: isn't it worrying that we need to access `MODULE$` in a java test here?
Is this using the scala or java generated things?

Seems a bit weird for the Service hm...

play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
dwijnand: 🤷‍♂️ 30s is what it is in the specs2 and ScalaTest version.

Maybe @marcospereira has some good guidelines from Play/testing in Play?

play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java
dwijnand: I guess it's using Scala generated things? I'll look into this.

dwijnand:
I'm done with this, so ready for (re)review.

play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
```diff
@@ -0,0 +1,51 @@
...
+  private JavaAkkaGrpcClientHelpers() {}
+
+  public static GrpcClientSettings grpcClientSettings(final NewTestServer testServer) {
+    final ServerEndpoint http2Endpoint = unsafeGetHttp2Endpoint(testServer.endpoints());
+    return grpcClientSettings(http2Endpoint, testServer.testServer().application().actorSystem());
+  }
+
+  public static ServerEndpoint unsafeGetHttp2Endpoint(final ServerEndpoints serverEndpoints) {
+    final scala.collection.Traversable<ServerEndpoint> possibleEndpoints =
+        serverEndpoints.endpoints().filter(e->e.scheme().equals("https") && e.httpVersions().contains("2"));
```
ignasi35: Technically, gRPC (and HTTP/2) don't require `https` so this filter should only assert `e.httpVersions().contains("2")`.

play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public static GrpcClientSettings grpcClientSettings(
+      final ServerEndpoint http2Endpoint,
+      final ActorSystem actorSystem
+  ) {
+    final SSLContext sslContext = http2Endpoint.ssl().get().sslContext();
+    return GrpcClientSettings
+        .connectToServiceAt(http2Endpoint.host(), http2Endpoint.port(), actorSystem)
+        .withSSLContext(sslContext);
+  }
```
ignasi35: Can we create an issue to add javadoc (or add the javadocs on this PR), please

play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
ignasi35: Ah, I see this is a refactor from pre-existing code. We can address separately when we get there.

play-testkit/src/main/java/akka/grpc/play/JavaAkkaGrpcClientHelpers.java
```diff
@@ -14,16 +14,23 @@
...
   public static GrpcClientSettings grpcClientSettings(final NewTestServer testServer) {
     final ServerEndpoint http2Endpoint = unsafeGetHttp2Endpoint(testServer.endpoints());
     return grpcClientSettings(http2Endpoint, testServer.testServer().application().actorSystem());
   }
 
+  /**
+   * Unsafely gets the HTTP/2 endpoint from the given ServerEndpoints.
+   *
+   * If no HTTP/2 endpoint exists this throws an IllegalArgumentException.
+   */
```
ignasi35: Why is this called `unsafeGet...` instead of a simple `get...`? I think in Java it's common to just call stuff `get...` and have exceptions thrown in case of failure.

dwijnand: To try and communicate the lack of safety in calling it.

ignasi35: we can improves this name later.

PR #395, opened 2018-09-13
Fixes a setting overload (typo)

ignasi35:
`mechanism` is a field of the `service-discovery` config object.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -45,7 +45,7 @@ object GrpcClientSettings {
     val defaultServiceConfig = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
       .withValue("service-name", ConfigValueFactory.fromAnyRef(serviceName))
       .withValue("port", ConfigValueFactory.fromAnyRef(defaultPort))
-      .withValue("service-discovery-mechanism", ConfigValueFactory.fromAnyRef(serviceDiscoveryMechanism))
+      .withValue("service-discovery.mechanism", ConfigValueFactory.fromAnyRef(serviceDiscoveryMechanism))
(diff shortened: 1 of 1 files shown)


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -45,7 +45,7 @@ object GrpcClientSettings {
     val defaultServiceConfig = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
       .withValue("service-name", ConfigValueFactory.fromAnyRef(serviceName))
       .withValue("port", ConfigValueFactory.fromAnyRef(defaultPort))
-      .withValue("service-discovery-mechanism", ConfigValueFactory.fromAnyRef(serviceDiscoveryMechanism))
+      .withValue("service-discovery.mechanism", ConfigValueFactory.fromAnyRef(serviceDiscoveryMechanism))
```
raboof: How would `discoverService` be used? Would it make more sense to replace it by something like `connectTo(name: String, discovery: SimpleServiceDiscovery, portName: Option[String], protocol: Option[String])` (and introduce `withServiceDiscovery`, `withServicePortName` and `withServiceProtocol` to implement it safely)?

ignasi35: That is a good idea. I'm fighting that exactly while working on https://github.com/lagom/akka-grpc-lagom-quickstart-scala

ignasi35: `connectTo(name: String, discovery: SimpleServiceDiscovery, portName: Option[String], protocol: Option[String])` looks better than passing a service discovery name around.

ignasi35:
PLEASE BUILD

raboof:
Added #399 for the idea discussed in the comments

PR #400, opened 2018-09-13
Factory method to use default service discovery in the actor system.

### LINKED ISSUE #399: Improve creating discovery-backed GrpcClientSettings from code
`GrpcSettings.discoverService` receives a `serviceDiscoveryMechanism: String` parameter, which appears awkward.

It might make sense to replace it by something like `connectTo(name: String, discovery: SimpleServiceDiscovery, portName: Option[String], protocol: Option[String])` (and introduce `withServiceDiscovery`, `withServicePortName` and `withServiceProtocol` to implement it safely).

Discussed in #395

ignasi35:
Fixes #399

Required by https://github.com/lagom/akka-grpc-lagom-quickstart-scala/pull/7

### DIFF
--- a/docs/src/main/paradox/client/configuration.md
+++ b/docs/src/main/paradox/client/configuration.md
@@ -65,7 +65,7 @@ Scala
 Java
 :  @@snip [GrpcClientSettingsCompileOnly](/runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #sd-settings }
 
-Alternatively if a `SimpleServiceDiscovery` is available else where in your system is can be passed in:
+Alternatively if a `SimpleServiceDiscovery` instance is available elsewhere in your system it can be passed in:
 
 Scala
 :  @@snip [GrpcClientSettingsCompileOnly](/runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly.scala) { #provide-sd }
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -22,7 +22,7 @@ import scala.concurrent.duration.{ Duration, _ }
 object GrpcClientSettings {
 
   /**
-   * Create a client hat uses a static host and port. Default configuration
+   * Create a client that uses a static host and port. Default configuration
    * is loaded from reference.conf
    */
   def connectToServiceAt(host: String, port: Int)(implicit actorSystem: ActorSystem): GrpcClientSettings = {
@@ -30,22 +30,6 @@ object GrpcClientSettings {
     val defaultServiceConfig = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
       .withValue("host", ConfigValueFactory.fromAnyRef(host))
       .withValue("port", ConfigValueFactory.fromAnyRef(port))
-    GrpcClientSettings.fromConfig(defaultServiceConfig)
-  }
-
-  /**
-   * Create a client with the given service discovery mechanism. Default configuration
-   * is loaded from reference.conf
-   *
-   * @param serviceName               Name to look up in serviceDiscovery
-   * @param defaultPort               Port to use if service discovery only returns a host name
-   * @param serviceDiscoveryMechanism Service discovery mechanism to use. Must be correctly configured in the ActorSystem's config
-   */
-  def discoverService(serviceName: String, defaultPort: Int, serviceDiscoveryMechanism: String)(implicit actorSystem: ActorSystem): GrpcClientSettings = {
-    val defaultServiceConfig = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
-      .withValue("service-name", ConfigValueFactory.fromAnyRef(serviceName))
-      .withValue("port", ConfigValueFactory.fromAnyRef(defaultPort))
-      .withValue("service-discovery.mechanism", ConfigValueFactory.fromAnyRef(serviceDiscoveryMechanism))
     GrpcClientSettings.fromConfig(defaultServiceConfig)
   }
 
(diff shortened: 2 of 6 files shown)


docs/src/main/paradox/client/configuration.md
```diff
@@ -65,7 +65,7 @@ Scala
 Java
 :  @@snip [GrpcClientSettingsCompileOnly](/runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java) { #sd-settings }
 
-Alternatively if a `SimpleServiceDiscovery` is available else where in your system is can be passed in:
+Alternatively if a `SimpleServiceDiscovery` instance is available elsewhere in your system it can be passed in:
```
ignasi35: some typos in docs

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
```diff
@@ -23,12 +26,13 @@ public static void sd() {
                 .withTls(false);
         //#simple-programmatic
 
+        SimpleServiceDiscovery serviceDiscovery= ServiceDiscovery$.MODULE$.apply(actorSystem).discovery();
+
```
ignasi35: This is not really necessary since it's not printed in docs and could be simply `SimpleServiceDiscovery serviceDiscovery= null;`

raboof: This could be `ServiceDiscovery.get(as).discovery()`, right? In that form it might be acceptable to include in the example.

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
```diff
@@ -23,12 +26,13 @@ public static void sd() {
...
         //#provide-sd
         // An ActorSystem's default service discovery mechanism
-        GrpcClientSettings.discoverService(
+        GrpcClientSettings.connectTo(
                 "my-service",
-                443,
-                "config", // config based service discovery must be defined in the ActorSystems' config
+                serviceDiscovery,
```
raboof: Hmm, now we're passing in both `ServiceDiscovery` and the `ActorSystem` - would it make sense to (also) have a `connectTo` that only takes the service name(s) and the `ActorSystem`?

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
raboof: Or pass in for example `ConfigSimpleServiceDiscovery(system)`

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -68,6 +52,29 @@ object GrpcClientSettings {
...
+   * Discovery implementation provided. Other than the provided parameters, this factory method will read all other
+   * values from `application.conf`.
+   *
+   * @param serviceName name of the remote service to lookup.
+   * @param discovery   service discovery instance to resolve the lookup.
+   */
+  def connectTo(serviceName: String, discovery: SimpleServiceDiscovery)(implicit actorSystem: ActorSystem): GrpcClientSettings = {
+    val clientConfiguration: Config = actorSystem.settings.config.getConfig("akka.grpc.client").getConfig("\"*\"")
+    val resolveTimeout = clientConfiguration.getDuration("service-discovery.resolve-timeout").asScala
+    new GrpcClientSettings(serviceName, discovery, -1, resolveTimeout)
```
raboof: Hrm I'm not sure we're consistently using the 'service name' here, but that's not due to this PR, so ticketified seperately in #403

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -68,6 +52,29 @@ object GrpcClientSettings {
...
+      .copy(
+        servicePortName = getOptionalString(clientConfiguration, "service-discovery.port-name"),
```
raboof: It would be good if we only needed those magic strings in one place... could we refactor this so we could do something like

```
fromConfig("*")
  .withDiscovery(discovery)
```
Currently `fromConfig` validates that there is a discovery mechanism defined, that would have to be factored out, but that seems doable. WDYT?

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
ignasi35: Sure! Fixing.

I prefer not passing `ConfigSimpleServiceDiscovery` since it's a specific implementation and use what's loaded by the extension instead.

runtime/src/test/java/jdocs/akka/grpc/client/GrpcClientSettingsCompileOnly.java
ignasi35: I like the simplification, but then I'd rename the method to `GrpcClientSettings.usingServiceDiscovery(String, ActorSystem)` (`usingServiceDiscovery` isntead of `connectTo`) so it's clear why only a `String` is necessary.

ignasi35:
Pushed some API improvements. Tests still pending.

runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala
```diff
@@ -169,6 +171,32 @@ class GrpcClientSettingsSpec extends WordSpec with Matchers with ScalaFutures {
...
+      // invoking usingServiceDiscovery provides a GrpcClientSettings instance
+      // that uses the ServiceDiscovery in the ActorSystem
+      val settings = GrpcClientSettings.usingServiceDiscovery("a-downstream-service")(actorSystem)
+      settings.serviceDiscovery should not be null
     }
+
   }
 }
+class FakeServiceDiscovery extends SimpleServiceDiscovery{
+  override def lookup(lookup: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = ???
```
dwijnand: How does this result in a non-null service discovery?

ignasi35:
```scala
      // given an ActorSystem with a ServiceDiscovery method configured
      val sdConfig =
        s"""
          |akka.discovery {
          |  method = "${classOf[FakeServiceDiscovery].getName}"
          |}
        """.stripMargin
      val actorSystem = ActorSystem("test-with-service-discovery", ConfigFactory.parseString(sdConfig))
```
this code makes sure the `ActorSystem` will have an instance of `FakeServiceDiscovery` as a service discovery mechanism. So, creating a `GrpcClientSettings` instance via `GrpcClientSettings.usingServiceDiscovery` will use that service discovery instance. But, if a user tried an operation implying a `lookup` with that service discovery, then the operation would fail.

This test and the previous one assert that if there's a service discovery it's used and if there's none things fail fast.

dwijnand:
Ready to be merged if Jenkins comes back green.

ignasi35:
Jenkins is green!

dwijnand:
Well then!

PR #406, opened 2018-09-14
Move AkkaGrpcClient out of 'internal'

dwijnand:
Co-authored-by: Arnout Engelen <arnouten@bzzt.net>

Supersedes #369.


dwijnand:
@raboof 

> I'm less sure about the `AkkaGrpcClientFactory` - perhaps it would make sense to move that one into play-testkit actually.

There doesn't seem anything Play-specific about that. So, holding back from spinning up another more general testkit, I just left it where it is.

PR #411, opened 2018-09-21
Extract method in SelfSigned

ignasi35:
This is a noop change. extracts a method in preparation for a refactor that will eventually move this code to ssl-config.

### DIFF
--- a/play-testkit/src/main/scala/play/api/test/SelfSigned.scala
+++ b/play-testkit/src/main/scala/play/api/test/SelfSigned.scala
@@ -24,15 +24,20 @@ object SelfSigned {
   lazy val (sslContext, trustManager): (SSLContext, X509TrustManager) = {
     val keyStore: KeyStore = FakeKeyStore.generateKeyStore
 
-    val kmf: KeyManagerFactory = KeyManagerFactory.getInstance("SunX509")
-    kmf.init(keyStore, "".toCharArray)
-    val kms: Array[KeyManager] = kmf.getKeyManagers
+    buildContextAndTrust(keyStore)
+  }
 
-    val tmf: TrustManagerFactory = TrustManagerFactory
-      .getInstance(TrustManagerFactory.getDefaultAlgorithm())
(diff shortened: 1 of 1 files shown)


ignasi35:
The plan is to move the extracted function away so that we can build `SSLContext` (et al.) given _any_ `keystore`. I think this could eventually be part of a new `testkit` module in https://github.com/lightbend/ssl-config

There is quite a few places/projects that have very similar needs (fake certs, trust stores built on-the-fly,...).

The long term effort would solve https://github.com/akka/akka-grpc/issues/89

renatocaval:
Travis said yes!

play-testkit/src/main/scala/play/api/test/SelfSigned.scala
```diff
@@ -24,16 +24,21 @@ object SelfSigned {
...
+  private def buildContextAndTrust(keyStore: KeyStore) = {
 
-    val tmf: TrustManagerFactory = TrustManagerFactory
-      .getInstance(TrustManagerFactory.getDefaultAlgorithm())
+    val tmf: TrustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
     tmf.init(keyStore)
     val tms: Array[TrustManager] = tmf.getTrustManagers
     val x509TrustManager: X509TrustManager = tms(0).asInstanceOf[X509TrustManager]
 
+    val kmf: KeyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
```
dwijnand: @ignasi35 was the change from "SunX509" to `KeyManagerFactory.getDefaultAlgorithm` intentional?

dwijnand: Moreover, is it significant?

ignasi35: It was intentional and _I think_ it's not significant.

Issue #345, opened 2018-08-14
Cross-build for Scala 2.11


raboof:

ktoso:
Sanity check, happening for 0.2? Would be nice if not troublesome

chbatey:
Not planned to but we can

2m:
We would like a release of akka-grpc for Scala 2.11 for the https://github.com/akka/alpakka/pull/1162

If you want I can try to cut it. :)

patriknw:
That would be great @2m, I don't think we have a reason for not doing it.

2m:
Can we cut a release of `0.4`? I think it should only be a matter of creating a release in GitHub.

PR #414, opened 2018-09-21
Add support to Scala binary version propagation from the build

2m:
To support other Scala version in the runtime (like 2.11 or the upcoming 2.13) the Scala binary version from the project that is using akka-grpc needs to be propagated to the akka-grpc sbt plugin and eventually to `suggestedDependencies` which is then used by the protoc sbt plugin to add additional `libraryDependencies`.

This PR adds support for that.

~WIP because Gradle and Maven plugins have not yet been updated.~ See https://github.com/akka/akka-grpc/pull/414#issuecomment-424026521

Towards #345

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
@@ -12,15 +12,21 @@ import protocbridge.Artifact
  * Code generator trait that is not directly bound to scala-pb or protoc (other than the types).
  */
 trait CodeGenerator {
+  import CodeGenerator._
 
   /** Generator name; example: `akka-grpc-scala` */
   def name: String
 
   def run(request: CodeGeneratorRequest, logger: Logger): CodeGeneratorResponse
 
-  /** Takes Scala binary version (ex.: "2.12") and returns suggested dependency Seq **/
-  def suggestedDependencies: String => Seq[Artifact]
+  /** Takes Scala binary version and returns suggested dependency Seq **/
+  def suggestedDependencies: ScalaBinaryVersion => Seq[Artifact]
 
   final def run(request: Array[Byte], logger: Logger): Array[Byte] =
     run(CodeGeneratorRequest.parseFrom(request), logger: Logger).toByteArray
+}
+
+object CodeGenerator {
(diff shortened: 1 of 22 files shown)


codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
```diff
@@ -18,7 +18,8 @@ trait CodeGenerator {
 
   def run(request: CodeGeneratorRequest, logger: Logger): CodeGeneratorResponse
 
-  def suggestedDependencies: Seq[Artifact]
+  /** Takes Scala binary version (ex.: "2.12") and returns suggested dependency Seq **/
+  def suggestedDependencies: String => Seq[Artifact]
```
dwijnand: should we introduce a type instead of unicode string?

runtime/src/main/scala/akka/grpc/javadsl/package.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  /**
+   * Helper for creating java.util.function.Function instances from Scala
+   * functions as Scala 2.11 does not know about SAMs.
+   */
+  def javaFunction[A, B](f: A => B): java.util.function.Function[A, B] =
+    new java.util.function.Function[A, B]() {
+      override def apply(a: A): B = f(a)
+    }
+}
```
raboof: These aren't found in any java compat lib? Then :+1:

2m: I could not find them anywhere. Have you seen something like this around? :)

2m:
I have found out that Gradle and Maven plugins do not use suggested dependencies.

In the Gradle akka-grpc plugin we [add runtime dependency manually](https://github.com/akka/akka-grpc/blob/78cd120c2d346648bbd47586a4c0927f8413a8d0/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L122). And when using Maven plugin we recommend adding runtime dependency manually.

2m:
Can we merge this? I'll work on updating the build so it cross-publishes `runtime` next.

dwijnand:
Yep!

PR #415, opened 2018-09-24
Upgrade scalapb to 0.8.0

nktpro:
There was one breaking change from 0.7.x: `DescriptorPimps` was renamed to `DescriptorImplicits` with a slight difference in constructor API.


raboof:
OK TO TEST

PR #416, opened 2018-09-24
Add "commonSettings" on play-testkit

dwijnand:
This ensures it's got the right organization.


dwijnand:
`organization`, probably. The other two I'd have to check if it works.

But that's a build refactoring question that I wouldn't want to slow down this fix.

PR #430, opened 2018-10-09
Cleanup JavaAkkaGrpcClientHelpers

dwijnand:


dwijnand:
No, it's not a Java thing.  The same method is used in the Scala API too.  But this is testkit API, so throwing an exception because the ServerEndpoints you've provided doesn't contain an endpoint that is HTTP/2 and has SSL is probably ok.  And no need to strain everything with "unsafe".

PR #431, opened 2018-10-09
Move specs2/scalatest things into dedicated modules

dwijnand:
This follows the convention in Play, so:
* Java users use akka-grpc-play-testkit,
* specs2 users use akka-grpc-play-specs2, &
* ScalaTest users use akka-grpc-scalatest.


build.sbt
```diff
@@ -199,11 +199,29 @@ lazy val playTestkit = Project(
...
+val playSpecs2 = Project("akka-grpc-play-specs2", file("play-specs2"))
+  .dependsOn(playTestkit, playTestkit % "test->test")
+  .settings(
+    commonSettings,
+    crossScalaVersions := Seq(scala211, scala212),
+    Dependencies.playSpecs2,
+  )
+  .pluginTestingSettings
+
+val playScalaTest = Project("akka-grpc-play-scalatest", file("play-scalatest"))
```
dwijnand: `play-scalatest` looks better than `play-scala-test` IMO.

project/Dependencies.scala
```diff
@@ -130,13 +136,13 @@ object Dependencies {
   )
 
   val playTestkit = l ++= Seq(
-    Compile.play
-  ) ++ (Seq(
     Compile.play,
-    Test.playTest,
-    Test.playSpecs2,
-    Test.scalaTestPlusPlay
-  ).map(_.withConfigurations(Some("provided"))))
```
dwijnand: I'm not sure what the thinking was here, so if anyone remembers, let me know. 🙂

raboof: I suspect the thinking was to avoid pulling in specs2 for scalatest users, avoid pulling in scalatest for specs2 users, etc. So it looks like that's indeed no longer necessary in the new structure :+1:

dwijnand: That sounds plausible! 👍

ignasi35:
LGTM. 

I guess we can defer the removal of all the `// RICH` comments to a final PR once this and play are stable enough.

dwijnand:
All the files with `//RICH` will eventually be removed, either because they were migrated to Play or because they were substituted by some other code in Play.

Issue #413, opened 2018-09-21
Create a ManagedClient generator


ignasi35:
We could provide a generator for most common uses. The Api provided offers a set of tools but we could include a convenience generator with some defaults in mind:

 - use the `ServiceDiscovery` configured in the `actorSystem`
 - wraps the `GrpcClient` on a `ReconnectClient[T <: AkkaGrpcClient]` but still provides an interface that extends the user's gRPC interface [1]
 - register a coordinated shutdown task to invoke `close` on the client

This should not be a default generator (not part of `AkkaGrpc.Scala` or `AkkaGrpc.Java`)  but an opt-in used as `akkaGrpcExtraGenerators`. 

[1] Using `ReconnectClient[T <: AkkaGrpcClient]`  mans the user no longer interacts with their interface. When injecting the instance they no longer inject `gs:GreeterService` but `rgs: ReconnectClient[GreeterService]`, etc...

dwijnand:
We think, for now, this in the "nice to have" camp, in terms of testing purposes.  If, however, we find that it's a blocker, we'll pull it into our focus.

ignasi35:
Related to #442 and #449

dwijnand:
Not sure what the dependency on this is, so I'm not sure what state it should be in (backlog/ready/in progress).

PR #440, opened 2018-10-18
Makes RestartingClient an AkkaGrpcClient

ignasi35:
`RestartingClient[T]` already provided `def close` and only differed from `AkkaGrpcClient` in not providing `def closed`. This turns `RestartingClient[T]` into a subclass of `AkkaGrpcClient`.

Once this and #442 are merged it'll be possible to implement a `RestartingTServiceClient` using the `RestartingClient[T]` improved here and implementing the `TServiceClient` `trait` introduced in #442. 

Related to #413

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/AkkaGrpcClient.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/AkkaGrpcClient.scala
@@ -12,6 +12,15 @@ import akka.annotation.DoNotInherit
 /** Common trait of all generated Akka gRPC clients. Not for user extension. */
 @DoNotInherit
 trait AkkaGrpcClient {
+  /**
+   * Initiates a shutdown in which preexisting and new calls are cancelled.
+   */
   def close(): CompletionStage[Done]
+
+  /**
+   * Returns a CompletionStage that completes successfully when shutdown via close()
+   * or exceptionally if a connection can not be established or reestablished
+   * after maxConnectionAttempts.
+   */
   def closed(): CompletionStage[Done]
 }
--- a/runtime/src/main/scala/akka/grpc/javadsl/RestartingClient.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/RestartingClient.scala
@@ -14,17 +14,17 @@ import scala.concurrent.{ ExecutionContext, Future }
 import scala.compat.java8.FutureConverters._
 
 /**
- * Wraps a Akka gRPC  client and restarts it if a [ClientConnectionException] is thrown.
- * All other exceptions result in closing and any calls to withClient throwing
+ * Wraps a Akka gRPC client and restarts it if a [ClientConnectionException] is thrown.
+ * All other exceptions result in closing any calls to withClient throwing
  * a [ClientClosedException].
  */
 @ApiMayChange
-class RestartingClient[T <: AkkaGrpcClient](create: () => T, ec: Executor) {
+class RestartingClient[T <: AkkaGrpcClient](create: () => T, ec: Executor) extends AkkaGrpcClient {
   private val delegate: ScalaRestartingClient[JavaClientWrapper[T]] = new ScalaRestartingClient[JavaClientWrapper[T]](() => new JavaClientWrapper[T](create()))(ExecutionContext.fromExecutor(ec))
 
   def withClient[A](f: T => A): A = delegate.withClient(t => f(t.javaClient))
-  def close(): CompletionStage[Done] = delegate.close().toJava
-
+  override def close(): CompletionStage[Done] = delegate.close().toJava
+  override def closed(): CompletionStage[Done] = delegate.closed().toJava
 }
 
 /**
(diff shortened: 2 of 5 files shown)


ignasi35:
By making `RestartingClient` extend `AkkaGrpcClient` it is possible to implement a `WrappedFooServiceClient` which is implemented internally with a `RestartingClient[FooService]`. That is useful so it's trivial to switch from non-restarting implementation to a restarting one. So instead of using:

```scala
class InventoryServiceImpl(
  greeterClient: RestartingClient[OrdersServiceClient]) extends InventoryService {
...
}
```

we can implement the `InventoryServiceImpl` just requiring an instance of `OrdersServiceClient` without caring if it's restarting or not:

```scala
class InventoryServiceImpl(
  greeterClient: OrdersServiceClient) extends InventoryService {
...
}
```

In the second example, it's users of `InventoryServiceImpl` who will inject a plain `OrdersServiceClient` or a `RestartingOrdersServiceClient` where: 

```scala
class RestartingOrdersServiceClient extends OrdersService with AkkaGrpcClient{...}
```

runtime/src/main/scala/akka/grpc/javadsl/AkkaGrpcClient.scala
```diff
@@ -12,6 +12,15 @@ import akka.annotation.DoNotInherit
...
+   * Initiates a shutdown in which preexisting and new calls are cancelled.
+   */
   def close(): CompletionStage[Done]
+
+  /**
+   * Returns a CompletionStage that completes successfully when shutdown via close()
+   * or exceptionally if a connection can not be established or reestablished
+   * after maxConnectionAttempts.
+   */
   def closed(): CompletionStage[Done]
```
ignasi35: Copied the docs from the code generation templates to the `trait` and `interface`.

runtime/src/main/scala/akka/grpc/javadsl/RestartingClient.scala
```diff
@@ -14,17 +14,17 @@ import scala.concurrent.{ ExecutionContext, Future }
...
 @ApiMayChange
-class RestartingClient[T <: AkkaGrpcClient](create: () => T, ec: Executor) {
+class RestartingClient[T <: AkkaGrpcClient](create: () => T, ec: Executor) extends AkkaGrpcClient {
   private val delegate: ScalaRestartingClient[JavaClientWrapper[T]] = new ScalaRestartingClient[JavaClientWrapper[T]](() => new JavaClientWrapper[T](create()))(ExecutionContext.fromExecutor(ec))
 
   def withClient[A](f: T => A): A = delegate.withClient(t => f(t.javaClient))
-  def close(): CompletionStage[Done] = delegate.close().toJava
-
+  override def close(): CompletionStage[Done] = delegate.close().toJava
+  override def closed(): CompletionStage[Done] = delegate.closed().toJava
```
ignasi35: This is the `javadsl` half of the main change on this PR.

runtime/src/main/scala/akka/grpc/scaladsl/AkkaGrpcClient.scala
```diff
@@ -12,6 +12,15 @@ import akka.annotation.DoNotInherit
...
+   * Initiates a shutdown in which preexisting and new calls are cancelled.
+   */
   def close(): Future[Done]
+
+  /**
+   * Returns a Future that completes successfully when shutdown via close()
+   * or exceptionally if a connection can not be established or reestablished
+   * after maxConnectionAttempts.
+   */
   def closed(): Future[Done]
```
ignasi35: See above. Copied from templates.

ignasi35:
The improvements I made on cdde1ee253a04b687ed517acbeba169356951e6e broke my own test. Looking.

dwijnand:
Thanks for this.  I ran out of time today to review it, but I'm interested in this change.  (Feel free to merge without me, I can review it post-merge too).

ignasi35:
Hmmm, I just realised this change is not completely fulfilling my idea to be able to [swap a non-restarting or a restarting client](https://github.com/akka/akka-grpc/pull/440#issuecomment-431090888).

With this PR, both `RestartingOrdersServiceClient` and `OrdersServiceClient` are of type `OrdersService with AkkaGrpcClient` but `RestartingOrdersServiceClient` is not a subtype of `OrdersServiceClient` so 

```scala
class InventoryServiceImpl(
  greeterClient: OrdersServiceClient) extends InventoryService {
...
}
```

can't be invoked with `RestartingOrdersServiceClient`. 

We can discuss on a separate issue/PR, though.

dwijnand:
What did you mean to link there?  You're linking back to this very PR.

ignasi35:
> What did you mean to link there? You're linking back to this very PR.

@dwijnand I am linking to a specific comment on this same PR were I stated an idea of a possible use case driving this change. I later realized the code in this PR doesn't fulfill the use case yet.

ignasi35:
> but RestartingOrdersServiceClient is not a subtype of OrdersServiceClient

I've submitted #442 to solve this.

ignasi35:
After discussing https://github.com/akka/akka-grpc/pull/442#issuecomment-433962519 and https://github.com/akka/akka-grpc/pull/442#issuecomment-434016505 I think there's an alternative option which would make this PR irrelevant. 

Please, do not merge yet.

ignasi35:
See #449

PR #442, opened 2018-10-19
[WIP] Introduces ServiceClient trait and RawServiceClient impl

ignasi35:
Related to #440 and #413

This PR introduces:

```scala
trait @{service.name}Client extends @{service.name} with AkkaGrpcClient
```

and changes:

```patch
- final class @{service.name}Client(...) extends ${service.name} with AkkaGrpcClient {
+ final class Raw@{service.name}Client(...) extends @{service.name}Client {
```

so it's possible to produce different types of clients. The default implementation of the ServiceClient is named `RawServiceClient` because it implements the interface _and_ produces lifted versions of the methods.


ignasi35:
So far I've only made improvements on the scaladsl. Once we've agreed this is a good path forward I'd make the equivalent changes on the javadsl.

raboof:
So the idea would be that after we introduce this, we could also introduce a generated `Restarting@{service.name}Client` which contains the restarting logic and has forwarders for all methods on `Raw@{service.name}Client`, both lifted and unlifted?

I think I see where you're going, but this is starting to make me doubt the decision to split the restarting logic from the client itself.

Earlier we made it an explicit choice, 'outside' of the client itself. I can see how this may be inconvenient in the DI case, where you get injected a `Client` rather than a client builder from which you can create a restarting client. Getting retry logic might be hard in that case. Making restarting an implementation detail of the `Client` instead of an 'external thing' might indeed make sense - I can imagine eventually we'd want to be able to have an akka-http client connection pool that is able to take care of reconnecting, so this would be a step in that direction.

I wonder if it might make sense to always build the restart logic right into the Client itself, rather than introducing both a `Raw` and a `Restarting` client both visible to the user. WDYT?

ignasi35:
My approach in #442 was to _not_ expose the lifted methods in `RestartingXClient`.

Then, `RawXClient` wouldn't auto-restart but give access to _raw_ headers while `RestartingXClient` would be an implementation with less power (with less power comes less responsibility).

There'd be a trade-off but users would still be able to opt-out from `RestartingClient[T]` if they wanted.

 - - - - - - - - - - - - - - - - - -

The current implementation of Akka gRPC Client (using netty-based grpc client) inherits a `Channel` abstraction under the covers that already handles reconnecting. Some combination of settings on `GrpcsClientSettings` make it  possible to tune infinite reconnects. What I'm saying is that `RawXClient` could be good enough when the downstream service instances sit behind a load balancer: we can use `GrpcClientSettings` only stating the `host:port` and enable [infinite reconnect retries](GrpcClientSettings) (the Load Balancer will take care of rediceting to moving instances).

But when the downstream service is a collection of movable instances then I would build a `GrpcClientSettings` instance setup to use `ServiceDiscovery` and low reconnect attempt settings are required. With those settings I would prefer using `RestartingXClient`.

 - - - - - - - - - - - - - 

I think the problem is we leaked the `Channel` reconnection via `GrpcClientSettings` and I'm beginning to embrace your idea to embed `RestartingClient[T]` into `XServiceClient` by default and internally tune it via `GrpcClientSettings`.

ignasi35:
`GrpcClientSettings` offers 3 types of builders:

 * `.usingServiceDiscovery(XService.name)` -> using a client-side discovery implies the service may not be using a server-side load balancer.
 * `.connectToServiceAt("mydomain.com", 8080)` -> this is often used when all the service instances sit behind a single entry point (`Service` in kubernetes, or load balancer).
 * `fromConfig(...)` (reads all details from `application.conf`)

Making `RestartingClient[T]` an implementation detail of `XServiceClient` doesn't make a lot of sense when the remote service is using a load balancer. OTOH, `RestartingClient[T]` is very useful when using client-side load balancing.

I think moving `RestartingClient[T]` is the right move but we need some more settings to: disable it, setup the maximum number of reconnects, ... (others?).

Currently, `GrpcClientSettings` includes a `connectionAttempts` setting to count the number of attempts on the low-level `Channel`. This value is useful in all scenarios since the network may be flaky. Then setting up the client under several scenarios has few possible setups:

 * I connect to `domain.com:8443` and I want to recover when network is flaky: `connectionAttempts=10, useRestarting=false` => reconnect at a low level using the same `host:port` every time. If reconnection fails 10 times, then fail.
 * I connect via service discovery to `downstream-service` and I try to recover when the network is flaky but may have to locate a new instance: `connectionAttempts=2, useRestarting=true` => reconnect at a low level using the current `host:port`. If reconnection fails 2 times, then use service discovery to reconnect.
 * etc...

ignasi35:
(extra notes to address in separate the future)

> I can imagine eventually we'd want to be able to have an akka-http client connection pool that is able to take care of reconnecting, so this would be a step in that direction.

Does a connection pool make sense when the transport is HTTP/2? 🤔 
OTOH, the netty implementation provides several implementations of `Channel`. There's one, for example, that builds several `SubChannel` instances and provides client-side load balancing. I think this could be replicated in Akka-gRPC when using `ServiceDiscovery` based connections and locating many downstream instances.

ignasi35:
See #449

raboof:
> Does a connection pool make sense when the transport is HTTP/2? thinking

Good question: this will certainly look different than a 'traditional' connection pool, but I suspect some concepts might still carry over

PR #449, opened 2018-11-01
Replace RerstartingClient with an always-restart ClientState

ignasi35:
Supersedes #440 and #442

Removes `RerstartingClient` moving its code into the new internal `ClientState` which also takes over the control of the lifecycle of `Channel`. This way using _restarting_ is no longer opt-in but enabled by default in all generated clients. 

`ClientState` provides the `withChannel` which behaves like `RerstartingClient#withClient`: it produces a new `Channel` instance whenever the current one was prematurely closed. Invoking `close()` on the `clientState` will prevent new instance creation. 

Also, moving `internalChannel` and other fields out of the generated code makes the generated code smaller. :-)


ignasi35:
PLS BUILD

ignasi35:
rebased in `master`

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * INTERNAL API
+ * Used from generated code so can't be private.
+ *
+ * Client utilities taking care of Channel reconnection and Channel lifecycle in general.
+ */
+@InternalApi
+class ClientState(settings: GrpcClientSettings)(implicit mat: Materializer, ex: ExecutionContext) {
+
+  private val internalChannelRef: AtomicReference[InternalChannel] = new AtomicReference[InternalChannel](create())
```
johanandren: Small potential improvement, use `OptionVal` instead of explicit null

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+  private val closeDemand: Promise[Done] = Promise[Done]()
+
+  mat match {
+    case m: ActorMaterializer =>
+      m.system.whenTerminated.foreach(_ => close())(ex)
```
johanandren: But what about when the materializer is shut down, or can it not be because we create it internally elsewhere?

Also, if the system was terminated, will that `ex` really be something you can schedule jobs on?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+    case _ =>
+  }
+
+  def withChannel[A](f: Future[ManagedChannel] => A): A = {
+    val c = internalChannelRef.get()
+    if (c != null)
+      f(internalChannelRef.get().managedChannel)
```
johanandren: No guarantee it hasn't turned `null` when we get here, use the null-checked `c` instead.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  def closed(): Future[Done] = {
+    // while there's no request to close this RestartingClient, it will continue to restart.
+    // Once there's demand, the `closed` future will redeem flatMapping with the `closed()`
+    // future of the clientRef that's active at that moment.
+    closeDemand.future.flatMap { _ =>
+      val c = internalChannelRef.get()
+      if (c != null)
+        internalChannelRef.get().done
```
johanandren: Another one that could be `null`

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+        Future.successful(Done)
+    }
+  }
+
+  @tailrec
+  final def close(): Future[Done] = {
+    closeDemand.trySuccess(Done)
+    val c = internalChannelRef.get()
+    if (c != null) {
+      val done = ChannelUtils.close(c)
```
johanandren: Two concurrent calls could both get here and cause 2 calls to close, is that ok? If not we likely will need to have another closing state cas and only let the winner close it.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+          // Only one client is alive at a time. However a close() could have happened between the get() and this set
+          if (!internalChannelRef.compareAndSet(old, newInternalClient)) {
+            // close the newly created client we've been shutdown
+            ChannelUtils.close(newInternalClient)
+          }
+        }
+      case Failure(_) =>
+        close()
+      case _ =>
+      // let all other exceptions and success through
```
johanandren: The exceptions are caught in the previous `case`

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
dwijnand:
```suggestion
        c.done
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
dwijnand:
```suggestion
      f(c.managedChannel)
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
dwijnand:
```suggestion
      // let success through
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+    val c = internalChannelRef.get()
+    if (c != null)
+      f(internalChannelRef.get().managedChannel)
+    else // Likely to happen if this is a shared client during shutdown
+      throw new ClientClosedException
+  }
+
+  def closed(): Future[Done] = {
+    // while there's no request to close this RestartingClient, it will continue to restart.
+    // Once there's demand, the `closed` future will redeem flatMapping with the `closed()`
```
dwijnand:
```suggestion
    // Once there's demand, the `closeDemand` future will redeem flatMapping with the `closed()`
```

... I think?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+    // future of the clientRef that's active at that moment.
```
dwijnand:
```suggestion
    // future of the internalChannelRef that's active at that moment.
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  def channelFactory: InternalChannel = NettyClientUtils.createChannel(settings)
+
+  private def create(): InternalChannel = {
+    val c: InternalChannel = channelFactory
+    c.done.onComplete {
+      case Failure(_: ClientConnectionException) =>
+        val old = internalChannelRef.get()
+        if (old != null) {
```
dwijnand: Why do we guard on the `internalChannelRef` not being `null`?  Don't we want to restart in that case too?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+  }
+}
+/**
+ * INTERNAL API
+ * Used from generated code so can't be private.
+ *
+ * Thrown if a withChannel call is called after closing the internal channel
+ */
+@InternalApi
+class ClientClosedException() extends RuntimeException("withChannel called after close()")
```
dwijnand:
```suggestion
final class ClientClosedException() extends RuntimeException("withChannel called after close()")
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import scala.annotation.tailrec
+import scala.concurrent.{ ExecutionContext, Future, Promise }
+import scala.util.Failure
+
+/**
+ * INTERNAL API
+ * Used from generated code so can't be private.
+ *
+ * Client utilities taking care of Channel reconnection and Channel lifecycle in general.
```
dwijnand: Is the capital-C "Channel" a reference to something specific?

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -9,36 +9,23 @@ package @service.packageName
...
-  mat match {
-    case m: ActorMaterializer =>
-      m.system.whenTerminated.foreach(_ => close())(ex)
-    case _ =>
-  }
+  private val clientState = new ClientState(settings)
 
   @for(method <- service.methods) {
-    private val @{method.name}RequestBuilder = {
+    private def @{method.name}RequestBuilder(channel:Future[ManagedChannel]) = {
```
dwijnand:
```suggestion
    private def @{method.name}RequestBuilder(channel: Future[ManagedChannel]) = {
```

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
dwijnand: Good questions.  But, for the record, this is just moved code: https://github.com/akka/akka-grpc/blob/v0.4.1/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt#L34-L38

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
johanandren: Yeah, I think that would fix it, only one can win the cas for null.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
ignasi35: Nice catch... I was copy-pasting from https://github.com/akka/akka-grpc/pull/449/files#diff-71dc613f3874ac7735ff57740885b292L38

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
ignasi35: > f not we likely will need to have another closing state cas and only let the winner close it.

Hmm, that sounds right.

I think the intermediate state is actually the future returned by `ChannelUtils.close(c)` so that when `AtomicRef[OptionVal[Future[Done]]]` is defined, the value is `Future[Done]` completes when the channel actually closes.

chbatey: close calls shutdown on the ManagedChannel which is idempotent

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
ignasi35: `Channel` is a concept used in gRPC's reference implementation (all languages) and it abstracts the lower layers of the stack (you could have a Channel representing a single socket or a Channel that's a load-balanced pool of connections to many remote endpoints).

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
ignasi35: True, but I think there's a problem with what happens after that: 

```scala
final def close(): Future[Done] = {
    closeDemand.trySuccess(Done)
    val c = internalChannelRef.get()
    if (c != null) {
      val done = ChannelUtils.close(c)
      if (internalChannelRef.compareAndSet(c, null))
        done
      else
        close()
    } else
      Future.successful(Done)
  }
```

Two concurrent calls will compete for nullifying `internalChannelRef`, one will win and get `done` and the other will recurse. On the recursion, it'll get `Future.successful(Done)` which is wrong since the closing operation hasn't finished yet.

If we add that intermediate cas, on the recursion it could get `done` too.

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
ignasi35: There's a ton of (potential) formatting improvements in the code templates. I think we can sort them out at once in a separate PR.

ignasi35:
With the changes introduced in this PR, the client always reconnects:
 * first the internal `Channel` using `connectinoAttempts` and exponential backoff.
 * then, when the current channel is close a channel is created using the service discovery

IIUC, `closed()` was introduced in https://github.com/akka/akka-grpc/pull/286 to monitor if client was alive, but this PR ensure this healthiness and only invoking `close()` will actually stop the infinite loop. `closed()` contributes some complexity in the `ClientState` code.

I think `closed()` could be removed entirely.

 - - - - - - - - - 

OTOH, if we want to put an absolute max number of connectionAttempts then we should add a setting and check it on the `create()` loop:
https://github.com/akka/akka-grpc/pull/449/files/c97dd7968caac17cefccd8109df5e54868fb3566#diff-d89113b7109648797adfe0bb70caf8a0R74

WDYT?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -9,23 +9,27 @@ import java.util.concurrent.atomic.AtomicReference
...
+  */
 @InternalApi
-class ClientState(settings: GrpcClientSettings)(implicit mat: Materializer, ex: ExecutionContext) {
+final class ClientState(settings: GrpcClientSettings)(implicit mat: Materializer, ex: ExecutionContext) {
 
-  private val internalChannelRef: AtomicReference[InternalChannel] = new AtomicReference[InternalChannel](create())
+  private val internalChannelRef: AtomicReference[OptionVal[InternalChannel]] =
+    new AtomicReference[OptionVal[InternalChannel]](OptionVal(create()))
+  // usually null, it'll have a value when the underlying InternalChannel is closing or closed.
+  private val closing: AtomicReference[Future[Done]] = new AtomicReference[Future[Done]](null)
```
dwijnand: Use `OptionVal` then. Or `Option`. No need to play with fire/null.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -34,69 +38,85 @@ class ClientState(settings: GrpcClientSettings)(implicit mat: Materializer, ex:
...
-      if (internalChannelRef.compareAndSet(c, null.asInstanceOf[InternalChannel]))
-        done
-      else // client has had a ClientConnectionException and been re-created, need to shutdown the new one
-        close()
-    } else
-      Future.successful(Done)
+  def close(): Future[Done] = {
+    val maybeChannel = internalChannelRef.get()
+    if (maybeChannel.isDefined) {
+      val channel = maybeChannel.get
```
dwijnand: pattern match?

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
dwijnand: Sure.

dwijnand:
> It still requires the `JavaClient` codegen

Still true.

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
ignasi35: This change has quite some impact and should be reconsidered. Using `val` builds all the `XyzRequestBuilders` once per `Channel` but using `def` creates a `XyzRequestBuilders` instance per request.

ignasi35:
This is ready for a final review and merge.

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
johanandren: Yeah, that was a very conscious design decision with the original, having immutable request builders that can be re-used. Otoh we haven't really worked on performance/benchmarking at all yet. I'd guess the bottlenecks are likely somewhere else than allocating the request builder.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -0,0 +1,129 @@
...
+    maybeChannel match {
+      case Some(channel) =>
+        // invoke `close` on the channel and capture the `channel.done` returned
+        val done = ChannelUtils.close(channel)
+        // set the `closing` to the current `channel.done`
+        closing.compareAndSet(None, Some(done))
+        // notify there's been close demand (see `def closed()` above)
+        closeDemand.trySuccess(Done)
+
+        if (internalChannelRef.compareAndSet(maybeChannel, None)) {
```
johanandren: Here we can lose this cas but we might have won the previous `closing` cas. Quite confusing, I'd try to avoid more than one atomic an represent the states with and adt instead, something like `case class Open(channelref)`, `case class Closing(Future[Done])` and do a single cas on that.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
johanandren: So should we remove this then?

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -110,7 +112,7 @@ public final class @{service.name}Client implements @{service.name}, AkkaGrpcCli
       public StreamResponseRequestBuilder<Source<@method.inputTypeUnboxed, NotUsed>, @method.outputTypeUnboxed> @{method.name}()
     }
     {
-      return @{method.name}RequestBuilder;
+      return clientState.withChannel( this::@{method.name}RequestBuilder);
```
dwijnand: This doesn't work for Scala 2.11, because "scala.Function1 is not a functional interface" in Scala.11: https://travis-ci.com/playframework/play-grpc/jobs/171208483#L542

I'm not sure why this didn't fail in CI.

ignasi35: It looks like, Scala 2.11 is not exercised on CI.

dwijnand: Yep: https://github.com/akka/akka-grpc/issues/498 https://github.com/akka/akka-grpc/issues/497

PR #466, opened 2018-11-16
Exclude unnecessary GPL dependency

ignasi35:


raboof:
(to be exact: it looks like this dependency is "either GPL or CDDL", and the CDDL is generally acceptable, but we don't actually rely on this dependency so it's cleaner to remove it entirely)

project/Dependencies.scala
```diff
@@ -51,7 +51,7 @@ object Dependencies {
 
     val plexusBuildApi = "org.sonatype.plexus" % "plexus-build-api" % "0.0.7" % "optional"// Apache v2
 
-    val play = "com.typesafe.play" %% "play" % Versions.play // Apache M2
+    val play = "com.typesafe.play" %% "play" % Versions.play exclude("javax.activation", "javax.activation-api")
```
ignasi35:
```suggestion
    val play = "com.typesafe.play" %% "play" % Versions.play exclude("javax.activation", "javax.activation-api")  // Apache M2 (exclusion is "either GPL or CDDL")
```

ignasi35: @raboof WDYT?

raboof: :+1: .

is the `M2` intentional?

ignasi35: 🤷🏼‍♂️

raboof:
Excellent let's merge when green

PR #469, opened 2018-11-22
Update gRPC to 1.16.1

TimMoore:


TimMoore:
I'm not sure why `grpc-alts` is failing to download on Travis. It looks like it's there in Maven Central: http://repo1.maven.org/maven2/io/grpc/grpc-alts/1.16.1/

raboof:
OK TO TEST

raboof:
PLS WHITELIST

raboof:
I have no idea. `Detected merged artifact: [NOT FOUND  ] io.grpc#grpc-alts;1.16.1!grpc-alts.jar` looks funny as well?

raboof:
(there are a couple more places that refer to the io.grpc version, so those would need to be adapted as well - I don't see how that could cause the download problem, though)

dwijnand:
I spent quite some time trying to figure that warning out, and have come back without any understanding.

TimMoore:
It's particularly strange that this is failing on both Travis and Jenkins, but passing on my laptop.

A mysterious clue https://github.com/grpc/grpc-java/blob/3c685062b7d73f5995b1670ee825c11f631eff89/build.gradle#L432-L433

It seems to just be about API documentation, however, as far as I can tell.

TimMoore:
I was able to reproduce this locally by disabling Coursier. When it's enabled, it resolves correctly.

This suggests a bug in sbt or Ivy. @eed3si9n do you have any thoughts?

TimMoore:
My local build has some additional output that indicates that the problem is that it is using the wrong filename for some reason (`grpc-alts-1.16.1-original.jar`):

```
[warn] ==== public: tried
[warn]   https://repo1.maven.org/maven2/io/grpc/grpc-alts/1.16.1/grpc-alts-1.16.1-original.jar
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::              FAILED DOWNLOADS            ::
[warn] 	:: ^ see resolution messages for details  ^ ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: io.grpc#grpc-alts;1.16.1!grpc-alts.jar
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

TimMoore:
https://github.com/grpc/grpc-java/blob/959323be974e77dcbf10695d4a3a5e213bfd838c/alts/build.gradle#L64-L68

TimMoore:
This is the change that introduced the problem https://github.com/grpc/grpc-java/pull/4754

TimMoore:
This seems to be a rather complex interoperation bug between sbt/ivy, Gradle, and the rather complex `grpc-java` build. I can't really pin down what the expected behavior should be here, but I was able to work around it by excluding the `grpc-alts` dependency from the interop tests module, without any obvious unwanted consequences.

dwijnand:
Maybe we can add enforce the versions are synced via CI?

https://github.com/TimMoore/akka-grpc/pull/1

project/GrpcVersionSyncCheckPlugin.scala
```diff
@@ -0,0 +1,85 @@
...
+import java.nio.file.{ Files, Path, Paths }
+
+import scala.collection.JavaConverters._
+import scala.sys.process._
+
+import org.eclipse.jgit.diff.RawText
+import sbt._
+import sbt.Keys._
+
+object GrpcVersionSyncCheckPlugin extends AutoPlugin {
```
raboof: Wow that's pretty elaborate. Thx!

sbt-plugin/src/sbt-test/gen-scala-server/00-interop/build.sbt
```diff
@@ -3,8 +3,12 @@ organization := "com.lightbend.akka.grpc"
 // For the akka-http snapshot
 resolvers += Resolver.bintrayRepo("akka", "maven")
 
+val grpcVersion = "1.16.1" // checked synced by GrpcVersionSyncCheckPlugin
+
 libraryDependencies ++= Seq(
-  "io.grpc"                  % "grpc-interop-testing"    % "1.14.0"                     % "protobuf",
+  // Excluding grpc-alts works around a complex resolution bug
```
raboof: So we don't really need grpc-alts on the classpath?

project/GrpcVersionSyncCheckPlugin.scala
TimMoore: All credit to @dwijnand for this!

sbt-plugin/src/sbt-test/gen-scala-server/00-interop/build.sbt
TimMoore: Well, it seems not, but I have to admit that I didn't dig in to see what it actually does. It was only used in interop testing, and our interop tests are passing without it.

I'm not sure if we should really merge in this state, or report the issue upstream. As @eed3si9n pointed out, the behavior of sbt/Ivy makes sense, so it's possible that this is actually a bug in Gradle, in the grpc-java build, or both (plus Coursier).

dwijnand:
> The POM says "original", so Ivy was just being Ivy

What do you mean by that?  Ivy doing what it should be?  Ivy doing the wrong thing?  Ivy not properly mapping Maven semantics to Ivy semantics? (my guess)

eed3si9n:
@dwijnand 

> > The POM says "original", so Ivy was just being Ivy
>
> What do you mean by that? Ivy doing what it should be? Ivy doing the wrong thing? Ivy not properly mapping Maven semantics to Ivy semantics? (my guess)

According to the [POM Reference](https://maven.apache.org/pom.html):
>
> The classifier distinguishes artifacts that were built from the same POM but differ in content. It is some optional and arbitrary string that - if present - is appended to the artifact name just after the version number.

sbt / Ivy builds this logic in pretty strictly using resolver pattern string, and the expectation that the `original` classifier should be located at `grpc-alts-1.16.1-original.jar` is correct. (This mapping is baked into sbt's resolver as pattern string)

However, given classifier can denote all sorts of random things (like `javadoc` and `sources`), I am not sure if it should block the dependency graph resolution by default, assuming Maven and Gradle are able to resolve it.

See also [Maven “shaded” JAR is prefixed with “original” in the file name](https://stackoverflow.com/questions/4088701/maven-shaded-jar-is-prefixed-with-original-in-the-file-name).

dwijnand:
JFYI, can't mix and match these versions.  I tried using akka-grpc 0.4.2 with grpc 1.16.1 and got:

```
[error] /d/play-grpc/play-testdata/target/scala-2.12/src_managed/main/example/myapp/helloworld/grpc/GreeterServiceClient.java:23:1: package io.grpc.stub does not exist
[error] import io.grpc.stub.*;
[error] /d/play-grpc/play-testdata/target/scala-2.12/src_managed/main/example/myapp/helloworld/grpc/GreeterServiceClientPowerApi.java:23:1: package io.grpc.stub does not exist
[error] import io.grpc.stub.*;
```

raboof:
Does adding grpc-stub as an explicit dependency help? (https://search.maven.org/artifact/io.grpc/grpc-stub/1.16.1/jar)

dwijnand:
That's what caused the issue: depending on both akka-grpc 0.4.2 and grpc-stub 1.16.1.  So I rolled back to 1.15.0.  This is in play-grpc, btw.

PR #473, opened 2018-11-30
WIP: update to scalapb 0.8.2

mcanlas:


raboof:
I agree it would be good to update scalapb, but it seems with this PR we're still somehow mixing versions (e.g. `java.lang.NoSuchMethodError: scalapb.options.compiler.Scalapb$ScalaPbOptions.getScope()Lscalapb/options/compiler/Scalapb$ScalaPbOptions$OptionsScope;`)

mcanlas:
This might have to be put on the back-burner until there's a deeper investigation.

I think there's something related to https://github.com/scalapb/ScalaPB/commit/1e5f1bd5f28482a8e6de2b14efafb186259fcef8#diff-010753e90c409f25d80093154f1eda1f in `ScalaPB` that may be breaking. Not sure if the call style to it needs changing or if it's straight broken.

bjaglin:
Indeed, it looks like the `Seq[FileDescriptor]` passed to the `DescriptorImplicits` constructor (https://github.com/akka/akka-grpc/blob/c7098f3abf211ebf55dda258f4dcb4df3b7010f8/codegen/src/main/scala/akka/grpc/gen/scaladsl/Service.scala#L20) must contain the protos where request/response types are defined to be able to lookup the `scalaName` of them.

A more global `DescriptorImplicits` could be requested as implicit in `Service.apply`, but that would obviously be a breaking change... I guess the next step is to file a ticket on ScalaPB to handle cache misses?

dwijnand:
@mcanlas happy to close this in favour of #500?

dwijnand:
Duplicate of #500

PR #477, opened 2018-12-07
Update to akka 2.5.19

raboof:
TODO: remove staging repo when the release is on maven central

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,6 @@ scalaVersion := scala212
 
 val commonSettings = Seq(
   organization := "com.lightbend.akka.grpc",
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -7,6 +7,7 @@ scalaVersion := scala212
 
 val commonSettings = Seq(
   organization := "com.lightbend.akka.grpc",
+  resolvers += "Staging Repo" at "https://oss.sonatype.org/content/repositories/comtypesafe-2276",
```
marcospereira: Leftover?

raboof: yeah this PR was also a 'smoke test' before we actually synced to maven central :D

marcospereira: But you can maybe add `https://oss.sonatype.org/content/repositories/releases/` so that we won't need to wait for Maven Central sync.

PR #481, opened 2018-12-13
expose akka version for suggested akka-stream dep in service traits

bjaglin:
In a multi-module environment, I am generating protobuf messages, fields & services in one `-api` module, and actual (Play server) gRPC implementation in the `-impl` one, and trying to keep dependencies to a minimum in the `-api` one.

Since generated service traits import `akka-stream` classes (for representing non-unary/streaming RPCs), I want to add it as a suggested dependency for my custom `ScalaCodeGenerator` that maps `ScalaCodeGenerator.generateServiceFile()` (in order to generate service traits only), and therefore reuse the actual version that the codegen needs instead of redefining it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,6 +32,7 @@ lazy val codegen = Project(
   .settings(Seq(
     buildInfoKeys ++= BuildInfoKey.ofN(organization, name, version, scalaVersion, sbtVersion),
     buildInfoKeys += "runtimeArtifactName" -> akkaGrpcRuntimeName,
+    buildInfoKeys += "akkaVersion" → Dependencies.Versions.akka,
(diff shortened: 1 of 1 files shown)


bjaglin:
@raboof good point about the anti-pattern sharing generated code. I realized my use-case was a bit more special/complex than described though: services defined in `-api` modules are exposed in `-impl` ones as gRPC servers (for external clients via Play) and/or as plain scala traits (injected in other build modules sharing the same JVM, bypassing the wire). For the second use case, I'd still want to generate the trait only (from the client module), so having the right/minimum runtime dependencies would be useful.

johanandren:
OK TO TEST

Issue #483, opened 2018-12-20
Exclude grpc-interop-testing from whitesource


raboof:
As it is only intended for internal use

dwijnand:
Should they even be published?

I mention it because one way to resolve this (if you're feeling generous) is via https://github.com/lightbend/sbt-whitesource/issues/53.  It auto-publishes from Travis CI, so the turnaround would be quick.

raboof:
If we can solve this by not publishing this subproject then that would indeed be the most elegant solution.

I have the sneaking suspicion that we are publishing them for a reason, though. We should definitely check ;).

dwijnand:
I suspect the same.

dwijnand:
Otherwise excluding is just a `whitesourceSkip := true` away.

raboof:
It is published because the 'scripted' integration tests rely on it. I think it is helpful to run the integration tests as scripted, as that is a little more realistic. I tried configuring sbt so that it'd include this subproject in `publishLocal` but not `publish`, but didn't find a way. I'll just exclude it from whitesource for now.

dwijnand:
(`publish := {}, deliver := file("")`)  Possibly also/just `publishTo := None`, you know know with these things.

raboof:
Tried a few combinations but that seems to make `sbt scripted` (which depends on the artifact of this subproject) fail

PR #484, opened 2018-12-20
Skip publishing interop-tests, update sbt-whitesource

raboof:
To clean up the dependency report, refs #483


dwijnand:
Can I interest you in some https://mergify.io/ experimentation in this repo?  It's the dog's bollocks.

raboof:
I'm a bit concerned about how much permissions we'd have to give such a tool. I know it's inevitable, and I don't suspect them of shenanigans themselves, but it does increase the attack surface...

Issue #252, opened 2018-06-25
Investigate performance of unary calls


patriknw:
The performance tests showed that the performance of unary calls wasn't great. Investigate and possibly improve.

patriknw:
Not top priority but we might do a timeboxed effort when other things of the MVP are under control

johanandren:
I'll spend a few hours on this now.

johanandren:
A few hours spent, findings:

Still no idea about what causes the worst case 4x degradation compared to java-grpc, although there seems to be a _lot_ of small things we could improve in general, both in Akka gRPC and the HTTP/2 server impl (tail on vector in header parsing, allocating 32k buffer for every outgoing response when compression is used etc).

Added support for benching without TLS in #278 and benched suspecting the Netty TLS stuff is much  faster than Jetty ALPN but that just showed more normal improvements. 

What I didn't get to in the time box I set out is profiling on Linux with flame graphs/native calls collected to see if there is something going on that YourKit doesn't catch.

johanandren:
I should also mention, I tried with the java-grpc benchmark client against the akka-grpc server to check if the 4x was on the client side, it was a bit faster but did not explain the 4x.

johanandren:
For future reference, this branch contains a few optimizations I tried out but didn't feel happy enough about to actually PR: https://github.com/johanandren/akka-grpc/tree/wip-252-unary-calls-performance-johanandren

PR #485, opened 2018-12-20
Respect TLS command line

assiotis:
When running the AsyncClient in akka.grpc.benchmarks.qps.AsyncClient, the `--tls` flag is meant to enable TLS, but if not specified, the expected behavior is to use plaintext (insecure) channel. However, the `GrpcClientSettings` case class defaults its `useTls` boolean to true. This PR simply sets up the GrpcClientSettings case class correctly based on the command line configuration.


lightbend-cla-validator:
Hi @assiotis,

raboof:
OK TO TEST

raboof:
FYI, we have not done extensive performance tuning yet, so we expect akka-grpc to be slower than other implementations at this point, but provide a nicer development experience. Some notes on performance can be found at #252.

PR #488, opened 2019-01-03
Bumping Year

hungaikev:


raboof:
OK TO TEST

raboof:
Hmm, in Akka we change `2018` to `2018-2019` which is rather neat (https://github.com/akka/akka/pull/26165), perhaps we could adapt that sbt-header configuration to work for akka-grpc also?

hungaikev:
Yap, that is neat. Let me work on that.

hungaikev:
Fixed with https://github.com/akka/akka-grpc/pull/489

PR #491, opened 2019-01-08
Cross publishing for SBT 0.13

andreaTP:
With this I add compatibility with SBT 0.13.X
I would need some guidance (or someone who can take over) to properly integrate publishing and scripted testing (the testing framework looks pretty elaborated, I hadn't spent on it enough to have a grasp).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ target
 .DS_Store
 /gradle-plugin/.gradle/
 /gradle-plugin/build/
+.metals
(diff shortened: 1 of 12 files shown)


raboof:
OK TO TEST

johanandren:
Are we sure we want to maintain support for 0.13? (I understand that it may be interesting to some people but sbt 1.x has been out for quite a while)

dwijnand:
Given sbt 0.13 itself is no longer supported (https://www.scala-sbt.org/0.13/docs/index.html) I don't think akka-grpc should endeavour to _start_ supporting it.

TimMoore:
Yes, I would rather focus efforts on removing any remaining obstacles preventing people from moving to 1.x.

andreaTP:
fair enough, I have had to make a try ....

Issue #448, opened 2018-10-29
Add a way to replace GrpcExceptionHandler with a custom implementation


KonstantinTsykulenko:
Currently GrpcExceptionHandler seems to be baked into the generated server code with no neat way to replace it. It would be nice to be able to provide a custom exception handler.

raboof:
Agreed! Would you be interested in looking into contributing this?

KonstantinTsykulenko:
Yep, I'll take a look and see what I can do.

ctoomey:
We'd like to have this too. @KonstantinTsykulenko  have you made any progress on this or do you expect to any time soon?

ctoomey:
Note than in addition to being able to customize exception handling for unary-response APIs, which `GrpcExceptionHandler` is used for, a separate means to customize exception handling for streaming-response APIs is desirable. This is currently hardcoded at https://github.com/akka/akka-grpc/blob/master/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala#L39 .

raboof:
I agree the exception handling for failed streams should also go through the (to-be-configurable) exception handler

ctoomey:
See https://github.com/akka/akka-grpc/pull/502

raboof:
Thanks @ctoomey for the contribution! @KonstantinTsykulenko could you have a look whether this is sufficient for your use case as well?

ctoomey:
@raboof happy to be able to contribute back what we need ourselves Thanks for merging.

wb14123:
Hi @ctoomey , is there any example about how to use a custom exception handler?

ctoomey:
@wb14123 I don't have a single code example handy, but an additional constructor taking a custom exception handler argument was added to the templates for the generated [Handler](https://github.com/akka/akka-grpc/pull/502/files#diff-89817a0a1f4a6377f320c5d7247026a84eeb520f85d772482e76520c1eb98a88) and [Router](https://github.com/akka/akka-grpc/pull/502/files#diff-747d8305ecef8da90c026953857ffd0200712ffce5989279b97becee9f7360fe). An example of such a custom exception handler is given in the [test](https://github.com/akka/akka-grpc/pull/502/files#diff-6e2b4d256de85e6481b31837bbc2cb086c0ebc12dc2b65557bb01c5bb1a0d0ad).

PR #502, opened 2019-01-19
#448 custom exception handler

ctoomey:
I've taken a stab at implementing this for both Scala and Java. As Scala is my primary language, suggest reviewing that first. I tried to make the Java implementation analogous.

### DIFF
--- a/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
@@ -21,6 +21,9 @@ import akka.grpc.ProtobufSerializer;
 import akka.grpc.javadsl.GoogleProtobufSerializer;
 import akka.grpc.javadsl.GrpcMarshalling;
 import akka.grpc.javadsl.GrpcExceptionHandler;
+import akka.grpc.javadsl.package$;
+
+import io.grpc.Status;
 
 import static @{service.packageName}.@{service.name}.Serializers.*;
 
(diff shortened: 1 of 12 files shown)


raboof:
OK TO TEST

runtime/src/test/scala/akka/grpc/akka/grpc/scaladsl/GrpcExceptionHandlerSpec.scala
```diff
@@ -0,0 +1,71 @@
+package akka.grpc.akka.grpc.scaladsl
```
raboof: This package name is rather strange - that seems unintentional? (and causes compilation problems, see CI)

ctoomey: Fixed

raboof:
Ah, I'm afraid travis now fails with `java.lang.RuntimeException: There are files without headers!` - you can generate copyright headers with `sbt headerCreate test:headerCreate`

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
```diff
@@ -7,31 +7,34 @@ package akka.grpc.javadsl
...
-    case _: UnsupportedOperationException ⇒
-      GrpcMarshalling.status(Status.UNIMPLEMENTED)
-    case other ⇒
-      println(other)
-      GrpcMarshalling.status(Status.INTERNAL)
+
+  def standard(t: Throwable): HttpResponse = standard(t, defaultMapper)
+
+  def standard(t: Throwable, mapper: jFunction[Throwable, Status]): HttpResponse = {
+    println(s"Caught exception $t")
```
RashadAnsari: Please do not println and use logger

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
```diff
@@ -7,31 +7,34 @@ package akka.grpc.javadsl
...
+        if (e.getCause == null) Status.INTERNAL
+        else defaultMapper(e.getCause)
+      case e: CompletionException ⇒
+        if (e.getCause == null) Status.INTERNAL
+        else defaultMapper(e.getCause)
+      case grpcException: GrpcServiceException ⇒ grpcException.status
+      case _: NotImplementedError ⇒ Status.UNIMPLEMENTED
+      case _: UnsupportedOperationException ⇒ Status.UNIMPLEMENTED
+      case other ⇒
+        println(other)
```
RashadAnsari: Please do not println and use logger

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
ctoomey: Fixed

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
ctoomey: Fixed

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -21,6 +21,9 @@ import akka.grpc.ProtobufSerializer;
 import akka.grpc.javadsl.GoogleProtobufSerializer;
 import akka.grpc.javadsl.GrpcMarshalling;
 import akka.grpc.javadsl.GrpcExceptionHandler;
+import akka.grpc.javadsl.package$;
```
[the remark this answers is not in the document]
ctoomey: I saw that `akka.japi.Function` was used consistently instead of `akka.japi.function.Function` in codegen/ and runtime/ so followed that example in using it for the exception handler. In this file we need the conversion to scala `PartialFunction` from the package object.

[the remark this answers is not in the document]
raboof: Yes, but using `akka.japi.Function` over `akka.japi.function.Function` might have been a mistake, as the javadocs on the former say "This class is kept for compatibility, but for future API's please prefer [[akka.japi.function.Function]]". But we can do that in a separate PR.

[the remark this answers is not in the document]
raboof: Added #513

raboof:
Refs #448

raboof:
Good improvement, thanks!

PR #509, opened 2019-01-24
add netty channel builder overrides

EthanMMF:
Power users should have control over the netty channel configurations at this can drastically improve performance.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -15,8 +15,8 @@ import com.typesafe.config.{ Config, ConfigValueFactory }
 import com.typesafe.sslconfig.akka.util.AkkaLoggerFactory
 import com.typesafe.sslconfig.ssl.{ ConfigSSLContextBuilder, DefaultKeyManagerFactoryWrapper, DefaultTrustManagerFactoryWrapper, SSLConfigFactory, SSLConfigSettings }
 import io.grpc.CallCredentials
+import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
 import javax.net.ssl.SSLContext
-
 import scala.collection.immutable
 import scala.concurrent.duration.{ Duration, _ }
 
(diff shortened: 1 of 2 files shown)


raboof:
OK TO TEST

raboof:
Thanks for starting this conversation!

A consequence of this change in its current implementation is that it makes `io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder` part of the Akka gRPC API, which I'm not sure we want to do: I think we want to keep the API relatively high-level so we keep the possibility to for instance replace Netty with Akka HTTP for the client in the future. What are the main things you would typically want to customize on the `NettyChannelBuilder` to improve performance? Perhaps those should become defaults, or some might be reasonable to expose as more high-level/abstract settings?

It would be nice if this could remove the need for a `withChannelBuilderOverrides` for most (even poweruser) cases - though perhaps for rare cases we might indeed still consider including it, clearly marked as a part of the API that might go away in the future?

EthanMMF:
Hi, hope this helps!

I was thinking between putting the configurations in GrpcClientSettings (or another high level settings) and the implementation as seen in my PR.

I think that the use case for different power users would vary with the implementation, especially for something lower level such as channels. Thus, maintenance for the different use cases would not be simple. 

On my end I would like control over some of the timeout values, as the default values did not suit my needs. But I felt that was too specific. Thus the more general approach was to allow use to override the channel builder to their needs.

The need and ability to override the channel builder should be ultimately based on the underlying implementation chosen (Netty or Akka HTTP etc.), and it might make more sense to deprecate this feature when the underlying implementation changes.

raboof:
OK, I think I can live with exposing this, but I think it would be good to explicitly mark `withChannelBuilderOverrides` with `@ApiMayChange` and clarify it may go away in the future.

I'm not sure why CI is failing (on the scripted 'cross-scala' test), that seems unrelated to this PR.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -212,6 +213,12 @@ final class GrpcClientSettings private (
   def withConnectionAttempts(value: Int): GrpcClientSettings =
     copy(connectionAttempts = Some(value))
 
+  /**
+    * To override any default channel configurations used by netty. Only for power users.
+    */
```
raboof: Oh, the build failure is because this comment is not formatted according to our current style guide: the last 2 lines should go 1 character to the left. Compiling from sbt (or running `sbt scalariformFormat`) should do this automatically for you. We'll look into making that more obvious.

ethan-redmart: @raboof thanks for the tip, was wondering why it failed too

Issue #510, opened 2019-01-25
UNAVAILABLE error status when return Future.failed


amsjavan:
In RPC handler when I return Future.failed, it will cause netty io exception and client gets an error with UNAVAILABLE status.

PR #511, opened 2019-01-25
Returning INTERNAL error status for non GrpcServiceException #510

amsjavan:


runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
```diff
@@ -26,7 +26,8 @@ object GrpcExceptionHandler {
     case _: UnsupportedOperationException ⇒
       Future.successful(GrpcResponseHelpers.status(Status.UNIMPLEMENTED))
     case other ⇒
-      Future.failed(other)
+      other.printStackTrace()
```
johanandren: We'd want this to go in a logger, not print to stdout.

runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
```diff
@@ -26,7 +26,8 @@ object GrpcExceptionHandler {
...
+      Future.successful(GrpcResponseHelpers.status(Status.INTERNAL))
```
johanandren: I think this makes sense.

johanandren:
OK TO TEST

runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
raboof: Agreed.

In Akka HTTP, the exception handler is a `PartialFunction[Throwable, Route]`, so you can get the logger from inside the `Route`. Route is a concept that is defined outside of akka-http-core, though, so we don't have access to it here.

The alternative is to change `val default` to `def default(...)` and close over the logger passed in there. That is still less flexible than doing it in the `Route` (the exception handler doesn't have access to the request context), but is likely sufficient? /cc @jrudolph

runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
raboof: Agreed. In a future version we might want to add further details to the response with `Status.INTERNAL.withDescription(...)`, but that seems outside the scope of this PR.

runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
amsjavan: I have done it

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
```diff
@@ -7,12 +7,16 @@ package akka.grpc.javadsl
...
-import scala.concurrent.ExecutionException
 import akka.http.javadsl.model.HttpResponse
 import io.grpc.Status
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.ExecutionException
 
 object GrpcExceptionHandler {
+
+  private val log = LoggerFactory.getLogger(getClass)
```
raboof: Rather than creating an slf4j logger statically, would it be possible to pass in an `akka.event.LoggingAdapter`? That might add a little more complexity, but also make things more flexible.

amsjavan: I prefer less complexity to more flexibility. It would not be better if we needed to do this?

raboof:
refs #510

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
raboof: This is indeed a tricky balance. Certainly as a first step it is usually smart to prefer simplicity over flexibility.

On the other hand, simplicity comes in many forms: isn't is simpler if akka-grpc-runtime depends only on Akka API's, and not also on the slf4j API's? In this case this could be especially relevant, since indeed akka-grpc-runtime doesn't introduce a dependency on slf4j, so people running without slf4j would get errors. The only reason this compiles is because we haven't had the time yet to move play-specific code to a separate project, so play (and thus slf4j) is included as 'provided' in akka-grpc-runtime.

raboof: Of course you could argue slf4j is so common it should always be included. This was not so clear when Akka started, but indeed we are actively considering making the slf4j dependency mandatory in Akka and having it replace some of Akka's own API's (https://github.com/akka/akka-meta/issues/31). We have not yet determined what would be the best way to achieve that, however, and there are a couple of features of Akka's logging subsystem for which it is not completely obvious how they could be replaced with slf4j.

raboof: Until that time I think it would be better to avoid slf4j and use the Akka API's in akka-grpc-runtime, however.

amsjavan: Ok I'll fix it

amsjavan:
Is it good?

raboof:
Nice contribution, thanks!

raboof:
I'm afraid there are some merge conflicts because we merged https://github.com/akka/akka-grpc/pull/502 in the mean time - could you have a look at resolving those (by rebasing or merging)?

amsjavan:
> I'm afraid there are some merge conflicts because we merged #502 in the mean time - could you have a look at resolving those (by rebasing or merging)?

Yes, of course, I'll do it.

raboof:
Thanks!

raboof:
Refs #510

Issue #515, opened 2019-01-29
Update to Akka 2.5.20


raboof:
Since we are using the Discovery API's which were not stable yet on 2.5.19, this needs a couple of changes.

PR #516, opened 2019-01-29
Akka 2.5.20 support

raboof:
Work in progress. Refs #515

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -9,7 +9,7 @@ object Dependencies {
     val scala211 = "2.11.12"
     val scala212 = "2.12.8"
 
-    val akka = "2.5.19"
+    val akka = "2.5.20"
     val akkaHttp = "10.1.5"
 
     val play = "2.7.0-RC3"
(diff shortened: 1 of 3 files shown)


johanandren:
One more API change in ResolvedTarget

project/Dependencies.scala
```diff
@@ -96,6 +97,7 @@ object Dependencies {
     Compile.akkaHttp2Support,
     Compile.akkaDiscovery,
     // these two are available when used through Play, which is also the only case when they are needed
+    Compile.akkaSlf4j % "provided",
```
ignasi35: The comment above seems to refer to:

```scala
    Compile.play % "provided",
    Compile.playAkkaHttpServer % "provided",
```
The comment should state `these three` or be moved one line down.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -74,13 +74,16 @@ object GrpcClientSettings {
...
+    var serviceName = clientConfiguration.getString("service-discovery.service-name")
     val port = clientConfiguration.getInt("port")
     val resolveTimeout = clientConfiguration.getDuration("service-discovery.resolve-timeout").asScala
     val sd = serviceDiscoveryMechanism match {
       case "static" =>
         val host = clientConfiguration.getString("host")
         require(host.nonEmpty, "host can't be empty when service-discovery-mechanism is set to static")
+        // Required by the Discovery infrastructure, even when we use static discovery.
+        if (serviceName.isEmpty)
+          serviceName = "static"
```
ignasi35: Oh wow. Sneaky default is sneaky.

project/Dependencies.scala
renatocaval: moved one line down

PR #518, opened 2019-01-29
[WIP] Scala 2.13.0-M5 support

### LINKED ISSUE #506: Build for scala 2.13 as well
Depends on
* [x] https://github.com/playframework/playframework/pull/8998 
* [x] https://github.com/scalapb/ScalaPB/issues/540

raboof:
Work in progress
Depends on https://github.com/scalapb/ScalaPB/issues/540 (milestone released) first
Fixes #506

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,7 +15,7 @@ addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.4")
 // scripted testing
 libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
 
-libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.8.4"
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.9.0-M1"
 
 // #java-agent-plugin
 addSbtPlugin("com.lightbend.sbt" % "sbt-javaagent" % "0.1.5")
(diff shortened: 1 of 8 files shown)


project/plugins.sbt
```diff
@@ -15,7 +15,7 @@ addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.5.4")
 // scripted testing
 libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
 
-libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.8.4"
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.9.0-M1"
```
raboof: Let's not merge before 0.9.0 is properly released

marcospereira: Can we instead merge and release a milestone version? Or maybe have a snapshot release. That way, we can also work on Scala 2.13 in play-grpc.

dwijnand: @raboof what about enabling publishing from pull requests?  that way we can consume this change and iterate in a PR in play-grpc, and then land both PRs when scalapb has a stable release.

raboof: I don't think we should publish every commit to every pull request, but indeed publishing snapshots is fine - I `sbt +publish`ed this branch, introducing 0.6.0+11-23237fc2 on bintray.

dwijnand: OK, thanks.

raboof:
Updated to M2, but this does not appear to work under JDK 8 - I suspect this is unintended, filed https://github.com/scalapb/ScalaPB/issues/567

raboof:
(to be revisited when 2.13.0-RC2 is out)

raboof:
superseded by #585

PR #522, opened 2019-02-01
remove a workaround of duplicated proto

jyane:
I think I've resolved this problem.

See also:
https://github.com/grpc/grpc-java/pull/4282

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -138,16 +138,6 @@ lazy val interopTests = Project(
     // sbt scripted test
     whitesourceIgnore := true,
   )
-  // proto files from "io.grpc" % "grpc-interop-testing" contain duplicate Empty definitions;
-  // * google/protobuf/empty.proto
-  // * io/grpc/testing/integration/empty.proto
-  // They have different "java_outer_classname" options, but scalapb does not look at it:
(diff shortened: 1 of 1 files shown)


raboof:
OK TO TEST

build.sbt
```diff
@@ -138,16 +138,6 @@ lazy val interopTests = Project(
...
-  // proto files from "io.grpc" % "grpc-interop-testing" contain duplicate Empty definitions;
-  // * google/protobuf/empty.proto
-  // * io/grpc/testing/integration/empty.proto
-  // They have different "java_outer_classname" options, but scalapb does not look at it:
-  // https://github.com/scalapb/ScalaPB/issues/243#issuecomment-279769902
-  // Therefore we exclude it here.
-  .settings(
-    excludeFilter in PB.generate := new SimpleFileFilter(
-      (f: File) => f.getAbsolutePath.endsWith("google/protobuf/empty.proto"))
-  )
```
johanandren: So this isn't needed anymore because the `grpc-interop-testing` does not contain duplicates in `1.16.1`?

jyane: @johanandren, yep :)
@2m Is it right?

2m: Seems to be alright. According to `grpc/grpc-java` issue tracker this was fixed in `1.12`.

PR #524, opened 2019-02-06
[WIP]: Fix code generation problems for Play

### LINKED ISSUE #523: Generate invalid code for server side in play-java
Akka gRPC Version: 0.5.0

The PR #502 and #511 introduce some conflict changes between `akka-grpc/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt` and `akka-grpc/runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala`.
These conflict changes haven't been fixed in 0.5.0, which make akka-grpc generate invalid code for play-java.

For example, in `Router.scala.txt`：
```
  public Abstract@{service.name}Router(Materializer mat) {
    this(mat, GrpcExceptionHandler.defaultMapper());
  }

  public Abstract@{service.name}Router(Materializer mat, Function<Throwable, Status> eHandler) {
    this.eHandler = eHandler;
    super(mat, @{service.name}.name);
  }
```
* `GrpcExceptionHandler.defaultMapper()` will return `Function<ActorSystem, Function<Throwable, Status>>` instead of `Function<Throwable, Status>`
* `super(..)` must be first line in ctor
* ref:
  * https://github.com/akka/akka-grpc/blame/master/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt

marcospereira:
Release 0.5.0 is not generating valid code. This fixes the templates to adapt them to the recent API changes.

Fixes #523.

## Status

WIP since the code now compiles, but the tests for play-grpc fail when using a local build of akka-grpc. To reproduce, just check out https://github.com/playframework/play-grpc/pull/45 and run the tests.

### DIFF
--- a/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
+++ b/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
@@ -15,6 +15,8 @@ import scala.concurrent.Future;
 import akka.http.scaladsl.model.HttpRequest;
 import akka.http.scaladsl.model.HttpResponse;
 import akka.grpc.javadsl.GrpcExceptionHandler;
+
+import akka.actor.ActorSystem;
 import akka.stream.Materializer;
 
 import akka.grpc.internal.PlayRouter;
@@ -25,15 +27,17 @@ import io.grpc.Status;
  * Abstract base class for implementing @{service.name} in Java and using as a play Router
  */
 public abstract class Abstract@{service.name}Router extends PlayRouter implements @{service.name} {
-  private final Function<Throwable, Status> eHandler;
+  private final ActorSystem actorSystem;
+  private final Function<ActorSystem, Function<Throwable, Status>> eHandler;
 
-  public Abstract@{service.name}Router(Materializer mat) {
-    this(mat, GrpcExceptionHandler.defaultMapper());
+  public Abstract@{service.name}Router(Materializer mat, ActorSystem actorSystem) {
+    this(mat, actorSystem, GrpcExceptionHandler.defaultMapper());
   }
 
-  public Abstract@{service.name}Router(Materializer mat, Function<Throwable, Status> eHandler) {
-    this.eHandler = eHandler;
+  public Abstract@{service.name}Router(Materializer mat, ActorSystem actorSystem, Function<ActorSystem, Function<Throwable, Status>> eHandler) {
     super(mat, @{service.name}.name);
+    this.eHandler = eHandler;
+    this.actorSystem = actorSystem;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
```diff
@@ -25,23 +27,25 @@ import io.grpc.Status;
  * Abstract base class for implementing @{service.name} in Java and using as a play Router
  */
 public abstract class Abstract@{service.name}Router extends PlayRouter implements @{service.name} {
-  private final Function<Throwable, Status> eHandler;
+  private final ActorSystem actorSystem;
```
marcospereira: Maybe it would be better to change `PlayRouter` to also requires an `ActorSystem` and not only a `Materializer`. But decide not to change it here to keep the impact of this PR small.

codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
```diff
@@ -25,23 +27,25 @@ import io.grpc.Status;
...
+    this.actorSystem = actorSystem;
   }
 
   /**
    * INTERNAL API
    */
   final public scala.Function1<HttpRequest, Future<HttpResponse>> createHandler(String prefix, Materializer mat) {
      return akka.grpc.internal.PlayRouterHelper.handlerFor(
-        @{service.name}HandlerFactory.create(this, prefix, mat, eHandler)
+        @{service.name}HandlerFactory.create(this, prefix, mat, eHandler, this.actorSystem)
```
marcospereira: It also makes more sense to have the exception handler as the last parameter, but again, avoid pilling up changes here.

codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
johanandren: I don't get this comment, because it has been changed to also require an actor system now.

johanandren:
OK TO TEST

codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
marcospereira: Sorry, I mean here, at the constructor:

https://github.com/akka/akka-grpc/blob/09de88af1f12a698eda61477e4c2763ed81bf8fb/runtime/src/main/scala/akka/grpc/internal/PlayRouter.scala#L41

Should it instead be:

```scala
@InternalApi abstract class PlayRouter(mat: Materializer, actorSystem: ActorSystem, serviceName: String) extends play.api.routing.Router
```

?

In the end, user extending the generated code will need to inject an `ActorSystem` anyway.

marcospereira:
> LGTM I guess

Thanks for reviewing, @johanandren. I will keep this as WIP until I have time to investigate why the tests for play-grpc are failing.

raboof:
@marcospereira I noticed akka-grpc was still depending on a RC of Play, added #531 to update.

raboof:
(looking at the conflicts now)

raboof:
Hmm, it seems like the changes in the PR were already part of #429 so are already on master - so this PR is perhaps no longer necessary?

marcospereira:
@raboof yes, we can close this.

I've created #533 to handle a small problem when trying a local build of akka-grpc in play-grpc. There are still some tests failing there though.

marcospereira:
By the way, you can still use https://github.com/playframework/play-grpc/pull/45 to see the tests failures.

PR #528, opened 2019-02-25
Remove leftover comment

raboof:


codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -14,8 +14,6 @@ import akka.grpc.gen.scaladsl.play.{ PlayScalaClientCodeGenerator, PlayScalaServ
 
 // This is the protoc plugin that the gradle plugin uses
 object Main extends App {
-  //  throw new RuntimeException(s"beginning of Main")
```
ctoomey: Sorry about this, thanks for catching and fixing.

PR #533, opened 2019-02-26
Re-add generators objects

marcospereira:
Power users can create instances of the case classes with `powerApis = true`, so no need to add another object here.

This keeps things consistent with how the generators are used in `build.sbt`.


raboof:
OK TO TEST

PR #447, opened 2018-10-29
More idiomatic play integration #420

ctoomey:
Here's a draft, scala-only implementation of #420 that's built on top of the changes for #179 in PR #429.

Per discussion in #429 (which this was originally included in), I've improved the implementation of the Play action-returning version of the code to reuse the std. auto-generated handler rather than duplicate much of it (see `codegen/src/main/twirl/templates/PlayScala/RouterUsingActions.scala.txt`).

Issue #420, opened 2018-09-28
More idiomatic Play integration


ctoomey:
I lead a team of long-time Play Scala  developers and we're about to start using grpc instead of REST for new APIs, as I've been discussing in #179.

Currently the auto-generated akka-grpc Play handler that wraps the grpc service method implementations returns a `Future[HttpResponse]` rather than an idiomatic Play `EssentialAction`. This means among other things that Play `Filter`s don't get called on each request as they do for REST calls, and so cross-cutting Play functionality like access logging provided by filters doesn't work.

To gain back such functionality for grpc requests, developers would need to not only port their Play filters to some new home-grown filter-like interface (since a replacement isn't provided for Play filters by akka-grpc), but also design and implement a home-grown filter chain interface.

At the very least akka-grpc should provide a Play integration that supports the existing Play idioms in addition to one like the current that doesn't.

I've taken the existing auto-generated code and akka-grpc code base and modified both so as to have grpc service implementations wrapped to return `EssentialActions` and verified that our filters work as expected. I'd love to contribute these changes back either as a replacement for or additional option for Play users. What do you say?

johanandren:
Sounds like it makes sense, please do PR so we can look at and discuss the details!

ignasi35:
There are several possible approaches:

1. add a capability in `akka-grpc` for request filtering
   1. make it low-level (more familiar to akka-http users). Using `Directive`s  (as hinted in #179)
   2. make it high-level (more familiar to play users). In this case, I think we should still make it play-agnostic and part of the `akka-grpc` library. Maybe refactoring `PlayRouter` to inject akka-grpc filters?
2. improve [Play's filtering code](https://github.com/playframework/playframework/blob/da68fc546da305b8fc215d0b8ad53b644b9b843c/framework/src/play/src/main/scala/play/api/http/HttpRequestHandler.scala#L222) so all requests are filtered (not just `EssentialAction`'s) 

 - - - - - - - 

As discussed internally, when adding a filter on some akka-grpc router in a Play Application there would be some limited features:

* the filter MAY block or pass the request
* the filter MAY NOT alter the request (any modification will be ignored). The reason for this limitation is that the generated code in Akka-gRPC ignores the play request and uses the _raw_ akka-http request.

 - - - - - - - 

The final concern, from a Play user point of view, is how would filters be enabled/injected. I haven't given any thought.

ignasi35:
@ctoomey looking forward to your code.

[...]

PR #534, opened 2019-02-28
More idiomatic play integration #420

ctoomey:
Continuation of PR #447, using a new branch with single composite commit.


johanandren:
Refs #420

raboof:
OK TO TEST

codegen/src/main/twirl/templates/PlayJavaServer/RouterUsingActions.scala.txt
```diff
@@ -0,0 +1,80 @@
...
+      this.system = system;
+      this.parsers = parsers;
+      this.actionBuilder = actionBuilder;
+    }
+
+    @@Override
+    public Function1<RequestHeader, EssentialAction> createHandler(String serviceName, Materializer mat) {
+      return new Function1<RequestHeader, EssentialAction>() {
+        public EssentialAction apply(RequestHeader reqOuter) {
+          Function<HttpRequest, CompletionStage<HttpResponse>> handler = @{serviceName}HandlerFactory.create(Abstract@{serviceName}Router.this, serviceName, mat, eHandler, system);
```
raboof: If I'm reading this correctly, this would now be called for every request. It looks like that would not be necessary, and it could be created once outside of the function. Would that work?

runtime/src/main/scala/akka/grpc/internal/PlayRouterUsingActions.scala
```diff
@@ -0,0 +1,141 @@
...
+    else
+      throw new UnsupportedOperationException(
+        "Prefixing gRPC services is not widely supported by clients, " +
+          s"strongly discouraged by the specification and therefore not supported. " +
+          s"Attempted to prefix with [$prefix], yet already default prefix known to be [${this.prefix}]. " +
+          s"When binding gRPC routers the path in `routes` MUST BE `/`.")
+
+  def createStreamingAction(handler: HttpRequest => Future[HttpResponse])(implicit ec: ExecutionContext): EssentialAction =
+    actionBuilder.async(streamBodyParser) { req =>
+      handler(playToAkkaRequestStream(req)).map(akkaToPlayResp)
```
raboof: As discussed in https://github.com/akka/akka-grpc/issues/420#issuecomment-426630873, this 'additional roundtrip' is of course more or less unavoidable if you want to take into account the changes to the Play request, but does come at some performance cost.

How much performance cost, and whether that cost is acceptable, of course highly depends on the application and should be measured/considered by the user.

I think we should add some documentation to this option, and explain clearly that the trade-off here: this option allows you to use your existing Play action composition approach, at the expense of some performance (which is why it is not the default). Later, eventually, we could also document alternate ways to achieve cross-cutting concerns there.

codegen/src/main/twirl/templates/PlayJavaServer/RouterUsingActions.scala.txt
ctoomey: No, this is only called once during initialization. See [here](https://github.com/akka/akka-grpc/pull/534/files#diff-7e9bbe2f58513cfce8675cfe00c583d7R44).

It shouldn't be public though, will change that.

raboof: Sure, `createHandler` is only called once, but line 66 is inside the function that is returned by `createHandler`, and that is called for every request - unless I'm misreading this somehow, it could be moved to between line 64 to only create the handler once? (haven't tried though)

ctoomey: Ah right, I was looking at line 64. Will fix.

ctoomey: Fixed

runtime/src/main/scala/akka/grpc/internal/PlayRouterUsingActions.scala
ctoomey: Currently there's no mention of the Play integration in the docs. Where do you suggest documenting this?

raboof: Ah, the Play gRPC documentation has already moved to https://github.com/playframework/play-grpc - I guess we will have to move the documentation for this feature to a separate PR over there.

ctoomey: Ok, I'll work on that before long, but you can go ahead and merge this right?

ctoomey:
Hi @marcospereira , when do you think you can review this? Want to get closure on this odyssey.

ctoomey:
Hi @raboof , nothing's happening from your added reviewer and I'd like this to get merged and done with. Please advise.

raboof:
This is now good to go - thanks for another great contribution! :tada:

ctoomey:
Yay! Thanks @raboof .

PR #542, opened 2019-03-04
Fix initialization order problem in Play Java router code generation

### LINKED ISSUE #396: Play Java clients should 500 on an empty gRPC request
In #391 when creating [PlayJavaFunctionalTest](https://github.com/akka/akka-grpc/blob/9dc5475783183d91c214f6ea581e23444a7a5079/play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java), based on [PlayScalaTestSpec](https://github.com/akka/akka-grpc/blob/9dc5475783183d91c214f6ea581e23444a7a5079/play-testkit/src/test/scala/akka/grpc/scalatestplus/play/PlayScalaTestSpec.scala) and [PlaySpecs2Spec](https://github.com/akka/akka-grpc/blob/9dc5475783183d91c214f6ea581e23444a7a5079/play-testkit/src/test/scala/akka/grpc/play/api/specs2/PlaySpecs2Spec.scala), I discovered that where in Scala a Play server bound to a gRPC router gives a 500 when routing an empty request to a gRPC method, in Java it gives a 200!

For now I resolved it by [`@Ignore`ing the test method](https://github.com/akka/akka-grpc/blob/9dc5475783183d91c214f6ea581e23444a7a5079/play-testkit/src/test/java/akka/grpc/play/test/PlayJavaFunctionalTest.java#L78-L81).

raboof:
Before, the exception handler field would be initialized in the constructor,
but the method reading the field was called from the superconstructor so saw a
null value here, leading to a null exception handler further down the road.

Fixes #396

### DIFF
--- a/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
+++ b/codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
@@ -27,26 +27,21 @@ import io.grpc.Status;
    * Abstract base class for implementing @serviceName in Java and using as a play Router
    */
   public abstract class Abstract@{serviceName}Router extends PlayRouter implements @{serviceName} {
-    private final Function<ActorSystem, Function<Throwable, Status>> eHandler;
-    private final ActorSystem system;
+    public scala.Function1<HttpRequest, Future<HttpResponse>> respond;
 
     public Abstract@{serviceName}Router(Materializer mat, ActorSystem system) {
       this(mat, system, GrpcExceptionHandler.defaultMapper());
     }
 
     public Abstract@{serviceName}Router(Materializer mat, ActorSystem system, Function<ActorSystem, Function<Throwable, Status>> eHandler) {
-      super(mat, @{service.name}.name);
-      this.eHandler = eHandler;
-      this.system = system;
+      super(@{service.name}.name);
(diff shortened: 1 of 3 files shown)


codegen/src/main/twirl/templates/PlayJavaServer/Router.scala.txt
```diff
@@ -27,26 +27,21 @@ import io.grpc.Status;
    * Abstract base class for implementing @serviceName in Java and using as a play Router
    */
   public abstract class Abstract@{serviceName}Router extends PlayRouter implements @{serviceName} {
-    private final Function<ActorSystem, Function<Throwable, Status>> eHandler;
-    private final ActorSystem system;
+    public scala.Function1<HttpRequest, Future<HttpResponse>> respond;
```
dwijnand: I have 3 questions, should this be:
* public
* non-final
* scala.Function1 (in PlayJavaServer)
?

raboof: You're right, changing public and non-final to `private final`.

`scala.Function1` seems OK since the user should never interact with it directly, and it's what's expected by the superclass `PlayRouter` (which is shared among Scala and Java API).

PR #544, opened 2019-03-06
Example of cross-cutting authentication/authorization example

raboof:
This is an example of a way to compose Akka gRPC with 'regular' Akka HTTP code
to achieve server-side authentication/authorization when it is a true
'cross-cutting concern': the implementation is still based on the regular API
trait (and does not need to know about the token), and is wrapped in a
Directive that performs the token verification.

This is not a complete example: documentation, a Java version and test coverage
are still missing, and authentication errors are reported as a 'normal' HTTP
400 response rather than a gRPC error code 7. Nonetheless it might be a good
starting point for further exploration.

### DIFF
--- a/plugin-tester-scala/src/main/scala/example/myapp/helloworld/AuthenticatedGreeterServer.scala
+++ b/plugin-tester-scala/src/main/scala/example/myapp/helloworld/AuthenticatedGreeterServer.scala
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2018-2019 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+//#full-server
+package example.myapp.helloworld
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.UseHttp2.Always
+import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
+import akka.http.scaladsl.server.{ Directive0, Route, RouteResult }
+import akka.http.scaladsl.server.Directives._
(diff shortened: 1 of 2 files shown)


plugin-tester-scala/src/main/scala/example/myapp/helloworld/AuthenticatedGreeterServer.scala
```diff
@@ -0,0 +1,87 @@
...
+    implicit val ec: ExecutionContext = sys.dispatcher
+
+    // Create service handlers
+    val handler: HttpRequest => Future[HttpResponse] =
+      GreeterServiceHandler(new GreeterServiceImpl())
+
+    // As a Route
+    val handlerRoute: Route = {
+      ctx => handler(ctx.request).map(RouteResult.Complete)
+    }
```
johanandren: Easy enough but perhaps we could provide a `handler.asDirective`? (not scope of this PR though)

raboof: I agree it would be nice if this would be easier. The reason this is not currently there is because we currently only depend on `akka-http-core`, and directives are an `akka-http` concept - so that's why those types are not used there. Perhaps we should just pull in all of akka-http at some point, but that's outside the scope of this PR ;).

johanandren: Ah, didn't think of that, d'oh!

Issue #530, opened 2019-02-26
Run the interop tests on jdk11


raboof:
Since ALPN is now included in the JDK, setting up the interop tests would have to be different when running on that.

dwijnand:
Some progress in #551.

dwijnand:
Failures on Java 11:

```
[0m[[0m[31merror[0m] [0m[0mFailed tests:[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaScalaWithIoSpec[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaJavaWithIoSpec[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaJavaWithAkkaScalaSpec[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaJavaWithAkkaJavaSpec[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaScalaWithAkkaScalaSpec[0m
[0m[[0m[31merror[0m] [0m[0m	akka.grpc.interop.GrpcInteropAkkaScalaWithAkkaJavaSpec[0m
[0m[[0m[31merror[0m] [0m[0m(akka-grpc-interop-tests / Test / [31mtest[0m) sbt.TestsFailedException: Tests unsuccessful[0m
[0m[[0m[31merror[0m] [0m[0mTotal time: 238 s, completed Mar 25, 2019, 10:32:52 AM[0m
```

(from https://api.travis-ci.org/v3/job/510854812/log.txt #551's merge commit)

raboof:
Fixed in #551 and #568

PR #551, opened 2019-03-25
Add AdoptOpenJDK 11 to CI

dwijnand:
... and cache all of ~/.sbt (particularly for ~/.sbt/launchers).

Refs #530.


.travis.yml
```diff
@@ -7,6 +7,19 @@ scala:
...
 
+env:
+  matrix:
+    - TRAVIS_JDK=adopt@1.8.202-08
+    - TRAVIS_JDK=adopt@1.11.0-2
+
+matrix:
+  fast_finish: true
+  allow_failures:
+    - env: TRAVIS_JDK=adopt@1.11.0-2 # not fully supported but allows problem discovery
```
raboof: Should we leave open #530 to fix these? Or introduce another ticket?

.travis.yml
```diff
@@ -49,7 +62,8 @@ before_cache:
 cache:
   directories:
     - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot
+    - $HOME/.jabba/jdk
+    - $HOME/.sbt
```
raboof: :+1:

.travis.yml
dwijnand: I don't mind.  Which do you prefer?

raboof: No preference - let's keep #530 open for this then?

PR #572, opened 2019-04-17
Update io.grpc to 1.20.0

raboof:


renatocaval:
I'm adding this one to 0.6.1 so we can track it. 

Play gRPC will update when this gets released.

PR #579, opened 2019-05-13
Docs: more settings to improve the generated site

ennru:
## Purpose

Enable version warnings, canonical URLs, description tags via Paradox.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,8 +5,12 @@ import akka.grpc.build.ReflectiveCodeGen
 
 scalaVersion := scala212
 
+ThisBuild / organization := "com.lightbend.akka.grpc"
+ThisBuild / organizationName := "Lightbend Inc."
+ThisBuild / organizationHomepage := Some(url("https://www.lightbend.com/"))
+ThisBuild / description := "Akka gRPC - Support for building streaming gRPC servers and clients on top of Akka Streams."
+
 val commonSettings = Seq(
-  organization := "com.lightbend.akka.grpc",
   scalacOptions ++= List(
     "-unchecked",
     "-deprecation",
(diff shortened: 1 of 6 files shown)


docs/src/main/paradox/_template/projectSpecificFooter.st
```diff
@@ -0,0 +1,4 @@
+<script type="text/javascript" src="$page.base$assets/js/warnOldVersion.js"></script>
+<script type="text/javascript">//<![CDATA[
+jQuery(function(jq){initOldVersionWarnings(jq, '$page.properties.("project.version")$', '$page.properties.("project.url")$')});
+//]]></script>
```
raboof: Should these eventually move to https://github.com/akka/akka-paradox or a separate paradox plugin?

build.sbt
```diff
@@ -5,8 +5,12 @@ import akka.grpc.build.ReflectiveCodeGen
 
 scalaVersion := scala212
 
+ThisBuild / organization := "com.lightbend.akka.grpc"
+ThisBuild / organizationName := "Lightbend Inc."
+ThisBuild / organizationHomepage := Some(url("https://www.lightbend.com/"))
+ThisBuild / description := "Akka gRPC - Support for building streaming gRPC servers and clients on top of Akka Streams."
```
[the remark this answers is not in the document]
ennru: I believe so, we could add a `Common` auto-plugin as we do in other projects to centralise settings.

docs/src/main/paradox/_template/projectSpecificFooter.st
ennru: Yes, I get tired of copying it everywhere. But there is Akka's more advanced version handling in the template, isn't it?

ennru:
I introduced a `Common` auto-plugin and moved the shared settings there. This aligns with what we do in most other projects.

project/Common.scala
```diff
@@ -0,0 +1,39 @@
...
+    developers += Developer("contributors",
+      "Contributors",
+      "https://gitter.im/akka/dev",
+      url("https://github.com/akka/akka-grpc/graphs/contributors")),
+    licenses := Seq("Apache-2.0" -> url("https://www.apache.org/licenses/LICENSE-2.0")),
+    description := "Akka gRPC - Support for building streaming gRPC servers and clients on top of Akka Streams."
+  )
+
+  override lazy val projectSettings = Seq(
+    crossVersion := CrossVersion.binary,
```
raboof: What is this intended to do? This is what causes the travis failure: due to this setting the maven plugin gets `_2.12` appended, which it shouldn't

ennru: I don't see where I got that from. Thanks for noticing.

PR #581, opened 2019-05-21
Ensure authority override is applied when TLS isn't being used

jroper:
Not sure why the authority override setting was only being applied if using TLS, but it should be applied regardless of whether you're using TLS. This is necessary when speaking to reverse proxies on an IP address that don't have a hostname configured - eg, common when testing things like Istio and Knative in the cloud.


jroper:
Do you know where in the grpc-java implementation that might have come from? The workaround, to use:

```scala
.withChannelBuilderOverrides(_.overrideAuthority("example.com"))
```

works, I thought the `NettyChannelBuilder` was the client facing API for grpc-java.

raboof:
hmm, I think I was thinking of https://github.com/grpc/grpc-java/blob/ba335f5e68607a4aa8755671b4d9933863933321/interop-testing/src/main/java/io/grpc/testing/integration/TestServiceClient.java#L421, but that always applied the override. Thanks for the fix!

PR #583, opened 2019-05-24
Update sbt build for akka.io

ennru:
* Add sbt-site 1.4.0
* Update URLs to point to akka.io
* Use extref to point to Akka docs, Akka HTTP docs


build.sbt
```diff
@@ -135,23 +135,34 @@ lazy val docs = Project(
...
     Compile / paradoxProperties ++= Map(
+      "akka.version" -> Dependencies.Versions.akka,
+      "akka-http.version" -> Dependencies.Versions.akkaHttp,
       "grpc.version" → Dependencies.Versions.grpc,
-      "project.url" -> "https://developer.lightbend.com/docs/akka-grpc/current/",
-      "canonical.base_url" -> "https://developer.lightbend.com/docs/akka-grpc/current/",
-      "akka-http.version" → Dependencies.Versions.akkaHttp,
-      "extref.akka-http.base_url" -> s"https://doc.akka.io/docs/akka-http/${Dependencies.Versions.akkaHttp}/%s",
+      "project.url" -> "https://doc.akka.io/docs/akka-grpc/current/",
+      "canonical.base_url" -> "https://doc.akka.io/docs/akka-grpc/current",
```
2m: These two are read by default from the `homepage` setting. So it might be better to set it like in Alpakka: https://github.com/akka/alpakka/blob/master/project/Common.scala#L20

docs/src/main/paradox/overview.md
```diff
@@ -25,7 +25,7 @@ It features:
 
 * Akka gRPC $project.version$ ([Github](https://github.com/akka/akka-grpc))
 * Scala $scala.binary.version$
-* Akka Http $akka-http.version$ (@extref[Docs](akka-http:index.html), [Github](https://github.com/akka/akka-http))
+* Akka Http $akka-http.version$ (@extref[Docs](akka-http-docs:index.html), [Github](https://github.com/akka/akka-http))
```
2m: This change is because we have it like that in Alpakka. I would rather make it shorter in Alpakka instead.

build.sbt
ennru: But here we use `akka.io` as `homepage`.

docs/src/main/paradox/overview.md
ennru: It had `-docs` attached in others, as well. Just make it consistent...

build.sbt
2m: Ahh, alright.

Issue #517, opened 2019-01-29
Move Play-specific code to their own repo's


raboof:
A lot of the Play-specifics are already in a separate repository at https://github.com/playframework/play-grpc , but the code generation itself is still in akka-grpc.

It made sense to keep it close to the Akka code generation when the code generation API's were still evolving quickly, but we do plan to move them out. Since the Akka gRPC infrastructure allows contributing code generators from 'outside' this should be relatively feasible.

Perhaps good to finish #420 first, though.

dwijnand:
Who knows the Play generators and could help maintain them in play-grpc?

raboof:
I think at least myself, @renatocaval, @johanandren and (esp. when that PR is merged) @ctoomey?

ctoomey:
Yep I could help a bit after #420, which I'll tackle again once #179  is merged. As far as that goes I've done everything except update the documentation, which I'll do soon.

raboof:
The next step here would be to make the akka-grpc configuration API's powerful enough so Play generators can be inserted without Play-specific knowledge in akka-grpc.

This is already the case for sbt (via `akkaGrpcExtraGenerators`, i.e. https://developer.lightbend.com/docs/play-grpc/current/play/serving-grpc.html), and we should find a way to expose that for Gradle users (who currently use the `generatePlay`/`usePlayActions` gradle plugin settings) and Maven users (who currently use specific generatePlayClient/generatePlayServer/usePlayActions settings).

After that, we can move the generators themselves to their own `play-grpc-generators` artifact, and introduce a `play-grpc-runtime` to hold the parts of the runtime that hook into Play.

raboof:
this has been done

PR #585, opened 2019-05-27
Scala2.13.0 rc2

raboof:
Refs #506 

Depends on:
* [ ] https://github.com/scalapb/ScalaPB/issues/586
* [ ] Either https://github.com/playframework/playframework/issues/9336 or #517

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: scala
 
 scala:
   - 2.12.8
+  - 2.13.0-M5
 
 # make sure to fetch enough commits, so that git describe still works
 git:
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,5 @@
 import akka.grpc.Dependencies
-import akka.grpc.Dependencies.Versions.scala212
+import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
 import akka.grpc.ProjectExtensions._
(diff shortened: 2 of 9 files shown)


dwijnand:
> * [ ]  Either [playframework/playframework#9336](https://github.com/playframework/playframework/issues/9336) or #517

https://github.com/playframework/playframework/releases/tag/2.8.0-M1 is out (http://repo1.maven.org/maven2/com/typesafe/play/play-server_2.13.0-RC2/2.8.0-M1/), just not announced.

raboof:
That means requiring Play 2.8 though... we might not (yet?) want that?

dwijnand:
I don't think we'll be developing more releases of akka-grpc/play-grpc for Play 2.7/Lagom 1.5.  Onwards and upwards.

marcospereira:
Or we can wait since we plan to have Scala 2.13 for Play 2.7 (but we will backport and release only 2.13 is GA).

dwijnand:
Differently to Akka 2.5/2.6, Play 2.8 won't be binary compatible with Play 2.7, so we'll probably want to upgrade akka-grpc/play-grpc for Play 2.8.  But we can also do that closer to Play 2.8 coming out rather than right now, if preferred.  (Not what I would do.)

raboof:
superseded by #599

PR #589, opened 2019-05-28
Use protoc settings for server_power_apis and use_play_actions

raboof:
This is a step towards configuring whether or not to generate Power API's
on a service-by-service basis rather than globally, as well as specifying
and configuring external generators in Maven and Gradle


raboof:
(reviewed as part of #592)

PR #592, opened 2019-05-29
Allow loading external generators in Gradle plugin

raboof:
Depends on #589

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/Main.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/Main.scala
@@ -35,7 +35,16 @@ object Main extends App {
 
   private val generateServer: Boolean = !reqLowerCase.contains("generate_server=false")
 
-  private val generatePlay: Boolean = reqLowerCase.contains("generate_play=true")
+  /**
+   * For example `akka.grpc.gen.scaladsl.play.PlayScalaClientCodeGenerator`
+   */
+  val ExtraGeneratorsRegex = """(?:.*,)extra_generators=([^,]+)(?:,.*)?""".r
+  private val extraGenerators: List[String] = req.getParameter match {
+    case ExtraGeneratorsRegex(generators) =>
+      generators.split(";").toList
+    case _ =>
+      List.empty
+  }
 
   val LogFileRegex = """(?:.*,)logfile=([^,]+)(?:,.*)?""".r
   private val logger = req.getParameter match {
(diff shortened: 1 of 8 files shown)


codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -35,11 +35,16 @@ object Main extends App {
...
+  /**
+   * For example `akka.grpc.gen.scaladsl.play.PlayScalaClientCodeGenerator`
+   */
+  val ExtraGeneratorsRegex = """(?:.*,)extra_generators=([^,]+)(?:,.*)?""".r
+  private val extraGenerators: List[String] = req.getParameter match {
+    case ExtraGeneratorsRegex(generators) =>
+      generators.split(";").toList
+    case _ =>
+      List.empty
+  }
```
johanandren: Reaching a point where it would make sense to parse req into a `Map` perhaps

codegen/src/main/scala/akka/grpc/gen/javadsl/play/PlayJavaServerCodeGenerator.scala
```diff
@@ -4,35 +4,40 @@
...
 
+import scala.collection.immutable
 import akka.grpc.gen.Logger
 import akka.grpc.gen.javadsl.{ JavaCodeGenerator, Service }
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse
 import templates.PlayJavaServer.txt.Router
 import templates.PlayJavaServer.txt.RouterUsingActions
 
-case class PlayJavaServerCodeGenerator(powerApis: Boolean = false, usePlayActions: Boolean = false) extends JavaCodeGenerator {
+class PlayJavaServerCodeGenerator extends JavaCodeGenerator {
```
johanandren: shouldn't this also be abstract?

johanandren: shouldn't this also be abstract?

codegen/src/main/scala/akka/grpc/gen/javadsl/JavaServerCodeGenerator.scala
```diff
@@ -4,36 +4,49 @@
...
 package akka.grpc.gen.javadsl
 
+import scala.collection.immutable
 import akka.grpc.gen.{ BuildInfo, CodeGenerator, Logger }
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse
 import protocbridge.Artifact
 import templates.JavaServer.txt.{ Handler, PowerApiInterface }
 
-case class JavaServerCodeGenerator(powerApis: Boolean = false) extends JavaCodeGenerator {
+abstract class JavaServerCodeGenerator extends JavaCodeGenerator {
```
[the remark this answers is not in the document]
raboof: Actually I'm dropping the `abstract` again in a different PR, because I want them to be instantiatable by reflection from the class name.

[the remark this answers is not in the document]
raboof: Wait, this *is* that other PR - I forgot to remove the `abstract` here, sorry

codegen/src/main/scala/akka/grpc/gen/Main.scala
raboof: agree but will do in a different PR

PR #595, opened 2019-05-31
Deploy documentation from Travis

2m:
## Purpose

Adds Travis configuration and encrypted private key to deploy snapshot and release documentation to the documentations server.


.travis.yml
```diff
@@ -46,8 +48,8 @@ jobs:
       name: sbt publish
     - script: openssl aes-256-cbc -K $encrypted_9c9c33071881_key -iv $encrypted_9c9c33071881_iv -in gradle.properties.enc -out ./gradle-plugin/gradle.properties -d && cd gradle-plugin && ./gradlew publishPlugins
       name: gradle publish
-    - stage: techhub-ping
-      script: curl -I https://ci.lightbend.com/job/techhub-publisher/build?token=$TECH_HUB_TOKEN
+    - script: eval "$(ssh-agent -s)" && cp .travis/id_rsa /tmp/id_rsa && chmod 600 /tmp/id_rsa && ssh-keygen -p -P "$DEPLOY_PASSPHRASE" -N "" -f /tmp/id_rsa && ssh-add /tmp/id_rsa && sbt akka-grpc-docs/publishRsync
+      name: Publish documentation
```
raboof: Do we also want a stage that runs for each compile on master that updates the `snapshot` docs?

2m: This will do that as well. The docs destination directory is determined dynamically (https://github.com/akka/akka-grpc/blob/master/build.sbt#L144) so it will be "snapshot" for non-version-tagged builds.

raboof: ahh neat

PR #597, opened 2019-06-03
Make release checklist a little more lightweight

raboof:


docs/release-train-issue-template.md
```diff
@@ -10,26 +10,20 @@ Variables to be expanded in this template:
...
 
 - [ ] For non-patch releases: rename the 'akka-grpc-x.x-stable' and 'akka-grpc-supported-x.x-stable' reporting projects in [WhiteSource](https://saas.whitesourcesoftware.com/Wss/WSS.html) accordingly (unfortunately this requires permissions that cannot be shared outside of Lightbend)
-- [ ] Prepare release notes listing contributors generated by [`sbt-authors`](https://github.com/2m/authors) (eg. `sbt authors v0.22 HEAD`)
 - [ ] For non-patch releases: Create a news item draft PR on [akka.github.com](https://github.com/akka/akka.github.com), using the milestone
 - [ ] Move all [unclosed issues](https://github.com/akka/akka-grpc/issues?q=is%3Aopen+is%3Aissue+milestone%3A$AKKA_GRPC_VERSION$) for this milestone to the next milestone
+- [ ] Close the [milestone](https://github.com/akka/akka-grpc/milestones?direction=asc&sort=due_date)
 
 ### Cutting the release
 
-- [ ] Wait until [master build finished](https://travis-ci.org/akka/akka-grpc/builds/) after merging the release notes
```
ennru: It is important that no master build is in progress when tagging the release.

docs/release-train-issue-template.md
```diff
@@ -54,14 +48,14 @@ Variables to be expanded in this template:
 
 ### Announcements
 
-- [ ] For non-patch releases: Merge draft news item for [akka.io](https://github.com/akka/akka.github.com)
+For non-patch releases:
```
ennru: Does this imply no announcing at all for patch releases?

raboof: Optionally - though if there's something notable to mention, perhaps it should be a minor release anyway (at least until we hit 1.0.0)

raboof:
> Ticking boxes isn't that hard, is it?

Ticking boxes isn't too hard, but I feel there's way too many manual steps in the release process, making it high-friction and error-prone. For a 'large' project such as Scala or Akka that might be fine, but for a satellite project like Akka gRPC I think we should be careful. Ideally it should be little more than setting the tag...

ennru:
I agree that more automation is better. The idea with this template is to make it less error-prone, that's why we create an explicit issue for it. 
Have a look at my latest invention to make it easier in https://github.com/akka/alpakka-kafka/pull/825

PR #599, opened 2019-06-11
Scala2.13.0

raboof:
Depends on:
* [x] https://github.com/scalapb/ScalaPB/issues/593
* [x] #598

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,7 +54,7 @@ lazy val scalapbProtocPlugin = Project(
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(
       prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultShellScript)
     ),
-    crossScalaVersions -= scala213,
+    crossScalaVersions := Seq(scala212),
   ))
   .settings(addArtifact(artifact in (Compile, assembly), assembly))
 
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,5 @@
 import akka.grpc.Dependencies
-import akka.grpc.Dependencies.Versions.scala212
+import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
 import akka.grpc.ProjectExtensions._
 import akka.grpc.build.ReflectiveCodeGen
 
@@ -28,6 +28,7 @@ lazy val codegen = Project(
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(
       prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultShellScript)
     ),
+    crossScalaVersions -= scala213,
   ))
   .settings(addArtifact(artifact in (Compile, assembly), assembly))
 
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -65,6 +67,7 @@ lazy val mavenPlugin = Project(
   .settings(Seq(
     publishMavenStyle := true,
     crossPaths := false,
+    crossScalaVersions := Seq(scala212),
```
johanandren: Different style here vs above with removing 2.13 vs replacing with a list with only 2.12

build.sbt
```diff
@@ -162,7 +166,8 @@ lazy val docs = Project(
     ),
     resolvers += Resolver.jcenterRepo,
     publishRsyncArtifact := makeSite.value -> "www/",
-    publishRsyncHost := "akkarepo@gustav.akka.io"
+    publishRsyncHost := "akkarepo@gustav.akka.io",
+    crossScalaVersions := List(scala212, scala213),
```
johanandren: docs need cross versions?

plugin-tester-java/build.gradle
```diff
@@ -27,7 +27,7 @@ akkaGrpc {
   generateClient = true
   generateServer = true
   serverPowerApis = true // This isn't default
-  extraGenerators = [ 'akka.grpc.gen.scaladsl.play.PlayScalaClientCodeGenerator', 'akka.grpc.gen.scaladsl.play.PlayScalaServerCodeGenerator' ]
+  extraGenerators = [ ]
```
johanandren: How is this related to this PR? Does it also remove play modules, is it on top of the remove play modules-pr?

marcospereira: Apparently, this is built on top of #598. So maybe @raboof need to do a rebase here?

raboof: (:+1: rebased for smaller diff)

build.sbt
raboof: I think I did this because docs depends on cross-built modules, and in that case sbt requires the 'depending' module to also be cross-built (https://github.com/sbt/sbt/issues/3465)

build.sbt
raboof: good point, changing to `:= Seq(scala212)` for all of those since they're only used at the 'plugin side'

project/Dependencies.scala
```diff
@@ -7,15 +7,16 @@ object Dependencies {
...
+    val scala213 = "2.13.0"
 
-    val akka = "2.5.20"
+    val akka = "2.5.23"
     val akkaHttp = "10.1.8"
 
     val grpc = "1.20.0" // checked synced by GrpcVersionSyncCheckPlugin
     val config = "1.3.3"
-    val sslConfig = "0.3.6"
+    val sslConfig = "0.4.0"
```
marcospereira: Maybe you want to use 0.3.8.

PR #610, opened 2019-06-27
Switch formatting to scalafmt.

tz70s:
* Use the same scalafmt rules as akka.
* Format all files (including to sbt build sources).
* Remove old formatting extension.

## Purpose

Use scalafmt as source code formatting tool.

## References

References #578 

## Changes

The mainly change is the `projects/plugins.sbt` by adding new plugin, `projects/Common.scala` by supplying a `scalafmtOnCompile` option and remove the old `Formatting` object.

Some minor changes for passing headerCheck:

In the prior code, some copyright headers were written in Javadoc style, which would be formatted into near above the package definition, without a newline:
```
/**
 * Copyright xxxxx
 */
package akka.grpc.abc
```
This will make the headerCheck fail as missing header.

Hence, I have replaced all Javadoc style copyright header comment into a universal and consistent form.

## Background Context

Similar approach can be found in [alpakka](https://github.com/akka/alpakka).

BTW, I have not added travis stage for checking format yet. Should it be done in this pr?

PR #611, opened 2019-06-28
Add code style check in travis CI.

tz70s:
Simply add to existing job, due to it's not time consuming in current codebase.

## Purpose

Style checker in Travis CI.

## References

References #610 

## Changes

`scalafmtCheckAll` for checking all sources and `scalafmtSbtCheck` for checking sbt build sources.

## Background Context

https://scalameta.org/scalafmt/docs/installation.html#task-keys

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install: jabba install "$TRAVIS_JDK" && jabba use "$_" && java -Xmx32m -version
 
 jobs:
   include:
-    - script: sbt test headerCheck test:headerCheck grpcVersionSyncCheck
+    - script: sbt test headerCheck test:headerCheck grpcVersionSyncCheck scalafmtCheckAll scalafmtSbtCheck
     - script: sbt scripted
       name: Test scripted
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ install: jabba install "$TRAVIS_JDK" && jabba use "$_" && java -Xmx32m -version
 
 jobs:
   include:
-    - script: sbt test headerCheck test:headerCheck grpcVersionSyncCheck
+    - script: sbt test headerCheck test:headerCheck grpcVersionSyncCheck scalafmtCheckAll scalafmtSbtCheck
```
tz70s: Question:

I guess the `test` is not necessary here due to we already have 2x2 matrix for tests.
WDYT?

raboof: I think the other jobs run integration tests, and this is the only job that runs the unit tests - unless I'm misunderstanding you?

raboof:
OK TO TEST

.travis.yml
tz70s: The default Test stage has 2x2 matrix for running unit tests.
I think the `jobs include` add additional jobs to the default Test stage.
(https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion).

e.g. All 6.1 - 6.4 jobs run unit tests, and the 6.1 & 6.5 run the same test for same Scala version and same env (JDK version).
screenshot

PR #621, opened 2019-07-11
Tweak prepended launch scripts for codegen plugins

mlangc:
This change should make it easier to fix #602

PR #298, opened 2018-07-19
Logging from the generators

johanandren:
Refs #257

Had to do some unethical trix with sbt, but in maven it was no funky business. Didn't check how to deal with it in gradle yet.

PR #622, opened 2019-07-11
Fix Code Generation with Gradle on Windows (see #602)

### LINKED ISSUE #602: generateProto gradle task error: %1 is not a valid Win32 application.
Java version: 11
Gradle version: 5.4.1
Akka-grpc plugin version: 0.6.2

During building I see error:
```
Execution failed for task ':commons:generateProto'.
> protoc: stdout: . stderr: P:\pvp-game\commons\src\main\protobuf: warning: directory does not exist.
  P:\pvp-game\commons\app\protobuf: warning: directory does not exist.
  P:\pvp-game\commons\app\proto: warning: directory does not exist.
  --akkaGrpc_out: protoc-gen-akkaGrpc: %1 is not a valid Win32 application.
```

Full gradle log with debug: [gradlelog.txt](https://github.com/akka/akka-grpc/files/3290043/gradlelog.txt)

mlangc:
The purpose of this PR is to fix #602 which currently affects code generation on Windows. It is based on #621. At the moment it does the following (see e146bf63d6d9fa43c95adf0fb6270d60abc47570):

1. Remove the logfile option, as this requires further tweaking to work
on Windows.
2. Add a hack for Windows that copies JAR assemblies to BAT files,
  skipping the preamble after the shebang, in order to make them
  directly executable.

While these tweaks work for me, I think there is still some work to do, to arrive at a technically viable solution. Your feedback regarding the following points would therefore be highly appreciated:

1.) could probably be fixed properly by investing some more time to make logging work on Windows. Is the log file important?

Regarding 2.) I think that it would be better to create a temporary bash script, that references the assembly JAR, and pass that to `protoc`, instead of creating copies in the local Maven repo that rely on the assembly to be in a particular format. However, since my experience in this area is rather limited, I'm not exactly sure how to do that in
```gradle
plugins {
    akkaGrpc {
        artifact = "com.lightbend.akka.grpc:akka-grpc-codegen_2.12:${pluginVersion}:assembly@${assemblySuffix}"
    }
    if (isScala) {
        scalapb {
            artifact = "com.lightbend.akka.grpc:akka-grpc-scalapb-protoc-plugin_2.12:${pluginVersion}:assembly@${assemblySuffix}"
        }
    }
}
``` 
How can I pass a temporary script instead of an artifact?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ lazy val codegen = Project(id = "akka-grpc-codegen", base = file("codegen"))
     },
     mainClass in assembly := Some("akka.grpc.gen.Main"),
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(
-        prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultShellScript)),
+        prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultUniversalScript(shebang = true))),
     crossScalaVersions -= scala213))
   .settings(addArtifact(artifact in (Compile, assembly), assembly))
 
@@ -40,7 +40,7 @@ lazy val scalapbProtocPlugin = Project(id = "akka-grpc-scalapb-protoc-plugin", b
     },
     mainClass in assembly := Some("akka.grpc.scalapb.Main"),
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(
-        prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultShellScript)),
+        prependShellScript = Some(sbtassembly.AssemblyPlugin.defaultUniversalScript(shebang = true))),
     crossScalaVersions := Seq(scala212)))
   .settings(addArtifact(artifact in (Compile, assembly), assembly))
 
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -1,11 +1,17 @@
 package akka.grpc.gradle
 
+import org.apache.commons.lang.SystemUtils
 import org.gradle.api.Plugin
 import org.gradle.api.Project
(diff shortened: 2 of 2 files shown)


lightbend-cla-validator:

raboof:
OK TO TEST

raboof:
1. we we don't have the log file, can we see the logging from the generators at all? #298 suggests it would be needed for that.

2. while this is a pretty weird hack, I'm not familiar enough with gradle to propose a different solution. I guess it might be fine for now since it is pretty self-contained?

mlangc:
1. I will look into making logging work next week
2. There are two things that I consider to be problematic:
    1. directly creating a file inside the local Maven/Ivy/Gradle repository. While not ideal, this is something we might tolerate until someone comes up with a better solution.
    2. creating an executable bat file by doing a tweaked copy of the executable jar file, that relies heavily on the exact layout of the former (by skipping everything till the first '@' when copying). I think it would be better to either create a separate bat that references the assembly jar, or if that is not possible, a self contained bat, but without relying on the internal layout of the assembly preamble script.

lightbend-cla-validator:

lightbend-cla-validator:

mlangc:
I've now addressed the issues mentioned [above](https://github.com/akka/akka-grpc/pull/622#issuecomment-510866296):
1. Logging now also works on Windows
2. I now create a separate bat file that references the assembly jar

If this approach is OK for you, I would do a final cleanup and we could prepare this PR for merging. Do you want me to squash everything into one commit?

raboof:
> While not ideal, this is something we might tolerate until someone comes up with a better solution.

Agreed

> If this approach is OK for you, I would do a final cleanup and we could prepare this PR for merging. 

Yes looks reasonable!

> Do you want me to squash everything into one commit?

Normally we could do that on squash, but I see some of your later commits aren't recognised as yours by github, causing our CLA checker to get confused. Might be good to squash them into one easily-attributable commit?

mlangc:
I have now squashed everything into one commit, and removed an usused method.

Issue #617, opened 2019-07-05
Gradle plugin not publishing generated Java sources in a location an IntelliJ project knows of


orendain:
### Version

- Using akka-grpc-gradle-plugin, version **0.6.2**
- Using Java 1.8

### Expected Behavior

`gradlew build` should publish classes in a location that IntelliJ picks up on.  

### Actual Behavior

A vanilla Gradle project, as described by the docs (including the akka-grpc-quickstart-java project), publishes classes (via `gradlew build`) in a location that IntelliJ isn't made aware of.

Specifically, IntelliJ cannot resolve:
- Any of the generated proto classes
- `ProtobufSerializer` in `import akka.grpc.ProtobufSerializer`
- `GoogleProtobufSerializer` in `import akka.grpc.javadsl.GoogleProtobufSerializer`

Gradle correctly builds and runs, but IntelliJ isn't able to recognize certain sets of classes.

orendain:
Currently, my sad workaround is generating the different proto/grpc classes (via regular `gradle build`), and copying them (or symlink) from the `build` directory out to `src` where it can be picked up by Intellij.

If anyone is having the same problem and modified the plugin to work, or has a cleaner IntelliJ config adjustment for having IntelliJ play nice with the plugin, a share would be appreciated 😃 .

raboof:
It would be great if someone can figure this out! I'm not much of a gradle expert either, I'm afraid...

johanandren:
Curious thing though: I just created a gradle project from scratch and loaded that into IntelliJ and it does pick up `build/generated/source/proto/main/*` but makes it a sources root, not generated sources like one would want.

Looks like it expects the directory to be `build/generated/sources` rather than `source` because when building that appears as well, perhaps that is the problem.

johanandren:
Looking at the gradle plugin it does something weird, it only adds the generated directories to the `sourceSets` if the project is a Scala project, that is perhaps a clue in a different direction.

orendain:
@johanandren Indeed - I found the same thing: generated directories are added only for scala projects and not for java ones.

orendain:
Updated title to better reflect the exact issue we're discussing.  I made a change to the gradle-plugin, rebuilt it, and have things working as expected.  Will open up a PR for this.

PR #624, opened 2019-07-12
Gradle plugin defines correct sourcesets for java

orendain:
## Purpose

In a Gradle project, after generation of the protobuf files, these files should be available as generated source for the rest of the project.  This is the case with Scala projects, but not with Java projects.

This PR directs Gradle to add the correct generated source directories for Java projects.

## References

References #617 

## Changes

- Source sets are now added for both scala and java

## Background Context

N/A - but #617 has a short discussion about how I came across this issue.

Edit: CLA signed.

### DIFF
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -25,6 +25,7 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
 
         project.configure(project) {
             boolean isScala = "${extension.language}".toLowerCase() == "scala"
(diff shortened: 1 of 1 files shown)


johanandren:
OK TO TEST

johanandren:
What about `sources` vs `source` that I mentioned in the issue, does this already make IntelliJ pick it up correctly?

orendain:
@johanandren In the builds I'm running, `build/generated/source` is where all generated classes are saved to.  The `build/generated/sources` path is , indeed, also generated, but it houses only an empty `annotationProcessor/java/main` directory - no sources appear to be generated there.

I'm not sure if something is ever saved to that directory, and I just haven't come across it yet, or if it's a dangling resource brought on by some package.

johanandren:
Ah, ok, I thought maybe that was the directory gradle plugins in general put stuff in.

PR #639, opened 2019-07-18
Don't automatically update to ssl-config 0.4.0

raboof:


raboof:
(manually verified travis is ok)

Issue #623, opened 2019-07-12
flat_package is not enabled by default for maven


raboof:
When generating Scala code via Maven, `flat_package` is not enabled by default.

PR #645, opened 2019-07-26
Enable flat_package as default in maven plugin.

tz70s:
* Fix the wrong initialization of generatorSettings.
* Remove flat_package option in plugin testing for validation.

## Purpose

Enable `flat_package` option in maven plugin.

## References

References #623 

## Changes

The `generatorSettings` bean is injected by Maven plugin dependency injection. Hence, the prior initialization was not correct. 
Add `flat_package` option at the execution call, and remove the config in `plugin-tester-scala` for plugin test.

### DIFF
--- a/docs/src/main/paradox/buildtools/maven.md
+++ b/docs/src/main/paradox/buildtools/maven.md
@@ -38,7 +38,7 @@ Scala
 ### Generating server "power APIs"
 
 To additionally generate server "power APIs" that have access to request metata, as described
-@ref[here](../server/walkthrough.md#accessing-request-metadata), set the `server_power_apis` option:
+@ref[here](../server/walkthrough.md#accessing-request-metadata), set the `serverPowerApis` tag as true:
 
 Java
 :   ```xml
@@ -47,7 +47,7 @@ Java
         <configuration>
           ...
           <generatorSettings>
-            <generatorSetting>server_power_apis</generatorSetting>
+            <serverPowerApis>true</serverPowerApis>
           </generatorSettings>
         </configuration>
     </plugin>
@@ -61,8 +61,7 @@ Scala
           ...
           <language>Scala</language>
           <generatorSettings>
-            <generatorSetting>flat_package</generatorSetting>
-            <generatorSetting>server_power_apis</generatorSetting>
+            <serverPowerApis>true</serverPowerApis>
           </generatorSettings>
         </configuration>
     </plugin>
--- a/maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
+++ b/maven-plugin/src/main/scala/akka/grpc/maven/GenerateMojo.scala
@@ -96,13 +96,10 @@ class GenerateMojo @Inject()(project: MavenProject, buildContext: BuildContext) 
 
   import scala.collection.JavaConverters._
   @BeanProperty
-  var generatorSettings: java.util.ArrayList[String] = {
-    val init = new java.util.ArrayList[String]()
-    init.add("flat_package")
-    init
-  }
+  var generatorSettings: java.util.ArrayList[String] = _
+
   @BeanProperty
-  var extraGenerators: java.util.ArrayList[String] = new java.util.ArrayList
+  var extraGenerators: java.util.ArrayList[String] = _
 
   override def execute(): Unit = {
     val chosenLanguage = parseLanguage(language)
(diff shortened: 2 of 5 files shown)


raboof:
OK TO TEST

raboof:
Thanks for looking into fixing this bug!

AFAICS, this now makes it impossible to disable `flat_package`. Do you know of any Maven convention to make that possible? Should we introduce an `excludedGeneratorSettings` or something like that?

tz70s:
> AFAICS, this now makes it impossible to disable `flat_package`. 

Make sense! I was confused about it because I think the current Gradle based generator can't tweak `flat_package` option as well. Maybe we need to fix it too.

https://github.com/akka/akka-grpc/blob/62e23269e5ff7a4e28eff7774ecefdf0c7857025/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L156

> Do you know of any Maven convention to make that possible? Should we introduce an `excludedGeneratorSettings` or something like that?
I'm not a Maven expert as well. 

Adding additional exclusion list is an option, but what about making `generatorSettings` into a `Map<String, Boolean>` based configuration, similar to Gradle approach?

Ref: https://maven.apache.org/guides/plugin/guide-java-plugin-development.html

e.g. 
```xml
<generatorSettings>
  <flatPackage>false</flatPackage>
  <serverPowerAPIs>true</serverPowerAPIs>
</generatorSettings>
```

raboof:
> what about making `generatorSettings` into a `Map<String, Boolean>` based configuration

Yes, that sounds great: it is more consistent with the other plugins, and keeps the door open for other non-boolean options in the future.

tz70s:
I updated the new implementation as discussed above.

After digging around, I got some more thoughts & questions here:

### Plugin Consistency

Let's iterate the current plugin options (configurations):
* Top level: `language`, `generateClient`, `generateServer`
* GeneratorSettings: `flat_package`, `server_power_apis`, `use_play_actions`, `generate_play`, etc. 
* Extra generators

The above hierarchy is consistent in SBT and Maven, but **not** Gradle.

Should we introduce an additional `generatorSettings` property in Gradle plugin?

### Options Consistency

Specific to `GeneratorSettings`. The data types vary in three different plugins.

* SBT: Use plain text in a `Seq`, e.g. `Seq("flat_package", "server_power_apis")`.
* Maven: Use a Map<String, String> data structure, and manually filter out the false values. (The new implementation)
* Gradle: continue to the above issue, the Gradle plugin has no such data structure.

To avoid complexity, I think we can make it more consistent, such as making all of them into a `String -> String` pair, and then perform different checking/error reporting mechanism in each plugin frontend.

For example, we can first type check of each options (AFAIK, there's only booleans) in SBT (compile time), Gradle/Maven (runtime), and then pass into a `Map[String, String].

Using both Seq("flat_package=true") and Seq("flat_package") in different plugins is quite confusing.

### Code Generation

The best way to have uniform options of all plugins are introducing/sharing Scala/Java codebase for defining them. Same as we did in SBT plugin: https://github.com/akka/akka-grpc/blob/master/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L39-L47

IIUC, the current workflow is launching a `CodeGenerator` executable JAR, and pass parameters from each plugin implementation.

Can we make the `CodeGenerator` as exposed APIs (introduce to classpath) and being called in each plugins accordingly?

* Pros: simpler, uniform and scalable code, may be faster too.
* Cons: heavier plugin size.

Or if I understood incorrectly?

tz70s:
> Can we make the `CodeGenerator` as exposed APIs (introduce to classpath) and being called in each plugins accordingly?

Oops, seems like it was already introduced. Let me rephrase the code generation section: can we move standard options (e.g. `flat_package`) into `codegen` module?

raboof:
> To avoid complexity, I think we can make it more consistent, such as making all of them into a String -> String pair, and then perform different checking/error reporting mechanism in each plugin frontend.

It might indeed make sense to convert to `String -> String` pairs 'internally' in each plugin, and share some more of the logic. However, since for example sbt has good support for removing options from lists, I don't think it is needed (and it would be inelegant) to expose it like that to the user.

I think being consistent with the conventions common in each build system is more important than having the 3 plugins be consistent with each other. Does that make sense?

> Using both Seq("flat_package=true") and Seq("flat_package") in different plugins is quite confusing.

Supporting both might not be so bad though?

> can we move standard options (e.g. flat_package) into codegen module?

Yeah, I don't see why not!

tz70s:
> > To avoid complexity, I think we can make it more consistent, such as making all of them into a String -> String pair, and then perform different checking/error reporting mechanism in each plugin frontend.
> 
> It might indeed make sense to convert to `String -> String` pairs 'internally' in each plugin, and share some more of the logic. However, since for example sbt has good support for removing options from lists, I don't think it is needed (and it would be inelegant) to expose it like that to the user.

> I think being consistent with the conventions common in each build system is more important than having the 3 plugins be consistent with each other. Does that make sense?

Definitely we should follow build systems' conventions, usually users do not build apps across multiple build systems. Just in case if these options will grow larger, then we might need a "standard" structure to support all kinds of options (which might be different types to `boolean`). For example, we can make it as ADT-based options in sbt.
```scala
akkaGrpcCodeGeneratorSettings += AkkaGrpc.GeneratorSettings.ServerPowerApis
akkaGrpcCodeGeneratorSettings := akkaGrpcCodeGeneratorSettings.value.filterNot(_ == AkkaGrpc.GeneratorSettings.FlatPackage)
```
For those ADTs, they can be embedded with different content(types).
```scala
akkaGrpcCodeGeneratorSettings += AkkaGrpc.GeneratorSettings.StringBasedOption("hi")
```
Something like this. Anyway, currently it's fine with me cause there's only few `boolean` options. 

> > Using both Seq("flat_package=true") and Seq("flat_package") in different plugins is quite confusing.
> 
> Supporting both might not be so bad though?

Yeah.. it is more flexible.

> > can we move standard options (e.g. flat_package) into codegen module?
> 
> Yeah, I don't see why not!

Should I do this in this PR or separately? It needs a larger refactoring across plugins though.

Let me know if this implementation for Maven plugin is good or not.

Thanks for reply!

raboof:
> Should I do this in this PR or separately? It needs a larger refactoring across plugins though.

If it needs more refactoring then let's keep that for a separate PR.

> Thanks for reply!

Thanks for the contribution! I'll have a closer look later but what I've seen looks good so far!

PR #646, opened 2019-07-26
Fix junit deprecation warning.

tz70s:
The junit related packages were moved under org.scalatestplus.

## Background Context

The changed occurred since scalatest 3.0.6 version, see [changelog](https://github.com/scalatest/scalatest/releases)


raboof:
OK TO TEST

Issue #641, opened 2019-07-18
Generate import statements


raboof:
Currently, the imports used in generated classes is static. This may lead to 'unsused imports' when some features are not used in that class. It would be neater to generate exactly the imports that are needed.

raboof:
An additional feature might be to detect name clashes and refer to FQCN's instead of imported classes only in that case. Makes you wish for a more 'semantic' code generation library than twirl, though.

PR #649, opened 2019-07-30
On demand import statements generation.

tz70s:
* Dynamic generate the import statements via twirl dynamic codegen.
* Remove several unused imports and deprecation warnings (mostly in plugin-tester-*) for reproducing changes.

## Purpose

This PR make the import statements generated by `codegen` module dynamic.

## References

References #641 

## Changes

This is basically a workaround of generating import statements.
From now, the generated scala code would not produce unused warnings.

However, there's still some questions opened:

1. No on demand Java import statements generation: I didn't find a way to produce (automatic) static analysis on **generated Java code**. If such tool doesn't exist, we can omit it for easier maintenance.

2. Should we test against this? For scala, we can add `-Ywarn-unused-import` and `-Xfatal-warnings` to `scalacOptions` for maybe `plugin-tester-scala`, but this would definitely break CI easily. For java, same as question 1.

3. Should we move the dynamic code generation codebase into scala codebase? Currently, there's only two parts for conditionally generating import statements: use `powerApis` and various method signatures (`Unary`, `ClientStreaming`, etc), and they don't follow a general rule. That means, we might need a large refactoring for each `CodeGenerator`, `Service`, etc. to generate those rules polymorphically. 

Any tips on these? Thanks!

### DIFF
--- a/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
@@ -7,7 +7,7 @@
 @akka.grpc.gen.Constants.DoNotEditComment
 package @service.packageName
 
-import scala.concurrent.{ ExecutionContext, Promise }
+import scala.concurrent.ExecutionContext
 
 import io.grpc.ManagedChannel
 import io.grpc.MethodDescriptor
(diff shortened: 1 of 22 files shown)


raboof:
OK TO TEST

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -7,27 +7,37 @@
...
 import akka.grpc.internal.Marshaller
 import akka.grpc.internal.NettyClientUtils
 import akka.grpc.internal.ClientState
 
 import akka.stream.Materializer
 
+@{
+def withSingleResponse(stmt: String) = Set("import akka.grpc.scaladsl.SingleResponseRequestBuilder", stmt)
+def withStreamResponse(stmt: String) = Set("import akka.grpc.scaladsl.StreamResponseRequestBuilder", stmt)
+service.methods.map(_.methodType).foldLeft(Set.empty[String]) { (imports, method) =>
```
raboof: Rather than a full `fold` I think a `flatMap` might be sufficient here and a bit simpler?

tz70s: Sure!

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -7,27 +7,37 @@
...
 import akka.grpc.internal.Marshaller
 import akka.grpc.internal.NettyClientUtils
 import akka.grpc.internal.ClientState
 
 import akka.stream.Materializer
 
+@{
+  def withSingleResponse(stmt: String) = Set("import akka.grpc.scaladsl.SingleResponseRequestBuilder", stmt)
+  def withStreamResponse(stmt: String) = Set("import akka.grpc.scaladsl.StreamResponseRequestBuilder", stmt)
+  service.methods.toSet.flatMap { method: akka.grpc.gen.scaladsl.Method =>
```
tz70s: I have no idea why the type can't be inferred here actually.

```
[error] /Users/tz70s/WorkSpace/scala/akka-grpc/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt:28:35: missing parameter type
[error]   service.methods.toSet.flatMap { method =>
[error]                                   ^
[error] one error found
[error] (akka-grpc-codegen / Compile / compileIncremental) Compilation failed
```

jrudolph: Probably because Set is contravariant. You could try

```suggestion
  service.methods.distinct.flatMap { method =>
```

tz70s: Thanks for suggestion! 

It's because `List` is covariant and `Set` is invariant (ref: https://stackoverflow.com/questions/26237939/toset-and-type-inference)

I think I need a `Set` here for flatMapping to avoid duplicate import statements. 
Another options are flatMapping `List` and toSet at the final step.
```scala
service.methods.flatMap { /* generate statements */ }.toSet.mkString("\n")
```
and create a temporary value for type inference.
```scala
val methodsSet = service.methods.toSet
methodsSet.flatMap { /* generate statements */ }.mkString("\n")
```

Anyway, I guess the explicit type signature is pretty safe here?

tz70s:
I just found a Intellij trick to inspect unused imports at generated java code, let me revise the Java imports generation as well.

PR #650, opened 2019-07-30
Don't preprend . to service name if package not set

### LINKED ISSUE #450: support .proto files without package declarations
I don't even know if this is possible, but I follow the quick-start-scala and used the same proto file to generate a Node.js client and it didn't work. Getting the error:  Received http2 header with status: 404

I was able to do this using pure ScalaPB and Node.js.

Any thoughts?

chbatey:
Fixes #450

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/Service.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/Service.scala
@@ -38,7 +38,7 @@ object Service {
     Service(
       fileDesc.scalaPackageName,
       serviceClassName,
-      fileDesc.getPackage + "." + serviceDescriptor.getName,
+      (if (fileDesc.getPackage.isEmpty) "" else fileDesc.getPackage + ".") + serviceDescriptor.getName,
       serviceDescriptor.getMethods.asScala.map(method => Method(method)).to[immutable.Seq],
       serverPowerApi,
       usePlayActions,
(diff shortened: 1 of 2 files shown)


chbatey:
Is there an easy way to add this to the interop tests? I didn't see the protos in there

codegen/src/main/scala/akka/grpc/gen/scaladsl/Service.scala
```diff
@@ -38,7 +38,7 @@ object Service {
     Service(
       fileDesc.scalaPackageName,
       serviceClassName,
-      fileDesc.getPackage + "." + serviceDescriptor.getName,
+      (if (fileDesc.getPackage.isEmpty) "" else fileDesc.getPackage + ".") + serviceDescriptor.getName,
```
helena: Maybe we should consider adding a new `FileDescriptorOps` for the logic in one versus two, duplicated.

helena: To provide one convention.

raboof:
> Is there an easy way to add this to the interop tests? I didn't see the protos in there

The interop tests mainly test against https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md and get the `.proto` from the upstream jar.

We could:
* widen the `interop tests` to also include other tests that require code generation
* introduce a new subproject for other tests that require code generation
* only test features that require code generation in 'scripted' tests in the sbt-plugin subproject

'1' sounds fine to me. Getting the code generated might be as simple as adding a proto in that subproject under `src/main/proto`.

chbatey:
> '1' sounds fine to me. Getting the code generated might be as simple as adding a proto in that subproject under src/main/proto.

Sounds good will do that

Issue #451, opened 2018-11-02
Code generation customizations via protobuf files


raboof:
An open question is how to pass in fine-grained options to configure how code should be generated.

So far we have been passing these in via the 'parameters' field of the protoc request. That works, but is awkward to scale.

One interesting option is to pass in these options via proto files. A requirement is that it should be possible to change the codegen configuration options for a protocol without modifying the proto file of the protocol itself, since the codegen is local and the proto might come from an external source.

scalaPB has recently introduced [package-scoped options](https://scalapb.github.io/customizations.html#package-scoped-options) which looks like it might work.

adarshaj:
If and when this is implemented, I'd like it to support the auxiliary options too (similar to [how scalapb does](https://scalapb.github.io/customizations.html#auxiliary-options)). That way, we could have a single proto file with all akka-grpc specific settings and keep original proto file free of cruft, this even makes it easier to skip that auxiliary options file altogether with a different runtime/protoc-plugin where akka-grpc stuff may not be needed!

raboof:
Agreed entirely :+1:

PR #651, opened 2019-08-02
update akkaGrpcCodeGeneratorSettings possible values

bjaglin:
* remove `grpc`, confusing within Akka gRPC as it's ScalaPB's
* align `single_line_to_proto_string` with ScalaPB, keeping the previous flag name for backward compatibility
* support ScalaPB's latest flags: `ascii_format_to_string`, `no_lenses` & `retain_source_code_info`
* reflect the possible values in the settings description

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
@@ -41,6 +41,7 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
     // Currently per-invocation options, intended to become per-service options eventually
     // https://github.com/akka/akka-grpc/issues/451
     val params = request.getParameter.toLowerCase
+    // flags listed in akkaGrpcCodeGeneratorSettings's description
     val serverPowerApi = params.contains("server_power_apis") && !params.contains("server_power_apis=false")
     val usePlayActions = params.contains("use_play_actions") && !params.contains("use_play_actions=false")
 
@@ -65,12 +66,16 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
     b.build()
   }
 
+  // flags listed in akkaGrpcCodeGeneratorSettings's description
   private def parseParameters(params: String): GeneratorParams =
     params.split(",").map(_.trim).filter(_.nonEmpty).foldLeft[GeneratorParams](GeneratorParams()) {
-      case (p, "java_conversions")      => p.copy(javaConversions = true)
-      case (p, "flat_package")          => p.copy(flatPackage = true)
-      case (p, "grpc")                  => p.copy(grpc = true)
-      case (p, "single_line_to_string") => p.copy(singleLineToProtoString = true)
-      case (x, _)                       => x
+      case (p, "java_conversions")            => p.copy(javaConversions = true)
+      case (p, "flat_package")                => p.copy(flatPackage = true)
+      case (p, "single_line_to_string")       => p.copy(singleLineToProtoString = true) // for backward-compatibility
+      case (p, "single_line_to_proto_string") => p.copy(singleLineToProtoString = true)
+      case (p, "ascii_format_to_string")      => p.copy(asciiFormatToString = true)
+      case (p, "no_lenses")                   => p.copy(lenses = false)
+      case (p, "retain_source_code_info")     => p.copy(retainSourceCodeInfo = true)
+      case (x, _)                             => x
     }
 }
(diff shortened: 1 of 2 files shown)


raboof:
OK TO TEST

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -70,7 +70,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
     val akkaGrpcGenerators = settingKey[Seq[protocbridge.Generator]](
       "Generators to evaluate. Populated based on akkaGrpcGeneratedLanguages, akkaGrpcGeneratedSources and akkaGrpcExtraGenerators, but can be extended if needed")
     val akkaGrpcCodeGeneratorSettings = settingKey[Seq[String]](
-      "Boolean settings to pass to the code generators, empty (all false) by default. ScalaPB settings: java_conversions, flat_package, single_line_to_proto_string, ascii_format_to_string. Akka gRPC settings: server_power_apis")
+      "Boolean settings to pass to the code generators, empty (all false) by default. ScalaPB settings: java_conversions, flat_package, single_line_to_proto_string, ascii_format_to_string, no_lenses, retain_source_code_info. Akka gRPC settings: server_power_apis, use_play_actions.")
```
raboof: (Hmm, I *think* `use_play_actions` is no longer needed, but that's unrelated to this PR)

codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
```diff
@@ -65,12 +66,16 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
...
-      case (p, "grpc")                  => p.copy(grpc = true)
-      case (p, "single_line_to_string") => p.copy(singleLineToProtoString = true)
-      case (x, _)                       => x
+      case (p, "java_conversions")            => p.copy(javaConversions = true)
+      case (p, "flat_package")                => p.copy(flatPackage = true)
+      case (p, "single_line_to_string")       => p.copy(singleLineToProtoString = true) // for backward-compatibility
+      case (p, "single_line_to_proto_string") => p.copy(singleLineToProtoString = true)
+      case (p, "ascii_format_to_string")      => p.copy(asciiFormatToString = true)
+      case (p, "no_lenses")                   => p.copy(lenses = false)
+      case (p, "retain_source_code_info")     => p.copy(retainSourceCodeInfo = true)
```
tz70s: I'm wondering this parsing can handle the key-value form passing via different plugins? (currently only `gradle`)
e.g. `single_line_to_string=true`. I guess current test infra can't catch this, we only have `serverPowerApis` been used. https://github.com/akka/akka-grpc/blob/f6536403a5cccecb6398ec0485e0ba45f9fa994f/plugin-tester-scala/build.gradle#L34

bjaglin: The `key=value` format is only supported for/via: https://github.com/akka/akka-grpc/blob/717731d479b8ad9e9103a6a2e5ac49b813b0ab93/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala#L45.

I have been thinking about the flag parsing when opening this PR, as the current solution clearly does not scale:
1. the akka-grpc specific enum values are declared in the plugin, but can't be accessed from the generators (otherwise we would have a cyclic dependency between the `codegen` and the sbt plugin modules) 
1. unhandled enum values are silently ignored.

I did not have the gradle use case in mind (I had started to push more stuff in the sbt plugin). Maybe we can follow-up on a ticket to brainstorm? My intent here was just a quick one-shot to expose new ScalaPB flags.

tz70s: > The `key=value` format is only supported for/via:
> 
> https://github.com/akka/akka-grpc/blob/717731d479b8ad9e9103a6a2e5ac49b813b0ab93/codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala#L45
> 
> .
> I have been thinking about the flag parsing when opening this PR, as the current solution clearly does not scale, but identified a couple of challenges:
> 
> 1. the Akka-gRPC-specific enum values are declared in the plugin, but can't be accessed from the generators (otherwise we would have a cyclic dependency between the `codegen` and the sbt plugin modules)
> 2. The ScalaPB-specific enums can't seem to be listed/retrieved from upstream
> 
> As a result, unhandled enum values are silently ignored.

> I did not have the gradle use case in mind (I had started to push more stuff in the sbt plugin). Maybe we can follow-up on a ticket to brainstorm? 

Yes, we have discussed this at (https://github.com/akka/akka-grpc/pull/645#issuecomment-516171609). I'm going to do a follow-up refactoring for standard generator settings enums in `codegen` module. That would be nice if we have a ticket to track these. The supposed changes including to:

1. Move **generator options** and **parsing** into a single place (in `codegen` module) to be shared across sbt and maven plugins, as well as scaladsl and javadsl code generator.
2. A better way I'm thinking is taking this https://github.com/akka/akka-grpc/issues/451 for a more scalable and unified way to tackle these options, and we can find some way to make the current configuration system compatible.

> My intent here was just a quick one-shot to expose new ScalaPB flags.

Yes, that make sense!

PR #654, opened 2019-08-06
Server Reflection

raboof:
## Purpose

Add an initial implementation class for a 'Server Reflection' API, based on the existing cloudstate implementation.

You can run the `03-server-reflection` example to see the API in action.

Remaining work:
* [x] Verify `.proto` resources parsed into `FileDescriptor`-like structures are available when generating code with protoc-java as well
* [x] Add user-facing API to serve these routes for a Java project
* [x] Add user-facing API to serve these routes for a Scala project
* [ ] Adding this service adds quite some extra classes to the classpath. should this be a separate module?

## References

References #380


runtime/src/main/protobuf/google/api/annotations.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2015, Google Inc.
```
viktorklang: This file is not needed for server reflection

runtime/src/main/protobuf/google/api/http.proto
```diff
@@ -0,0 +1,376 @@
+// Copyright 2019 Google LLC.
```
viktorklang: This file is not needed for server reflection

runtime/src/main/protobuf/google/api/annotations.proto
raboof: It is needed to generate `AnnotationsProto` which is used at https://github.com/akka/akka-grpc/pull/654/files#diff-55da3d8b2cdb4c2612a6d74c31c31ee5R55 . It might be possible to load it from a dependency instead of including the .proto here, I suppose?

runtime/src/main/protobuf/google/api/http.proto
raboof: it is needed to be able to process annotations.proto (https://github.com/akka/akka-grpc/pull/654/files#diff-bd58959d0ef01f941f2e2a00d2a788baR30)

viktorklang: Yeah, but I don't think you need annotations.proto, I only included it in Cloudstate because it is a builtin feature of Cloudstate.

raboof: Aaah I didn't realize, will remove it then!

raboof:
(I don't currently have the bandwidth to finish this PR, feel free to continue where I left off here!)

viktorklang:
@raboof I suggest using this: https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/Reflection.scala

sbt-plugin/src/sbt-test/gen-java/02-server-reflection/src/main/java/example/myapp/helloworld/Main.java
```diff
@@ -0,0 +1,54 @@
...
+        //#server-reflection
+        // Instantiate implementation
+        GreeterService impl = new GreeterServiceImpl();
+
+        // Bind service handler servers to localhost:8080
+        return Http.get(sys).bindAndHandleAsync(
+            ServiceHandler.concatOrNotFound(
+                GreeterServiceHandlerFactory.create(impl, mat, sys),
+                new ServerReflection(java.util.Arrays.asList(GreeterService.service)).create(mat, sys)
+            ),
```
johanandren: Good showing the server reflection second as that will likely be hit less often than the actual service

runtime/src/main/protobuf/reflection.proto
```diff
@@ -0,0 +1,136 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Service exported by server reflection
+
+syntax = "proto3";
+
+package grpc.reflection.v1alpha;
```
johanandren: provide a link to https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1alpha/reflection.proto somewhere int the comment here saying that is the original or something?

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
```diff
@@ -0,0 +1,108 @@
...
+package akka.grpc.internal
+
+import com.google.protobuf.Descriptors.FileDescriptor
+
+import akka.NotUsed
+import akka.stream.scaladsl._
+
+import _root_.grpc.reflection.v1alpha.reflection._
+
+class ServerReflectionImpl private (fileDescriptors: Map[String, FileDescriptor], services: List[String])
```
johanandren: `@InternalApi`and `private[akka]` etc. perhaps `final` as well?

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
```diff
@@ -0,0 +1,108 @@
...
+      ServerReflection.name +: services)
+
+  val AdditionalDescriptors =
+    flattenDependencies(ReflectionProto.javaDescriptor).distinct
+
+  private def flattenDependencies(descriptor: FileDescriptor): List[FileDescriptor] = {
+    descriptor :: descriptor.getDependencies.asScala.toList.flatMap(flattenDependencies)
+  }
+
+  def splitNext(name: String): (String, String) = {
```
johanandren: internal API most of these methods (and fields) or the object itself (apply is called from user code?)

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
raboof: yeah - that's mostly implicit from being in the `akka.grpc.internal` package, but good to make more obvious

sbt-plugin/src/sbt-test/gen-java/02-server-reflection/src/main/java/example/myapp/helloworld/Main.java
raboof: yes - and where it's hit it's probably not on a performance-critical path

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
raboof: user code shouldn't need to refer to the `Impl` at all, the `ServerReflection` in javadsl/scaladsl should be enough

raboof:
(the failure in akka-benchmarks is the one mentioned in #765 - not sure why that's popping up here now but I guess we should just fix it anyway...

build.sbt
```diff
@@ -138,6 +141,9 @@ lazy val docs = Project(id = "akka-grpc-docs", base = file("docs"))
     name := "Akka gRPC",
     publish / skip := true,
     whitesourceIgnore := true,
+    // We don't yet publish java/scaladoc, so this is not yet relevant
+    // https://github.com/akka/akka-grpc/issues/784
+    // apidocRootPackage := "akka.grpc",
```
johanandren: Adding something that is not yet relevant 👍 :D

codegen/src/main/twirl/templates/JavaCommon/ApiInterface.scala.txt
```diff
@@ -20,6 +20,7 @@ public interface @{service.name} {
   }
 
   static String name = "@service.grpcName";
+  static akka.grpc.ServiceObject service = new akka.grpc.ServiceObjectImpl(name, @service.descriptor);
```
johanandren: Are these final automatically because they are in an interface? (if not, they probably should be)

runtime/src/main/scala/akka/grpc/ServiceObject.scala
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+package akka.grpc
+
+import akka.annotation.ApiMayChange
+
+import com.google.protobuf.Descriptors.FileDescriptor;
+
+@ApiMayChange
+trait ServiceObject {
```
johanandren: Not a great name. Something `WithDescriptor` `SelfDescribing` `ServiceMetadata`?

project/plugins.sbt
```diff
@@ -1,11 +1,17 @@
+enablePlugins(BuildInfoPlugin)
+
+val sbtProtocV = "0.99.26"
+
+buildInfoKeys := Seq[BuildInfoKey]("sbtProtocVersion" -> sbtProtocV)
+
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.0")
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.4.0")
-addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.19")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % sbtProtocV)
```
johanandren: Weird to have one version that is only used once in a val but not the others?

project/plugins.sbt
```diff
@@ -1,11 +1,17 @@
...
 addSbtPlugin("com.dwijnand" % "sbt-dynver" % "4.0.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % "1.5.0")
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.10")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
-addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.9.2")
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
```
johanandren: Is dependency graph used in the build or just here for convenience looking at dependencies (if so does it belong in the build vs a local build config, maybe skip it instead and keep the build lean)

docs/src/main/paradox/server/reflection.md
```diff
@@ -0,0 +1,53 @@
+# Server Reflection
+
+@@@note
+
+This feature is **experimental**. The API may still change in further versions
+of Akka gRPC, and future versions of this feature may not work with services
+generated with older versions of Akka gRPC.
```
johanandren: The code API or the gRPC reflection API or both may change?

project/ReflectiveCodeGen.scala
```diff
@@ -20,6 +20,9 @@ object ReflectiveCodeGen extends AutoPlugin {
   val extraGenerators = SettingKey[Seq[String]]("reflectiveGrpcExtraGenerators")
   val codeGeneratorSettings = settingKey[Seq[String]]("Code generator settings")
 
+  // needed to be able to override the PB.generate task reliably
+  override def requires = ProtocPlugin
```
johanandren: Heh, curious it worked consistently without it.

runtime/src/main/scala/akka/grpc/ServiceObjectImpl.scala
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc
+
+import com.google.protobuf.Descriptors.FileDescriptor;
+
+import akka.annotation.ApiMayChange
+
+@ApiMayChange
```
johanandren: Should be in `akka.grpc.internal` and `@InternalApi`, no?

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
johanandren: ok, so private[akka] the and add `@InternalApi` annotation to the object then rather

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def findExtensionNumbersForContainingType(
+      container: String,
+      fileDescriptors: Map[String, FileDescriptor]): List[Int] =
+    (for {
+      fileDesc <- fileDescriptors.values.iterator
+      extension <- fileDesc.getExtensions.iterator.asScala
+      if extension.getFullName == container
+    } yield extension.getNumber).toList
+}
```
johanandren: Some unit tests for these contains/find methods would be nice, both for verifying that they do what they should and to document how they are expected to behave.

runtime/src/main/scala/akka/grpc/javadsl/ServerReflection.scala
```diff
@@ -0,0 +1,34 @@
...
+import akka.annotation.ApiMayChange
+import akka.grpc.ServiceObject
+import akka.grpc.internal.ServerReflectionImpl
+import akka.http.javadsl.model.{ HttpRequest, HttpResponse }
+import akka.stream.Materializer
+
+import grpc.reflection.v1alpha.reflection.ServerReflectionHandler
+
+@ApiMayChange
+class ServerReflection(objects: Collection[ServiceObject]) {
```
johanandren: Since this doesn't really have a lifecycle and just has one method operating on the set of objects, does it really make sense to make it a class vs just a method `create(objects,  system, materializer)`?

runtime/src/main/scala/akka/grpc/scaladsl/ServerReflection.scala
```diff
@@ -0,0 +1,28 @@
...
+  def apply(objects: List[ServiceObject])(
+      implicit mat: Materializer,
+      sys: ActorSystem): HttpRequest => scala.concurrent.Future[HttpResponse] =
+    ServerReflectionHandler.apply(ServerReflectionImpl(objects.map(_.descriptor), objects.map(_.name)))
+  @ApiMayChange
+  def partial(objects: List[ServiceObject])(
+      implicit mat: Materializer,
+      sys: ActorSystem): PartialFunction[HttpRequest, scala.concurrent.Future[HttpResponse]] =
+    ServerReflectionHandler.partial(ServerReflectionImpl(objects.map(_.descriptor), objects.map(_.name)))
+}
```
johanandren: Ah here it is just methods. I like this better.

project/plugins.sbt
raboof: It's used in the `buildInfoKeys` as well, which makes it available to the main project build (#765)

codegen/src/main/twirl/templates/JavaCommon/ApiInterface.scala.txt
raboof: checked with javap, they're indeed final

project/plugins.sbt
johanandren: Ah, didn't connect those dots

Issue #670, opened 2019-08-19
grpcClient service discovery is not retried if the first lookup call returns no target


fredfp:
The service lookup is never retried if it returns an empty list of target.

In my case this happens using `akka-dns` discovery mechanism to discover kubernetes services.

One should not expect the kubernetes resources to be deployed in any particular order. 

Consequently, service lookup should be retried if it returns an empty list of targets.

fredfp:
@johanandren is 0.7.2 due soon? I'd love to have this released...

johanandren:
Not planned, we have been very busy with the Akka 2.6 work, I'll see if we can get 0.7.2 out soonish.

johanandren:
0.7.2 is available on maven central now.

fredfp:
Fantastic, thank you @johanandren !

PR #671, opened 2019-08-19
Retry service discovery if no target is found

fredfp:
proposed fix for #670

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
@@ -25,7 +25,7 @@ import scala.concurrent.{ ExecutionContext, Future, Promise }
  * Can be caught to re-try lookup if it is likely that
  * your service discovery mechanism will resolve to different instances.
  */
-class NoTargetException(msg: String) extends RuntimeException(msg)
+final class NoTargetException(msg: String) extends RuntimeException(msg)
 
 /**
  * INTERNAL API
--- a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
@@ -65,8 +65,9 @@ object NettyClientUtils {
           ChannelUtils.monitorChannel(promise, channel, settings.connectionAttempts)
           Future.successful(channel)
         } else {
-          promise.tryFailure(new NoTargetException("No targets returned for name: " + settings.serviceName))
-          Future.failed(new IllegalStateException("No targets returned for name: " + settings.serviceName))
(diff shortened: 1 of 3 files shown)


fredfp:
I'm wondering about some kind of backoff mechanism, any advice or idea?

raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -57,6 +65,7 @@ object NettyClientUtils {
           ChannelUtils.monitorChannel(promise, channel, settings.connectionAttempts)
           Future.successful(channel)
         } else {
+          promise.tryFailure(new NoTargetException("No targets returned for name: " + settings.serviceName))
```
raboof: Perhaps return this exception as well?

fredfp: I would be fine with it, but this would break other people's code if they depend on the exception, are you fine with this?

fredfp:
My use case is not a direct connection to server nodes. I am using a kubernetes service as a load balancer in front of the actual nodes (i.e., some kind of infrastructure proxy a you said) and I still have no guarantee that it is deployed before the actual client.

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
raboof: Yes, I don't think it is likely people were depending on the specific exception, this is a clear improvement and we're 'allowed' to make such changes since we're still in the 0.x.y phase.

raboof:
> no guarantee that it is deployed before the actual client

Makes sense :+1:

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
fredfp: Alright, changed

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -19,6 +19,14 @@ import javax.net.ssl.SSLContext
 import scala.concurrent.duration.FiniteDuration
 import scala.concurrent.{ ExecutionContext, Future, Promise }
 
+/**
+ * Used to indicate that the service discovery returned no target.
+ *
+ * Can be caught to re-try lookup if it is likely that
+ * your service discovery mechanism will resolve to different instances.
+ */
+class NoTargetException(msg: String) extends RuntimeException(msg)
```
johanandren: "if you like it, make it final"

fredfp: done, thanks

PR #673, opened 2019-08-23
Use a dedicated managed sources directory

### LINKED ISSUE #672: Don't put generated sources in sourceManaged
sbt-protoc likes to completely own directories it writes to, when it runs, it first deletes all the old files from them. While that behavior can be turned off, it's good because otherwise sbt-protoc will leave stale files (eg, if you change the name of a message, the generated classes for the old message) lying around. But it does mean that you should never tell sbt-protoc to output to a shared directory. sbts `sourceManaged` directory is a shared directory, so if you have another sbt plugin that writes to it (eg, the sbt-buildinfo plugin), then you'll get very weird intermittent errors (because source generators run in parallel) where sbt-protoc will delete the files just generated by the other source generating plugin.

sbt-protoc doesn't actually specify an output directory, it requires the users to do that, but akka-grpc does. Akka grpc must define its own source directory to output to. I recommend using:

```scala
crossTarget.value / "akka-grpc" / Defaults.nameForSrc(configuration.value.name)
```

If you use this, you'll also have to add that directory to the `managedSourceDirectories` setting.

jroper:
Fixes #672

Since sbt-protoc deletes all files in its target directories, it's inappropriate to use the shared sourceManaged directory, because it will potentially clobber the files of other source generators immediately after they've generated them.

This gives akka-grpc its own target directory, which is managed heirarchically using `target in akkaGrpcCodeGeneratorSettings`.  Convention would say that it should be `target in PB.recompile`, since usually you scope the target setting to the task that generates to it, but since Akka doesn't provide `PB.recompile` itself, and sbt-protoc may some day decide to define `target in PB.recompile` for its own purposes, it's safer to use something that akka-grpc provides itself.  

So, the target directory is set to `target/scala_2.12/akka-grpc` at the no configuration scope, and then redefined in each configuration (test, compile) to be that plus `/main` or `/test`.

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -93,16 +93,15 @@ object AkkaGrpcPlugin extends AutoPlugin {
         // hack to get our (dirty) hands on a proper sbt logger before running the generators
         generatorLogger.logger = streams.value.log
         (PB.recompile in Test).value
-      },
-      target in akkaGrpcCodeGeneratorSettings := crossTarget.value / "akka-grpc")
+      })
 
   def configSettings(config: Configuration): Seq[Setting[_]] =
     inConfig(config)(
       (if (config == Compile || config == Test) Seq() // already supported by sbt-protoc by default
        else sbtprotoc.ProtocPlugin.protobufConfigSettings) ++
       Seq(
-        target in akkaGrpcCodeGeneratorSettings := (target in akkaGrpcCodeGeneratorSettings).value / Defaults
-            .nameForSrc(configuration.value.name),
+        target in akkaGrpcCodeGeneratorSettings := crossTarget.value / "akka-grpc" / Defaults.nameForSrc(
+            configuration.value.name),
         managedSourceDirectories += (target in akkaGrpcCodeGeneratorSettings).value,
         unmanagedResourceDirectories ++= (resourceDirectories in PB.recompile).value,
         watchSources in Defaults.ConfigGlobal ++= (sources in PB.recompile).value,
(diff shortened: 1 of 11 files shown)


sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -93,11 +93,15 @@ object AkkaGrpcPlugin extends AutoPlugin {
         // hack to get our (dirty) hands on a proper sbt logger before running the generators
         generatorLogger.logger = streams.value.log
         (PB.recompile in Test).value
-      })
+      },
+      target in akkaGrpcCodeGeneratorSettings := crossTarget.value / "akka-grpc")
```
johanandren: Isn't one issue with this that IntelliJ won't pick it up automagically anymore?

jroper: IntelliJ adds all `managedSourceDirectories` as generated source directories, so since we add this to that, it gets picked up.

jroper: Actually, maybe you're right. We should get that fixed in IntelliJ though.... needing to manually mark the directory as a generated sources root in intellij is better than a broken build... and the Scala compiler is really weird when you tell it to compile a non existent file, here's the error message you get:

```
error while decoding /home/travis/build/cloudstateio/cloudstate/java-support/target/src_managed/main/sbt-buildinfo/BuildInfo.scala with UTF-8: /home/travis/build/cloudstateio/cloudstate/java-support/target/src_managed/main/sbt-buildinfo/BuildInfo.scala (No such file or directory)
Please try specifying another one using the -encoding option
one error found
```

ennru:
In `gen-scala-server/06-compatibility-plugins` the generated class for test sources ends up in `target/scala-2.12/akka-grpc/main/test/example/myapp/echo/grpc/EchoMessage.scala` (see "**main**/test").
So something is not quite right there.

Issue #456, opened 2018-11-13
Configure PB.targets


jeongwoo-ji:
Is there some way to configure `PB.targets`?
If i configure `build.sbt` like below,
```scala
lazy val root = (project in file("."))
  .enablePlugins(AkkaGrpcPlugin)
  .enablePlugins(JavaAgent)
  .settings(
    PB.targets in Compile := Seq(
      scalapb.gen() -> (Compile / sourceManaged).value
    )
  )
```
then original scalapb generates the code (with StreamObserver..)

Issue #3, opened 2018-02-15
Integrate with sbt-protoc for scala codegen


ktoso:
Seems we want to integrate with https://github.com/thesamet/sbt-protoc rather than scalapb directly.
It also will allow us to generate java inside scala (sbt) projects.

It will also make adding our codegen easier, see:

```
PB.targets in Compile := Seq(
  scalapb.gen() -> (sourceManaged in Compile).value,
  MyCodeGenerator -> (sourceManaged in Compile).value
)
```

https://github.com/thesamet/sbt-protoc/tree/master/examples/custom-gen

PR #696, opened 2019-10-07
play nice on projects with other sbt-protoc-based plugins

bjaglin:
## Purpose

Updates the way generators are injected in the sbt plugin to coexist better with other plugins.

## References

Slightly related to https://github.com/akka/akka-grpc/issues/456 (although this one is about sbt settings of the project itself)

## Changes

I could not find in the original commit 7f2f96c or original issue https://github.com/akka/akka-grpc/issues/3 a reason for overriding `PB.targets`, so akka-grpc generators are now appended. Writing a scripted test to guard this behavior is hard as the order in which plugins are flattened is not guaranteed/deterministic.

## Background Context

When `AkkaGrpcPlugin` is added to a project with other plugins appending generators to `PB.targets`, these extra generators could be ignored, when `AkkaGrpcPlugin` was injected last (something which cannot be controlled, unless there is an explicit dependency from that plugin to `AkkaGrpcPlugin` via `AutoPlugin#requires`).

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -111,7 +111,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
         },
         // configure the proto gen automatically by adding our codegen:
         // FIXME: actually specifying separate Compile and Test target stub and languages does not work #194
-        PB.targets :=
+        PB.targets ++=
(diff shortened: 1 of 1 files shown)


raboof:
OK TO TEST

raboof:
I understand this is a draft, but just wanted to share that I don't remember any particular reason for overriding rather than appending, so this looks great to me ;)

bjaglin:
@raboof - I just wanted to wait for CI to be green to flap the review status ;)

raboof:
Aah cool ;) - feel free to go straight to 'open', whatever you prefer

Issue #194, opened 2018-05-03
Different settings for Compile/Test with sbt


johanandren:
The current somewhat complicated sbt settings are motivated by the idea that it would be possible to generate only a server in `Compile` and then the client in `Test` to do integration tests in the same project. However, there is no test coverage for this and when I tried to create a scripted test it seems it does in fact not work.

I think we should either make sure it works, has a test covering it and is described in the sbt plugin docs or simplify the plugin settings and simply not support it.

viktorklang:
I managed to get this to work without any code changes,

This is what I added to be able to do it:

```scala
    inConfig(Test)(
      sbtprotoc.ProtocPlugin.protobufConfigSettings ++ Seq(
        PB.protoSources ++= Seq(sourceDirectory.value / "protos"),
        akkaGrpcGeneratedSources := Seq(AkkaGrpc.Client),
      )
    ),
```

This configures the ProtocPlugin to work under Test, and then adds test/protos as the proto sources to compile, and the akkaGrpcGeneratedSources will then only generate the client.

raboof:
Together with #702 I think we can now consider this issue closed

PR #702, opened 2019-10-10
sbt: improve support for configurations

bjaglin:
## Purpose

After integrating sbt-akka-grpc 0.7.2 in my project, I realized #696 introduced a regression for the `Test` configuration (`protoc` hangs every single time, because the same `JvmGenerator` is used multiple times, most likely resulting in a race on the FIFOs used to bridge `protoc` to the JVM generators), so I looked at how `Test` configuration could be better supported/tested.

## References

* Fix regression introduced by #696 
* Make updates so that #194 does not regress, enforce its behavior via scripted & document it

## Changes

See commit messages

## Background Context

* See https://github.com/thesamet/sbt-protoc/issues/108 for how/why this was fixed upstream.
* https://github.com/scalapb/protoc-bridge/pull/52 also "fixes" the problem as `protoc` no longer hangs, but hides the problem that generators are ran twice when they should not.
* The fix for https://github.com/thesamet/sbt-protoc/issues/108 actually broke the [out-of-the-box solution](https://github.com/akka/akka-grpc/issues/194#issuecomment-502995198) for #194, as `protocSources` are not inherited from one configuration (`Compile`) to its child (`Test`), so some extra `ScopeFilter` logic is required to be able to capture proto sources that are not part of the configuration itself to feed them to the local generators.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -67,8 +67,7 @@ object Dependencies {
   }
 
   object Plugins {
-    // This should follow the version of sbt-protoc ScalaPB is depending on
-    val sbtProtoc = "com.thesamet" % "sbt-protoc" % "0.99.19"
+    val sbtProtoc = "com.thesamet" % "sbt-protoc" % "0.99.26"
   }
 
   private val l = libraryDependencies
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -100,6 +100,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
 
   def configSettings(config: Configuration): Seq[Setting[_]] =
     inConfig(config)(
+      sbtprotoc.ProtocPlugin.protobufConfigSettings ++
       Seq(
         unmanagedResourceDirectories ++= (resourceDirectories in PB.recompile).value,
         watchSources in Defaults.ConfigGlobal ++= (sources in PB.recompile).value,
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -114,10 +114,11 @@ object AkkaGrpcPlugin extends AutoPlugin {
             toGenerator(g, ScalaBinaryVersion(scalaBinaryVersion.value), generatorLogger))
         },
         // configure the proto gen automatically by adding our codegen:
-        // FIXME: actually specifying separate Compile and Test target stub and languages does not work #194
         PB.targets ++=
           targetsFor(sourceManaged.value, akkaGrpcCodeGeneratorSettings.value, akkaGrpcGenerators.value),
         PB.protoSources += sourceDirectory.value / "proto",
+        // include proto files from parent configurations to be able to run extra generators or generators with different settings
+        PB.protoSources ++= PB.protoSources.all(ancestorConfigsFilter(config)).value.flatten,
         // include proto files extracted from the dependencies with "protobuf" configuration by default
         PB.protoSources += PB.externalIncludePath.value) ++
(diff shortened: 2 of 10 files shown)


bjaglin:
Reopening to restart CI after that seems to be a transient failure

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -110,9 +114,16 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
+          val configTargets =
+            targetsFor(sourceManaged.value, akkaGrpcCodeGeneratorSettings.value, akkaGrpcGenerators.value)
+          // To play nice with other sbt-protoc plugins that might have injected their own targets, don't simply
+          // override the key. Append instead, after excluding targets added by parent configurations (transitively)
+          // since they already ran and would cause protoc to freeze as their name collides with the own generators
+          // of the config.
+          val parentConfigsGenerators = akkaGrpcGenerators.all(ancestorConfigsFilter(config)).value.flatten
+          PB.targets.value.filterNot(t => parentConfigsGenerators.contains(t.generator)) ++ configTargets
+        },
         PB.protoSources += sourceDirectory.value / "proto",
```
bjaglin: I thought about adding the same "exclude parent config entries" behavior to this key for consistency. That would avoid sending the same input to `protoc` potentially several times (and ending up with the same class file (s) shadowing the one (s) compiled earlier), but I realized that would go against #194.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -110,9 +114,16 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
         // configure the proto gen automatically by adding our codegen:
-        // FIXME: actually specifying separate Compile and Test target stub and languages does not work #194
-        PB.targets ++=
-          targetsFor(sourceManaged.value, akkaGrpcCodeGeneratorSettings.value, akkaGrpcGenerators.value),
+        PB.targets := {
+          val configTargets =
+            targetsFor(sourceManaged.value, akkaGrpcCodeGeneratorSettings.value, akkaGrpcGenerators.value)
+          // To play nice with other sbt-protoc plugins that might have injected their own targets, don't simply
+          // override the key. Append instead, after excluding targets added by parent configurations (transitively)
+          // since they already ran and would cause protoc to freeze as their name collides with the own generators
```
bjaglin: If

bjaglin: https://github.com/thesamet/sbt-protoc/issues/108

bjaglin: https://github.com/thesamet/sbt-protoc/pull/111 was merged, so I'll replace https://github.com/akka/akka-grpc/pull/702/commits/9b35ce64fc0ecfa8abfabcbf409577fe2387f6ab by a `sbt-protoc` bump once a new release is cut.

raboof:
OK TO TEST

project/Dependencies.scala
```diff
@@ -67,8 +67,7 @@ object Dependencies {
   }
 
   object Plugins {
-    // This should follow the version of sbt-protoc ScalaPB is depending on
```
bjaglin: this was not honored, and there is nothing that really requires it anyway, so I removed this note

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
bjaglin: Initial implementation replaced by the bump in https://github.com/akka/akka-grpc/pull/702/commits/76eba3c76e917c9396fa10f5f10ee3709313d0e6

bjaglin:
I just revisited and force-pushed this as a fix was released in `sbt-protoc`. Description updated.

PR #715, opened 2019-10-22
Enable the client-side load balancing provided by grpc-java

pasikon:
## Purpose

Allows to use client-side load balancing by providing akka-grpc configuration options

## Changes

- configuration options for client-side LB
- setting up NettyChannelBuilder with client-side LB

## Background Context

I just need client-side LB ;)

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -95,6 +95,13 @@ object GrpcClientSettings {
         if (serviceName.isEmpty)
           serviceName = "static"
         staticServiceDiscovery(host, port)
+      case "grpc-dns" =>
+        val host = clientConfiguration.getString("host")
+        require(host.nonEmpty, "host can't be empty when service-discovery-mechanism is set to grpc-dns")
+        // Required by the Discovery infrastructure, even when we use static discovery.
+        if (serviceName.isEmpty)
+          serviceName = "static"
+        staticServiceDiscovery(host, port)
       case other =>
         require(serviceName.nonEmpty, "Configuration must contain a service-name")
         Discovery(sys).loadServiceDiscovery(other)
@@ -116,7 +123,8 @@ object GrpcClientSettings {
       userAgent = getOptionalString(clientConfiguration, "user-agent"),
       sslContext = getOptionalSSLContext(clientConfiguration, "ssl-config"),
       connectionAttempts = getOptionalInt(clientConfiguration, "connection-attempts"),
-      useTls = clientConfiguration.getBoolean("use-tls"))
+      useTls = clientConfiguration.getBoolean("use-tls"),
+      grpcLoadBalancingType = getOptionalString(clientConfiguration, "grpc-load-balancing"))
 
   private def getOptionalString(config: Config, path: String): Option[String] = config.getString(path) match {
     case ""    => None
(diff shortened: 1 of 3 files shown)


raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -57,6 +58,10 @@ object NettyClientUtils {
               .getOrElse(builder.negotiationType(NegotiationType.PLAINTEXT))
           }
 
+          builder = settings.grpcLoadBalancingType
+            .map(builder.defaultLoadBalancingPolicy(_))
+            .map(_ => builder.nameResolverFactory(new DnsNameResolverProvider()))
```
raboof: This indeed works because the build methods are side-effecting, but wouldn't it be slightly neater to:

```suggestion
            .map(_.nameResolverFactory(new DnsNameResolverProvider()))
```

?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -95,6 +95,13 @@ object GrpcClientSettings {
         if (serviceName.isEmpty)
           serviceName = "static"
         staticServiceDiscovery(host, port)
+      case "grpc-dns" =>
```
raboof: So there is currently no difference between `static` and `grpc-dns`? Perhaps we could use `case "static" | "grpc-dns"` to make that clearer?

(it is a bit strange to introduce a `grpc-dns` discovery method in the config, but actually have a 'static' implementation for it instead of having a 'real' `grpc-dns` discovery implementation.

I think this is acceptable as long as we're using the netty client implementation, so this is a special-case delegating the discovery to that. Eventually when we move from netty to akka-http for the client this would probably change, but that's OK)

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -57,6 +58,10 @@ object NettyClientUtils {
...
+            .getOrElse(builder)
```
raboof: So this is applied only when `settings.grpcLoadBalancingType` is `Some(_)` :+1:

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
pasikon: Sure!

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
pasikon: Indeed thats a bit strange, I named service discovery mechanism `grpc-dns` to indicate that we are delegating actual discovery to grpc internals, not ideal but had no better idea...

raboof:
(the travis failure is #724, so unrelated and can be ignored here)

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -88,9 +88,9 @@ object GrpcClientSettings {
     val port = clientConfiguration.getInt("port")
     val resolveTimeout = clientConfiguration.getDuration("service-discovery.resolve-timeout").asScala
     val sd = serviceDiscoveryMechanism match {
-      case "static" =>
+      case "static" | "grpc-dns" =>
```
raboof: Is adding `grpc-dns` here actually needed? It looks like the `static` discovery method will already have the desired effect?

pasikon: Yes, because when enabling GRPC LB `DnsNameResolverProvider` is being used. I also wanted to reflect that in the configuration.

Issue #719, opened 2019-10-31
Don't pass unknown generator options to the java code generator


raboof:
The built-in protoc Java code generator will throw an `Unknown generator option` error when it encounters an option it does not know.

Currently we pass it all `akkaGrpcCodeGeneratorSettings` except those we know are options meant for scalapb or our own generators. We should probably turn this around and only pass in options we know are valid for the java code generator.

Disadvantage: when options are added to the java code generator, we have to update akka-grpc to make them available to our users.
Advantage: when options are added to any other code generators, we no longer have to update akka-grpc to make them available to our users.

raboof:
(relevant code is at https://github.com/akka/akka-grpc/blob/v0.7.2/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L143)

viktorklang:
Here's a start for a possible solution, although I am not sure how it can be tested:

```patch
diff --git sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
index f7f3a55e..7d1ecf9b 100644
--- sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -59,6 +59,16 @@ object AkkaGrpcPlugin extends AutoPlugin {
       sealed trait Language
       case object Scala extends Language
       case object Java extends Language
+
+      val acceptedOptions: Map[Language, Set[String]] = Map(
+        Scala -> Set(
+          "java_conversions",
+          "flat_package",
+          "single_line_to_proto_string",
+          "ascii_format_to_string",
+          "no_lenses",
+          "retain_source_code_info"),
+        Java -> Set("single_line_to_proto_string", "ascii_format_to_string", "retain_source_code_info"))
     }

     val akkaGrpcGeneratedLanguages = settingKey[Seq[AkkaGrpc.Language]](
@@ -70,7 +80,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
     val akkaGrpcGenerators = settingKey[Seq[protocbridge.Generator]](
[...]

PR #720, opened 2019-11-01
Whitelist allowed protoc settings

raboof:
This is needed because the built-in java generator and the scalapb scala
generator throw an error when an unrecognized option is passed to them.

Added for sbt and maven, not needed yet for the groovy plugin because
there we don't allow arbitrary settings.

Refs #719


johanandren:
Ups, a bit eager on LGTM there, a file missing from the checkin?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -139,12 +142,17 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
-        case protocbridge.JvmGenerator("scala", ScalaPbCodeGenerator) =>
-          settings.filterNot(GeneratorOption.settings.contains)
-        case _ => settings
-      })
+      protocbridge.Target(
+        generator,
+        targetPath,
+        generator match {
+          case PB.gens.java =>
+            settings.filter(ProtocSettings.protocJava.contains)
```
johanandren: One thought: if we pass these indirectly via a sbt setting it would be possible to change it without a new release of the plugin (as a workaround)

raboof: I like the idea, but in practice it requires passing around the values or wrapping functions in `Def.taskDyn` and friends - I think it's not worth it at this point.

PR #727, opened 2019-11-07
Retry channel creation when it fails

raboof:
Retry when creating the channel initially already fails.

Refs #728

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/ClientState.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/ClientState.scala
@@ -35,12 +35,12 @@ final class ClientState(
   def this(settings: GrpcClientSettings, log: LoggingAdapter)(implicit mat: Materializer, ex: ExecutionContext) =
     this(settings, log, s => NettyClientUtils.createChannel(s))
 
-  private val internalChannelRef = new AtomicReference[Option[Future[InternalChannel]]](Some(create()))
-  internalChannelRef.get().foreach(c => recreateOnFailure(c.flatMap(_.done), settings.creationAttempts))
-
   // usually None, it'll have a value when the underlying InternalChannel is closing or closed.
   private val closing = new AtomicReference[Option[Future[Done]]](None)
   private val closeDemand: Promise[Done] = Promise[Done]()
+
+  private val internalChannelRef = new AtomicReference[Option[Future[InternalChannel]]](Some(create()))
+  internalChannelRef.get().foreach(c => recreateOnFailure(c.flatMap(_.done), settings.creationAttempts))
 
   mat match {
     case m: ActorMaterializer =>
@@ -112,25 +112,32 @@ final class ClientState(
   private def recreateOnFailure(done: Future[Done], creationsLeft: Int): Unit =
     done.onComplete {
       case Failure(e) =>
-        log.warning(s"Client error [${e.getMessage}], recreating client after ${settings.creationDelay}")
-        // TODO #733 remove cast once we update Akka
-        val scheduler = mat.asInstanceOf[ActorMaterializer].system.scheduler
-        val ec = mat.asInstanceOf[ActorMaterializer].system.dispatcher
-        Patterns.after(
-          settings.creationDelay,
-          scheduler,
-          ec,
-          () =>
-            Future {
-              if (!closeDemand.isCompleted) {
-                if (creationsLeft > 0)
+        if (creationsLeft <= 0) {
+          // Error does not need to be explicitly propagated here
+          // since it's in the internalChannelRef already
+          log.warning(s"Client error [${e.getMessage}], not recreating client")
+          close()
+        } else if (settings.creationDelay.length < 1) {
+          if (!closeDemand.isCompleted) {
+            log.warning(s"Client error [${e.getMessage}], recreating client")
+            recreate(creationsLeft - 1)
+          }
+        } else {
+          log.warning(s"Client error [${e.getMessage}], recreating client after ${settings.creationDelay}")
+
+          Patterns.after(
+            settings.creationDelay,
+            // TODO #733 remove cast once we update Akka
+            mat.asInstanceOf[ActorMaterializer].system.scheduler,
+            mat.asInstanceOf[ActorMaterializer].system.dispatcher,
+            () =>
+              Future {
+                log.info("Recreating channel now")
+                if (!closeDemand.isCompleted) {
                   recreate(creationsLeft - 1)
-                else
-                  // Error does not need to be explicitly propagated here
-                  // since it's in the internalChannelRef already
-                  close()
-              }
-            })
+                }
+              })
+        }
       case _ =>
         log.info("Client closed")
       // completed successfully, nothing else to do (except perhaps log?)
--- a/runtime/src/main/scala/akka/grpc/internal/ClientState.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/ClientState.scala
@@ -15,7 +15,7 @@ import io.grpc.ManagedChannel
 
 import scala.annotation.tailrec
 import scala.concurrent.{ ExecutionContext, Future, Promise }
-import scala.util.Failure
+import scala.util.{ Failure, Success }
 import scala.compat.java8.FutureConverters._
 
 /**
@@ -43,6 +43,7 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
     case _ =>
   }
 
+  // only used in tests
   def withChannel[A](f: Future[ManagedChannel] => A): A =
     f {
       internalChannelRef.get().getOrElse(throw new ClientClosedException).managedChannel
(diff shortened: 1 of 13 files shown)


runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -97,24 +98,40 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
...
+          case _ =>
+          // let success through
         }
       case Failure(_) =>
-        close()
-      case _ =>
-      // let success through
+        // Would be better to retry with backoff
+        // Would be good to log
+        recreate()
```
johanandren: I guess we could use `akka.pattern` `retry` for a cheap backoff, delay there is constant but better than nothing?

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -62,15 +61,17 @@ object NettyClientUtils {
           builder = settings.channelBuilderOverrides(builder)
 
           val channel = builder.build()
+
+          val promise = Promise[Done]()
```
johanandren: Um, when is this ever completed?

johanandren: Ignore that, it's on the next line!

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
johanandren: I take that back, I quick skimmed and missed how it is side effecting, so probably not possible to leverage that.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -94,25 +98,41 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
...
-  private def create(): InternalChannel = {
-    val internalChannel: InternalChannel = channelFactory(settings)
-    internalChannel.done.onComplete {
+  private def create(): Future[InternalChannel] =
+    Patterns.retry(
+      () => channelFactory(settings),
+      // TODO get from settings
+      1000,
+      400.millis,
+      // TODO remove case once we update Akka
```
chbatey: cast

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -94,25 +98,41 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
     }
   }
 
-  private def create(): InternalChannel = {
-    val internalChannel: InternalChannel = channelFactory(settings)
-    internalChannel.done.onComplete {
+  private def create(): Future[InternalChannel] =
+    Patterns.retry(
+      () => channelFactory(settings),
```
chbatey: So we retry creation for any type of exception but once it has been re-created we only retry for the two specific ones

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -94,25 +98,41 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
...
+      // TODO get from settings
+      1000,
+      400.millis,
+      // TODO remove case once we update Akka
+      mat.asInstanceOf[ActorMaterializer].system.scheduler,
+      mat.asInstanceOf[ActorMaterializer].system.dispatcher)
+
+  private def recreateOnFailure(done: Future[Done]): Unit =
+    done.onComplete {
       case Failure(_: ClientConnectionException | _: NoTargetException) =>
```
chbatey: Is NoTargetException needed any ore here? Won't that be caught by the outer retry as it is on creation?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
raboof: I agree that's a bit weird. my reasoning was that since we delay between attempts it's not so heavy - for the re-creation we don't have a delay yet so non-transient errors would 'spin'. I think once we add a delay there it might make sense to retry for all exceptions there, too.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
raboof: Ah, indeed that can be dropped there I think.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
raboof: (it would also be good to have a (configurable, perhaps resetting) limit on the number of retries here)

raboof:
> Would it be possible to do some form of backoff?

Yes, perhaps with https://github.com/akka/akka/pull/27290 . Let's keep it out of scope for this PR though.

runtime/src/test/scala/akka/grpc/GrpcClientSettingsSpec.scala
```diff
@@ -153,10 +155,12 @@ class GrpcClientSettingsSpec extends WordSpec with Matchers with ScalaFutures {
     }
 
     "fail to parse configuration with non-existent certificate" in {
+      val system = sysWithCert("no-such-cert.pem")
       val thrown = the[IllegalArgumentException] thrownBy
-        GrpcClientSettings.fromConfig("project.WithSpecificConfiguration")(sysWithCert("no-such-cert.pem"))
+        GrpcClientSettings.fromConfig("project.WithSpecificConfiguration")(system)
       // We want a good message since missing classpath resources are difficult to debug
       thrown.getMessage should include("certs/no-such-cert.pem")
+      system.terminate()
```
johanandren: try-finally because now if it fails it leaks the system, also `TestKit.shutdownActorSystem`

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -94,25 +100,40 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
...
+      mat.asInstanceOf[ActorMaterializer].system.dispatcher)
+
+  private def recreateOnFailure(done: Future[Done]): Unit =
+    done.onComplete {
+      case Failure(e: ClientConnectionException) =>
+        log.warning(s"Transient connection error [${e.getMessage}], recreating client")
+        recreate()
+      case Failure(e) =>
+        // TODO This makes the client unusable. Perhaps we should retry here too?
+        log.warning(s"Unrecoverable connection error [${e.getMessage}], closing client")
```
johanandren: Yeah, maybe retry for all exceptions instead and see if that has any surprising consequences?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
johanandren: I think we should use the the configured delay and max retries here as well (using `akka.pattern.Patterns#after` and passing the recreateCount as a parameter to `recreateOnFailure` and `recreate` for example).

Having a backoff and max retries only apply to initial connection seems quite confusing to me.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -94,25 +100,47 @@ final class ClientState(settings: GrpcClientSettings, channelFactory: GrpcClient
...
+          () =>
+            Future {
+              if (!closeDemand.isCompleted)
+                recreate(creationsLeft - 1)
+            })
       case _ =>
-      // let success through
+        log.info("Client closed")
+      // completed successfully, nothing else to do (except perhaps log?)
+    }
```
johanandren: Probably stop at some point when `creationsLeft` reached 0 as well?

raboof: forget me own head next :hushed:

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -122,8 +122,14 @@ final class ClientState(
...
-              if (!closeDemand.isCompleted)
-                recreate(creationsLeft - 1)
+              if (!closeDemand.isCompleted) {
+                if (creationsLeft > 0)
+                  recreate(creationsLeft - 1)
+                else
+                  // Error does not need to be explicitly propagated here
+                  // since it's in the internalChannelRef already
+                  close()
+              }
```
johanandren: Better :)

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -35,13 +35,13 @@ final class ClientState(
...
     this(settings, log, s => NettyClientUtils.createChannel(s))
 
-  private val internalChannelRef = new AtomicReference[Option[Future[InternalChannel]]](Some(create()))
-  internalChannelRef.get().foreach(c => recreateOnFailure(c.flatMap(_.done), settings.creationAttempts))
-
   // usually None, it'll have a value when the underlying InternalChannel is closing or closed.
   private val closing = new AtomicReference[Option[Future[Done]]](None)
   private val closeDemand: Promise[Done] = Promise[Done]()
 
+  private val internalChannelRef = new AtomicReference[Option[Future[InternalChannel]]](Some(create()))
```
johanandren: Flatmap this! 😆

PR #736, opened 2019-11-14
Update scalapb to 0.10.1

raboof:
This breaks binary compatibility of the generated classes (but not wire compatibility)


raboof:
WIP until 0.10.0 is actually released (it's just a milestone so far)

Issue #735, opened 2019-11-14
Remove unnecessary printout of gradle tasks on plugin apply


kudlatykaskader:

raboof:
Good one - would you be interested in picking this up?

jewertow:
What exactly do you mean? Could you paste logs and say when you get it?

raboof:
There is a "println project.getTasks()" in
https://github.com/akka/akka-grpc/blob/master/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
, probably it?

jewertow:
@raboof Thanks for giving the opportunity to make the first change in that project.

raboof:
Thanks for the PR!

kudlatykaskader:
Great! 
Sorry, I had notifications turned off and didn't saw that you picked it.

PR #739, opened 2019-11-22
Remove unnecessary print out of gradle tasks on plugin apply

jewertow:
## Purpose

Gradle plugin unnecessary prints out a list o tasks.

## References

See #735

## Changes

- Removed printing tasks


raboof:
OK TO TEST

PR #765, opened 2019-12-19
Use same sbt-protoc version everywhere

renatocaval:
akka-grpc plugin is depending on sbt-protoc v0.99.26, but the project itself is using v0.99.19.  

play-grpc is pulling v0.99.26 and failing to build because v0.99.26 sets `PB.targets := Nil` after we set it to ListBuffer (in ReflectiveCodeGen). Later, when we try to cast it back to ListBuffer, we get an error because we can't cast `immutable.Nil` to `ListBuffer`. 

This is not yet a fix for that issue, but it makes sure that the version we use in akka-grpc is the same that we add as a dependency in the sbt plugin we produce. 

Because the issue is not yet solved in `ReflectiveCodeGen`, I'm reverting the version to 0.99.25 as it doesn't set `PB.targets` to `Nil`. 

We do have a fix in play-grpc's ReflectiveCodeGen that allows us to use 0.99.26, but that same fix doesn't work on akka-grpc. I'm still looking for a solution.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,6 +1,16 @@
+
+
+enablePlugins(BuildInfoPlugin)
+
+val sbtProtocV = "0.99.25"
+
+buildInfoKeys := Seq[BuildInfoKey](
+  "sbtProtocVersion"      -> sbtProtocV
+)
+
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.0")
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.3.1")
-addSbtPlugin("com.thesamet" % "sbt-protoc" % "0.99.19")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % sbtProtocV)
 addSbtPlugin("com.dwijnand" % "sbt-dynver" % "4.0.0")
 addSbtPlugin("com.typesafe.sbt" % "sbt-twirl" % "1.5.0")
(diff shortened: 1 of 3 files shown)


project/project/buildinfo.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
```
raboof: So meta!

project/plugins.sbt
```diff
@@ -1,6 +1,12 @@
+enablePlugins(BuildInfoPlugin)
+
+val sbtProtocV = "0.99.25"
```
raboof: Good to make those consistent!

I think I've indeed seen that 'listbuffer casting' issue indeed, though I'm not sure where - would indeed be great to fix that.

I'm not sure about he downgrade from 0.99.26 to 0.99.25: while it's only a patchversion, there's a number of commits in there, including the https://github.com/thesamet/sbt-protoc/commit/ac1cfa086577f0e5a9a1bf5f5a14d464a7c550a8 revert by @bjaglin, and I'm not sure if downgrading would lead to trouble here...

Not quite sure about the downgrade though -

renatocaval: Sure, a fix is the final intention. Let's keep it open. I will try to find a solution that is last hacky and more stable.

renatocaval:
superseded by #788

PR #766, opened 2019-12-19
Make sure akka-grpc can be cross-built

raboof:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.5
+sbt.version=1.3.4
```
patriknw: downgrade? was there a problem?

raboof: yes, for some reason since the bump `sbt +compile` would also try to compile the sbt plugin with scala 2.13, which won't work. I'd like to make a reproducer and report upstream but one thing at a time ;)

PR #768, opened 2019-12-19
Require akka 2.6

raboof:
This branch is intended to be the starting point for Akka gRPC 1.1.0. We expect to rebase it onto master periodially while we work towards 1.0.0 there.

Refs #751

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
@@ -11,6 +11,7 @@ import akka.grpc.GrpcClientSettings
 import akka.grpc.internal.ClientConnectionException
 import akka.grpc.scaladsl.tools.MutableServiceDiscovery
 import akka.http.scaladsl.Http
+import akka.stream.{ Materializer, SystemMaterializer }
 import example.myapp.helloworld.grpc.helloworld._
 import io.grpc.Status.Code
 import io.grpc.StatusRuntimeException
@@ -26,7 +27,7 @@ import scala.concurrent.duration._
 
 class LoadBalancingIntegrationSpec extends AnyWordSpec with Matchers with BeforeAndAfterAll with ScalaFutures {
   implicit val system = ActorSystem("NonBalancingIntegrationSpec")
-  implicit val mat = akka.stream.ActorMaterializer.create(system)
+  implicit val mat = SystemMaterializer(system).materializer
   implicit val ec = system.dispatcher
 
   override implicit val patienceConfig = PatienceConfig(5.seconds, Span(10, org.scalatest.time.Millis))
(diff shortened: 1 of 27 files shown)


patriknw:
why is this needed this early? for compatibility checks?

raboof:
yes, the plan would be to use this branch to make sure there are no API's in 1.0.0 that would later make it hard to require Akka 2.6 without breaking backwards compatibility in 1.1.0. If this branch turns out to be hard to maintain we can drop it and reintroduce it later or find another way.

project/Dependencies.scala
```diff
@@ -9,7 +9,7 @@ object Dependencies {
     val scala212 = "2.12.10"
     val scala213 = "2.13.1"
 
-    val akka = "2.5.27"
+    val akka = "2.6.2"
```
patriknw: hold this until 2.6.3 since 2.6.2 is dead on arrival

ignasi35:
Did a rebase of this one on top of `master`. I've pushed it to https://github.com/ignasi35/akka-grpc/tree/require-akka-2.6 (to not `push --force` here yet).

project/Dependencies.scala
ignasi35: The PR is now using `2.6.4`. Is there anything we explicitly need from `2.6.4`? We'd be better off requiring _only_ `2.6.3`.  Hmm, I'll try that out.

ignasi35: Ok, `2.6.4` is required because of, at least, `parasitic`.

raboof:
Rebased and cleaned up a bit, I think this is now good to go.

interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
```diff
@@ -55,7 +56,7 @@ class LoadBalancingIntegrationSpec extends AnyWordSpec with Matchers with Before
     }
 
     "re-discover endpoints on failure" in {
-      val service1materializer = akka.stream.ActorMaterializer.create(system)
+      val service1materializer = Materializer(system)
```
ennru, line 59: Is there a relevant difference to `SystemMaterializer(system).createAdditionalSystemMaterializer()` used in `NonBalancingIntegrationSpec`?

plugin-tester-scala/src/main/scala/example/myapp/helloworld/PowerGreeterServiceImpl.scala
```diff
@@ -13,9 +13,8 @@ import example.myapp.helloworld.grpc._
 
 import scala.concurrent.Future
 
-class PowerGreeterServiceImpl(materializer: Materializer) extends GreeterServicePowerApi {
+class PowerGreeterServiceImpl(implicit materializer: Materializer) extends GreeterServicePowerApi {
```
ennru, line 16: Shouldn't this take an `ActorSystem` now?

project/Dependencies.scala
```diff
@@ -7,7 +7,7 @@ import buildinfo.BuildInfo
 object Dependencies {
   object Versions {
     val scala212 = "2.12.11"
-    val scala213 = "2.13.1"
+    val scala213 = "2.13.2"
```
ennru, line 10: Take the latest?

ennru (approved):
LGTM.

patriknw (approved):
LGTM

Issue #774, opened 2019-12-22
"head of empty stream" exception when sending response in akka-http route


jokester:
### Versions used 

akka version: 2.6.1
akka-http version: 10.1.11
akka-grpc version: 0.7.3
rest: ArchLinux x64 / openjdk 11 / scala 2.13.1

### Steps to produce

1. clone the minimal example at https://github.com/jokester/experimental-akka-grpc (it contains a dummy grpc service implementation that returns `Future.failed`), and `sbt run`
2. in another terminal, run the following command:

```
curl 'http://localhost:9090/examplecom.library.BookService/GetBook' --compressed -H 'content-type: application/grpc-web+proto' -H 'x-grpc-web: 1' --data $'\x00\x00\x00\x00\x00' | wc
```

### Expected Behavior

- curl will read some bytes from chunked (I expected there to be at least 1 trailing chunk for `grpc-status`)
- akka throws no exception

### Actual Behavior

- curl gets 0 bytes
- the following exception thrown in akka

// the stream might be the one created in [GrpcResponseHelpers](https://github.com/akka/akka-grpc/blob/4bef58f0fe8f6bd91333b51722a4153be6021713/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala#L64), but I was not able to go deeper

```
20:13:10.211 [main-system-akka.actor.default-dispatcher-9] ERROR akka.actor.ActorSystemImpl - head of empty stream
java.util.NoSuchElementException: head of empty stream
	at akka.stream.scaladsl.Sink$.$anonfun$head$3(Sink.scala:197)
	at scala.Option.getOrElse(Option.scala:201)
	at akka.stream.scaladsl.Sink$.$anonfun$head$2(Sink.scala:197)
	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:430)
	at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
	at akka.dispatch.BatchingExecutor$Batch.run(BatchingExecutor.scala:73)
	at akka.dispatch.ExecutionContexts$sameThreadExecutionContext$.unbatchedExecute(Future.scala:85)
	at akka.dispatch.BatchingExecutor.execute(BatchingExecutor.scala:122)
	at akka.dispatch.BatchingExecutor.execute$(BatchingExecutor.scala:116)
	at akka.dispatch.ExecutionContexts$sameThreadExecutionContext$.execute(Future.scala:84)
[...]

PR #779, opened 2020-01-02
Return INVALID_ARGUMENT when no parameter is found

raboof:
Refs #774

### DIFF
--- a/runtime/src/main/scala/akka/grpc/scaladsl/GrpcMarshalling.scala
+++ b/runtime/src/main/scala/akka/grpc/scaladsl/GrpcMarshalling.scala
@@ -7,7 +7,7 @@ package akka.grpc.scaladsl
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.grpc._
-import akka.grpc.internal.{ CancellationBarrierGraphStage, GrpcResponseHelpers }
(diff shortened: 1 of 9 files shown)


runtime/src/main/scala/akka/grpc/scaladsl/GrpcMarshalling.scala
```diff
@@ -19,7 +19,17 @@ import scala.concurrent.Future
...
-    req.entity.dataBytes.via(Grpc.grpcFramingDecoder(messageEncoding)).map(u.deserialize).runWith(Sink.head)(mat)
+    implicit val ec = mat.executionContext
+    req.entity.dataBytes
+      .via(Grpc.grpcFramingDecoder(messageEncoding))
+      .map(u.deserialize)
+      .runWith(Sink.headOption)(mat)
+      .flatMap {
+        _ match {
+          case Some(element) => Future.successful(element)
+          case None          => Future.failed(new MissingParameterException())
```
johanandren: Does this only happen on a missing parameter, should it rather be something like "invalid request" rather because it could be anything with the encoding/serialized form? Or would that lead to a failed stream already rather?

raboof: Yes, I would expect other errors to lead to a failed stream rather than an empty one

PR #782, opened 2020-01-08
Add coursier cache

raboof:


johanandren:
Oh wait, it's the other way around `~/.coursier/cache/`

raboof:
Are you sure? e.g. https://github.com/scala/scala-xml/pull/357/files#r324386303

johanandren:
That's where they live on my machine at least

johanandren:
Ah, I see, nvm then I guess.

johanandren:
Hah, actually on MacOS, after deleting `~/.courser` it now uses `~/Library/Caches/Coursier/` (which I guess is where MacOS wants it but still...).

jrudolph:
> Hah, actually on MacOS, after deleting `~/.courser` it now uses `~/Library/Caches/Coursier/` (which I guess is where MacOS wants it but still...).

That behavior seems to be documented here: 

https://get-coursier.io/docs/cache.html#compatibility-with-former-versions

Issue #786, opened 2020-01-11
Log file can not be found for multiproject Gradle projects


orendain:
### Versions used 

Akka version: 
akka-grpc-gradle-plugin:0.7.3

### Expected Behavior

When building a Gradle project with multiple subprojects (e.g., a root gradle build and at least one subproject build), building the project from the root project should work.

Specifically: `./gradlew build` from the root project's directory should complete successfully.

The same goes for if the project is built not from the root project, but from the specific subproject.

### Actual Behavior

If the build is triggered from the **_subproject_**, things work as expected for that subproject.

If the build is triggered from the **_root_**, the Gradle plugin complains that a log does not exist (see below for relevant logs).

This regression is due to a change made in PR #674, which addresses issue #669.

### Relevant logs

```
> Task :subProject:generateProto FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':subProject:generateProto'.
> protoc: stdout: . stderr: /Users/eorendain/Documents/myProject/subProject/src/main/protobuf: warning: directory does not exist.
  /Users/eorendain/Documents/myProject/subProject/app/protobuf: warning: directory does not exist.
  /Users/eorendain/Documents/myProject/subProject/app/proto: warning: directory does not exist.
  bytestream.proto: warning: Import google/protobuf/wrappers.proto but not used.
  Exception in thread "main" java.io.FileNotFoundException: build/akka-grpc-gradle-plugin.log (No such file or directory)
        at java.base/java.io.FileOutputStream.open0(Native Method)
        at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
        at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:187)
        at java.base/java.io.PrintWriter.<init>(PrintWriter.java:208)
        at java.base/java.io.PrintWriter.<init>(PrintWriter.java:248)
        at akka.grpc.gen.FileLogger.<init>(Logging.scala:35)
        at akka.grpc.gen.Main$.$anonfun$logger$1(Main.scala:48)
        at scala.Option.map(Option.scala:230)
        at akka.grpc.gen.Main$.delayedEndpoint$akka$grpc$gen$Main$1(Main.scala:48)
        at akka.grpc.gen.Main$delayedInit$body.apply(Main.scala:14)
        at scala.Function0.apply$mcV$sp(Function0.scala:39)
[...]

PR #787, opened 2020-01-11
Adjusted Gradle plugin logfile to be generated as an absolute path.

orendain:
## Purpose

To attempt to correct the log file path that is generated so it works no matter if a build is triggered from the root project or the subproject.

## References

References #786.

## Changes

- Adjust the generated log file path so that it is absolute, rather than relative, which causes issues when the build is not triggered from the specific subproject the plugin is used in.

## Background Context

Please see issue #786 for an explanation of the project.  This is caused by the following:

The path to a `logFile` is correctly generated by:
https://github.com/akka/akka-grpc/blob/c1eb6fa184d45c289ade3befde9c0551597b6542/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L30

However, the offending line is the following, which generates a relative path _from the active project directory (e.g., the root project directory)_ .
https://github.com/akka/akka-grpc/blob/c1eb6fa184d45c289ade3befde9c0551597b6542/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L99

Consider that in a multi-build Gradle project, the `build` folder is located at `rootProject/subProject/build`.  The log file then has the path of `rootProject/subProject/build/file.log`  By relativizing the log file path with the _subproject path_, the log file is reduced to have the path `build/file.log`.

Thus, when the build is executed from the root project, the file is written to `rootProject/file.log` which does not exist.

The suggested fix is to generate an absolute path so that it works no matter where the build is triggered from.

## Testing

Please note that I **did not** test this on a Windows machine, which was the reason that this file was originally changed and this regression popped up.

This fix was tested on both Linux and Mac (and on a number of *nix Jenkins machines).

However, I would humbly request that someone try this on their Windows machine to make sure that Java's [Path.AbsolutePath](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#toAbsolutePath) does not require any special escaping for Windows.  If someone is 100% sure that the above method works out of the box, then I presume that there shouldn't be any issues, but I personally am not sure how Java renders Windows paths.

## Credit

Credit to @hithran for the original fix to this file which I'm patching (hopefully this patch does not conflict with the Windows compatibility he built-in).

### DIFF
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -96,7 +96,7 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
                                 option "server_power_apis=${extension.serverPowerApis}"
                                 option "use_play_actions=${extension.usePlayActions}"
                                 option "extra_generators=${extension.extraGenerators.join(';')}"
-                                option "logfile=${project.projectDir.toPath().relativize(logFile).toString()}"
+                                option "logfile=${logFile.toAbsolutePath().toString()}"
                                 if (extension.generatePlay) {
                                     option "generate_play=true"
                                 }


raboof:
OK TO TEST

raboof:
Thanks! It'd indeed be great if someone could check this still works on Windows and report back ;)

orendain:
Decided to give this a go on my Windows machine (😓) and the fix is not compatible (as is) with Windows.  I'll take another look at this later tonight and report back with a fix.

hithran:
I can explain why the fix won't work on Windows. As soon as there is a colon(:) in the path, the log file path breaks. And as the absolute path on Windows will contain the drive letter (C:/D:) it won't work. That is why I chose to use a relative path, to avoid the drive letter. 
Did you find a solution?

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
```diff
@@ -96,7 +96,7 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
                                 option "server_power_apis=${extension.serverPowerApis}"
                                 option "use_play_actions=${extension.usePlayActions}"
                                 option "extra_generators=${extension.extraGenerators.join(';')}"
-                                option "logfile=${project.projectDir.toPath().relativize(logFile).toString()}"
+                                option "logfile=${logFile.toAbsolutePath().toString()}"
```
hithran:
```suggestion
                                option "logfile=${project.rootProject.projectDir.toPath().relativize(logFile).toString()}"
```
Maybe inserting rootProject will resolve the problem? rootProject will return the current project when it is a single project, and will return the root project when this is a sub project.
I have not tested it yet, but looking at the code and options available on project, this would be my first guess

hithran: ok never mind that does not do what I hoped it would do

hithran: Ok that did not do what I expected it would do, now it works when the task is run from the main project, but no longer works when it is run from the subproject.
But it is something, it generates the correct path, now only need to find out which should be used when...

raboof:
Closing this for now due to inactivity - it would be great if someone were to pick this back up, though!

PR #788, opened 2020-01-13
Update the build

raboof:
Combination of a #765 and #776 to fix build problems in one go


project/ReflectiveCodeGen.scala
```diff
@@ -20,6 +20,9 @@ object ReflectiveCodeGen extends AutoPlugin {
   val extraGenerators = SettingKey[Seq[String]]("reflectiveGrpcExtraGenerators")
   val codeGeneratorSettings = settingKey[Seq[String]]("Code generator settings")
 
+  // needed to be able to override the PB.generate task reliably
+  override def requires = ProtocPlugin
```
renatocaval: I have tried this and the above change but was missing `PB.protoSources.value` below. 

I confess that this is all goes above my head.

PR #798, opened 2020-01-28
Avoid deprecated UseHttp API

raboof:

PR #799, opened 2020-01-28
Fatal Scala warnings

raboof:
Includes #798


runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -20,7 +20,7 @@ object ServiceHandler {
   private val notFound = CompletableFuture.completedFuture(HttpResponse.create().withStatus(StatusCodes.NOT_FOUND))
 
   /**
-   * Creates a [[HttpRequest]] to [[HttpResponse]] handler that can be used in for example
+   * Creates a [HttpRequest] to [HttpResponse] handler that can be used in for example
```
ennru: Did you intend to use just back ticks? Alternatively links should use the FQCN.

Issue #809, opened 2020-02-05
Client-side load balancing with Akka Discovery


raboof:
Client-side loadbalancing with 'static' (or 'grpc-dns') discovery mechanisms is already supported.

There are 2 ways we could implement client-side load balancing with other discovery mechanisms:

* Plug in service discovery into the grpc-java client through the [NameResolver](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/NameResolver.java) interface and use that.
* Support HTTP2 and loadbalancing in the akka-http client and switch to that.

ennru:
Relates to https://github.com/akka/akka-http/issues/2828

raboof:
while there can be more to gRPC load balancing (see for example https://grpc.io/blog/grpc-load-balancing/), #811 implements the basic functionality. Let's close this issue and open more specific issues when we want to implement further features.

PR #811, opened 2020-02-06
Plug Akka Discovery into the grpc-java NameResolver

raboof:
Instead of performing discovery 'outside' of the Channel.

This would allow configuring client-side loadbalancing, #809, but I have
not yet tested that that goal is actually achieved.

Before un-drafting this generally needs some more testing whether it
does the right thing, especially re-resolving names in case of updates
etc. It may also mean we can simplify/remove the recreation code in
ClientState.scala, but that also still needs to be verified.

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
@@ -23,6 +23,7 @@ import org.scalatest.BeforeAndAfterAll
 
 import example.myapp.helloworld.grpc.helloworld._
 import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.Span
 
 class CountingGreeterServiceImpl extends GreeterService {
   var greetings = new AtomicInteger(0);
@@ -41,10 +42,34 @@ class CountingGreeterServiceImpl extends GreeterService {
 
 }
 
+final class MutableServiceDiscovery(targets: List[Http.ServerBinding]) extends ServiceDiscovery {
+  var services: Future[Resolved] = _
+
+  setServices(targets)
+
+  def setServices(targets: List[Http.ServerBinding]): Unit =
+    services = Future.successful(
+      Resolved(
+        "greeter",
+        targets.map(
+          target =>
+            ResolvedTarget(
+              target.localAddress.getHostString,
+              Some(target.localAddress.getPort),
+              Some(target.localAddress.getAddress)))))
+
+  override def lookup(query: Lookup, resolveTimeout: FiniteDuration): Future[Resolved] = {
+    require(query.serviceName == "greeter")
+    services
+  }
+}
+
 class LoadBalancingIntegrationSpec extends AnyWordSpec with Matchers with BeforeAndAfterAll with ScalaFutures {
   implicit val system = ActorSystem("LoadBalancingIntegrationSpec")
   implicit val mat = akka.stream.ActorMaterializer.create(system)
   implicit val ec = system.dispatcher
+
+  override implicit val patienceConfig = PatienceConfig(5.seconds, Span(10, org.scalatest.time.Millis))
 
   "Client-side loadbalancing" should {
     "send requests to multiple endpoints" in {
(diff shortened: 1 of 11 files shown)


raboof:
This is now ready for review, though I'd like to double-check if it ever makes sense to explicitly pass in `DnsNameResolverProvider`

runtime/src/test/scala/akka/grpc/internal/AkkaDiscoveryNameResolverProviderSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.testkit.TestKit
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class AkkaDiscoveryNameResolverProviderSpec extends TestKit(ActorSystem()) with AnyWordSpecLike with Matchers {
+  "AkkaDiscoveryNameResolverProviderSpec" should {
+    "provide a NameResolver that uses the supplied serviceName" in {
+//      AkkaDiscoveryNameResolver("")
+    }
+  }
```
ignasi35: is this WIP or an oversight?

runtime/src/test/scala/akka/grpc/internal/NettyClientUtilsSpec.scala
```diff
@@ -26,12 +26,13 @@ class NettyClientUtilsSpec extends AnyWordSpec with Matchers with ScalaFutures w
...
-      channel.failed.futureValue shouldBe a[AskTimeoutException]
-    }
+//    The channel can now retry service discovery as needed itself,
+//    I guess we should test that instead?
+//    "fail to create a channel when service discovery times out" in {
+//      val settings = GrpcClientSettings.usingServiceDiscovery("testService")
+//
+//      val channel = NettyClientUtils.createChannel(settings)
+//    }
   }
```
ignasi35: With the changes in `NettyClientUtils` the [invocation](https://github.com/akka/akka-grpc/pull/811/files#diff-977da58f66c832d603f6ad9ac553d99dR71 to `val channel = builder.build()` is not longer within a `try{}catch` (or a `future.map{}` as it used to be). Therefore an invocation to `createChannel()` will either return a successful `Future` or throw a runtime exception but never a `Future.failed`?

runtime/src/main/scala/akka/grpc/internal/AkkaDiscoveryNameResolver.scala
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  override def refresh(): Unit =
+    listener.future.onComplete {
+      case Success(l) => lookup(l)
+      case Failure(_) => // We never fail this promise
+    }
+
+  def lookup(listener: Listener): Unit = {
+    discovery.lookup(Lookup(authority, portName, protocol), resolveTimeout).onComplete {
```
ignasi35: I think the name of the `authority` field is confusing. Here, for example, it's used as a service name on a service discovery lookup, but on line 28 it's used as an actual `authority`. Instead, we may need to pass a `serviceName` for the lookups and the value of `GrpcSettings#overrideAuthority` (or `serviceName` if unset).

This change here is related to the line `.forTarget("//" + settings.serviceName)` above.

interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
```diff
@@ -0,0 +1,141 @@
...
+      server1.unbind().futureValue
+      server1.terminate(hardDeadline = 100.milliseconds).futureValue
+      service1materializer.shutdown()
+      Thread.sleep(100)
+
+      for (i <- 1 to requestsPerServer) {
+        client.sayHello(HelloRequest(s"Hello $i")).futureValue
+      }
+
+      service1.greetings.get + service2.greetings.get should be(2 * requestsPerServer)
```
ignasi35: (-:

runtime/src/test/scala/akka/grpc/internal/NettyClientUtilsSpec.scala
raboof: I didn't mean for that code to ever throw runtime exceptions. I kept the `Future` in the API for now to not change too much at once.

runtime/src/main/scala/akka/grpc/internal/AkkaDiscoveryNameResolver.scala
raboof: I agree this is somewhat confusing, but I'm following the lead of the upstream impl here (https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L174)

runtime/src/test/scala/akka/grpc/internal/AkkaDiscoveryNameResolverProviderSpec.scala
raboof: good one, added

runtime/src/main/scala/akka/grpc/internal/AkkaDiscoveryNameResolver.scala
ignasi35: Makes sense, but that implementation is only valid for DNS `A` lookups (it's a `DnsNameResolver`) and in that case the name and the authority will likely match. `AkkaDiscoveryNameResolver` may use any discovery the user wants to inject. 

And it doesn't matter anyway because we're using:

```
    builder = settings.overrideAuthority.map(builder.overrideAuthority(_)).getOrElse(builder)
```

in `NettyClientUtils`. :-)

PR #815, opened 2020-02-11
Docs: add project info

ennru:
Use the Paradox project-info plugin to render standardised information.

bild

### DIFF
--- a/project/project-info.conf
+++ b/project/project-info.conf
@@ -1,2 +1,42 @@
 project-info {
+  version: "current"
+  shared-info {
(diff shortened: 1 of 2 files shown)


project/project-info.conf
```diff
@@ -1,2 +1,42 @@
 project-info {
+  version: "current"
+  shared-info {
+    jdk-versions: ["Adopt OpenJDK 8", "Adopt OpenJDK 11"]
+    // snapshots: { }
```
raboof: We do publish snapshots to https://bintray.com/akka/maven/akka-grpc

Issue #701, opened 2019-10-09
Document how to fail a request


raboof:
With `Source.failed`/`Future.failed` and `GrpcServiceException`, like in:

```
import scala.concurrent.Future

import io.grpc.Status

import io.grpc.health.v1._

import akka.NotUsed
import akka.grpc.GrpcServiceException
import akka.stream.scaladsl.Source

class HealthServiceImpl(services: Seq[String]) extends Health {
  override def check(in: HealthCheckRequest): Future[HealthCheckResponse] =
    if (services.contains(in.service))
      Future.successful(HealthCheckResponse(HealthCheckResponse.ServingStatus.UNKNOWN))
    else
      Future.failed(new GrpcServiceException(Status.NOT_FOUND))

  override def watch(in: HealthCheckRequest): Source[HealthCheckResponse, NotUsed] =
    Source.failed(new GrpcServiceException(Status.UNIMPLEMENTED))
}
```

donpaul120:
Honestly this needs to be documented...  I struggled before i could find this.

PR #825, opened 2020-02-18
Document status codes

raboof:
Refs #701


interop-tests/src/test/java/example/myapp/helloworld/grpc/ExceptionGreeterServiceImpl.java
```diff
@@ -0,0 +1,49 @@
+package example.myapp.helloworld.grpc;
+
+import java.util.concurrent.CompletionStage;
+import java.util.concurrent.CompletableFuture;
+
+import io.grpc.Status;
```
ennru: Make the import show in the snippets? It's a very generic class name.

Issue #695, opened 2019-10-03
gRPC-Web Support


otto-dev:
### Short description

Let's add support for [gRPC-Web](https://github.com/grpc/grpc-web)

### Details

Maybe we can add a configuration or alternative implementation that is compatible with the specifications of [gRPC-Web](https://github.com/grpc/grpc-web), so that browsers can access the Akka HTTP endpoints without another HTTP proxy.

[The specification doesn't look too complicated](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md#protocol-differences-vs-grpc-over-http2). I hacked around with it a bit today, and it didn't take too much to be able to communicate via unary calls.

All I had to do for unary calls to work was:
- base64 encode request and response bodies
- set headers

Streaming seems to be a bit more complicated. But probably quite doable for someone familiar with the gRPC protocol. From the documentation, and what I saw seems to be some extra metadata appended to the response bodies. From memory it was something like this:
```
3grpc-status: 0
Date: <curent date/time>
```
The first bytes (`3`) appear to be the length of the following body in binary format. It talks in the specs about these key-value pairs in the response body. I didn't play around with that part any more, mainly because I don't know enough about the `gRPC` specs to know what's going on.

I will attach some very hacky wrapper that made it possible for me to access a gRPC service from the browser - except streaming. See the following comment.

otto-dev:
```scala
    // this is our gRPC service
    val service = SomeHandler(new SomeService())

    // we wrap our service to do some hacking.
    val handler: HttpRequest => Future[HttpResponse] = { request: HttpRequest =>
      val withoutEncoding = request.copy(
        // gRPC web doesn't play with gzip encoding. This is a hack to stop Akka HTTP from encoding the response
        headers = request.headers.filterNot(_.name == "grpc-accept-encoding"),

        // base64 body instead of plain bytes
        entity = request.entity.transformDataBytes(Flow[ByteString].map { v => ByteString(Base64.getDecoder.decode(v.toByteBuffer)) })
      )
      // now get the response of our service and modify it - then return it
      service(withoutEncoding).map { response =>
        // We need some special headers for gRPC web
        response.withHeaders(
          `Access-Control-Allow-Origin`.*,
          `Access-Control-Allow-Headers`("x-user-agent", "x-grpc-web", "content-type"),
          `Access-Control-Expose-Headers`("grpc-status", "grpc-message"),
        )
        // we need this content type
[...]

Issue #827, opened 2020-02-19
Akka gRPC 1.0.0 plans


ennru:
We expect to fix up a few things in Akka gRPC and go for 1.0.0 in Q2 2020.

We already consider Akka gRPC production-ready for use in projects: the implementation is well-tested and network API's (being defined by the gRPC protocol) are stable. The main relevance of 1.0.0 is that we start providing binary compatibility (#761), making Akka gRPC more suitable for use in libraries.

See [the milestone](https://github.com/akka/akka-grpc/milestone/13) for what is currently scheduled for 1.0.0

raboof:
See #1008 for progress

PR #829, opened 2020-02-20
Grpc-web support

timw:
This is a full implementation of the grpc-web protocol variant(s), supporting the `application/grpc-web` and `application/grpc-web` media types.
All current tests pass (testing only the grpc protocol) and [basic grpc-web compatibility tests](https://github.com/johanbrandhorst/grpc-web-compatibility-test) pass.

References #695 

At the moment I'm doing a PR for this to get discussion on the implementation approach and some of the remaining items.

Approach Taken:

This is an implementation of grpc-web (binary and text variants) as a grpc variant (i.e. a slightly different physical version of the same logical protocol) as opposed to a grpc proxy.
Essentially this involves abstracting the details of a `GrpcVariant` beyond the existing message encoding `Codec` and detecting/negotiating the `GrpcProtocolUnmarshaller` and `GrpcProtocolMarshaller` during the request/response. These are expressed in terms of a stream of logical gRPC frames (data or trailer) that are then mapped into a stream of (transfer encoding) chunks as appropriate to the protocol: grpc maps to Chunk and LastChunk, with the latter magicked into HTTP/2 trailers, and grpc-web maps both types to Chunks with a field type (and base64 encoding in the case of grpc-web-text).

Refs:
- [gRPC-Web spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md)
- [gRPC over HTTP/2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md)
- [Behaviour of Google grpc-web client in browsers](https://github.com/grpc/grpc-web/blob/master/BROWSER-FEATURES.md)

Discussion/interesting points:
- This implementation approach allows grpc and grpc-web to be served in the same handler (i.e. from the same port), and probably with a mix of HTTP/2 and HTTP/1.1
- The `ServiceHandler` functions to support grpc-web use [https://github.com/lomigmegard/akka-http-cors]. Maintainers may have some thoughts about whether they want to downstream that, or import/implement a minimal CORS implementation.
- The codegen needed to be updated to allow protocol detection and unmarshaller/marshallers to be used properly - principally the negotiated marshaller needs to be used for exception/status responses as well as the main path, and the existing marshaller code calculated the codec in the generated handler without passing on the request/media type.
- I've taken the approach of deprecating the marshall/unmarshall/exception handler functions that can't accomodate grpc-web, but left them in place to retain source/binary compatibility with existing codegen plugins and compiled handler code. I _think_ I've only broken signatures that are not used from generated code (e.g. `GrpcResponseHelpers`).
- It wouldn't take too much more work to support non `+proto` variants (e.g. `application/grpc-web+json` with the changes made. Would require a meaningful client implementation to make it worthwhile though, but grpc-web could get there soon.

TODOs:
- The generated Marshallers in the scala codegen have not been updated - not sure what these are for TBH 
- ServiceHandler in the javadsl is not updated to the new builder style that can serve grpc-web
- Docs
- Unit tests for new functions
- Interop/integration test - have proven interop with [https://github.com/johanbrandhorst/grpc-web-compatibility-test] and will probably add akka-grpc backend to that project if this is merged, but not sure on best way to include an in-tree integration test. Probably using node or Karma to drive grpc-web in the existing interop framework would be an approach.
- Most of the potential marshall/unmarshall paths are implemented, but reading grpc-web responses will require a multi-port shape on the stream to split off status trailers. In theory the AkkaGrpcClient could support grpc-web with minimal work.
- Someone who knows CORS better than me should probably review the default config to confirm its minimal and safe.
- As part of the rework, the response content types now adapt their compressibility depending on whether a compressed message encoding is used, which should allow HTTP level content compression to be used. This is notable as the Google grpc-web client only supports Identity encoding, as GZip in Javascript has poor performance, and gzip/deflate HTTP content compression is well supported. I haven't enabled this in the `ServiceHandler` functions at the moment, not being 100% sure what the best approach is.


lightbend-cla-validator:

raboof:
Hi Tim - thanks for the draft, this is very cool! Previously I was suspicious adding grpc-web support into akka-grpc itself would add too much complexity, in which case we would be better off recommending a dedicated proxy instead. This work however looks promising enough that we might be able to do it in akka-grpc after all.

As for timing, I hope to release akka-grpc 0.8.0 (#830) soon, and 1.0.0 (#827) within a couple of weeks. For 0.8.0 I hope to get #811 in, and the plan for 1.0.0 is/was to focus mainly on preparing the API's so we can guarantee binary compatibility from 1.0.0 forward.

I see you have been careful to avoid binary compatibility problems, deprecating methods instead. That is great and means we could introduce grpc-web support after 1.0.0. However, it would of course be nicer if we could get the PR in before 1.0.0, so we would not need those deprecations and can just modify the API's. I'd be happy to work with you to try and make that happen, but on the other hand I don't want to postpone the 1.0.0 release too much.

I'm still in the middle of some other work but looking forward to reviewing your approach in more detail soon!

lightbend-cla-validator:

raboof:
I've started looking into this PR a bit, and I think it's promising.

I think there's 2 main criteria based on which we should adopt this feature or not: 
- first, since gRPC-web is a relatively niche use case (and might eventually be replaced by whatwg-streams, https://github.com/whatwg/streams), we should not sacrifice performance of the native gRPC feature in order to add support for the gRPC-Web.
- second, it should not make the Akka gRPC codebase so much more complicated that it becomes hard to maintain.

Looking at this PR so far, I'm hopeful we can get this PR in a shape where it passes both those criteria: of course wrapping things in the `Frame` trait and discovering what protocol the connecting party wants to speak adds some overhead but that seems negligible. As far as maintenance goes, it looks like we could fit this in in a way that keeps things managable. I'll probably have a lot review comments nitpicking about naming and code structure, though ;).

I'll try to do a perhaps somewhat superficial pass today, and get back in more detail somewhere next week.

runtime/src/main/scala/akka/grpc/GrpcWeb.scala
```diff
@@ -0,0 +1,162 @@
...
+import akka.http.scaladsl.model.headers.{ `Accept-Encoding`, `Content-Encoding`, `Content-Type`, Accept }
+import akka.stream.scaladsl.Flow
+import akka.util.ByteString
+import ch.megard.akka.http.cors.scaladsl.model.HttpHeaderRange
+import ch.megard.akka.http.cors.scaladsl.settings.CorsSettings
+import io.grpc.{ Status, StatusException }
+
+import scala.collection.immutable
+
+trait GrpcWebVariant extends GrpcVariant {
```
raboof: I wonder if the specific variant(/protocol)s should be in `akka.grpc.internal` - are they ever part of a user-facing API?

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -7,10 +7,12 @@
 @akka.grpc.gen.Constants.DoNotEditComment
 package @service.packageName
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.{ ExecutionContext, Future }
```
raboof: We avoided importing `Future` here and instead spelled it out below to make sure the generated code also works when there is an object called `Future` in the proto definition.

runtime/src/main/scala/akka/grpc/Grpc.scala
```diff
@@ -5,85 +5,135 @@
...
-  val contentType = MediaType.applicationBinary("grpc+proto", MediaType.NotCompressible).toContentType
-
-  // Flag to signal the start of an uncompressed frame
-  val notCompressed = ByteString(0)
-  // Flag to signal the start of an frame compressed according to the Message-Encoding from the header
-  val compressed = ByteString(1)
+/**
+ * Implementation of the gRPC protocol.
+ */
+object Grpc extends GrpcVariant {
```
raboof: It was initially unclear to me that this class is intended to contain logic that is specific to the native gRPC protocol, and not things that are generic for across native grpc, grpc-web and grpc-web-text. Perhaps have 'Native' in the name?

runtime/src/main/scala/akka/grpc/Codecs.scala
```diff
@@ -4,23 +4,52 @@
...
+   * @return a codec to compress data frame bodies with, which will be [[Identity]] unless the client specifies support for another supported encoding.
+   */
   def negotiate(request: jm.HttpRequest): Codec =
-    `Message-Accept-Encoding`
-      .findIn(request.getHeaders)
-      .intersect(supported)
-      .headOption
-      .map(byName(_))
-      .getOrElse(Identity)
+    `Message-Accept-Encoding`.findIn(request.getHeaders).headOption.map(byName(_)).getOrElse(Identity)
```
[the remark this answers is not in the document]
timw: I agree - I think I got confused here. Will fix with unit test.

runtime/src/main/scala/akka/grpc/Codecs.scala
```diff
@@ -4,23 +4,52 @@
...
 import akka.http.javadsl.{ model => jm }
+import io.grpc.Status
 
 import scala.collection.immutable
+import scala.util.{ Failure, Success, Try }
 
 object Codecs {
   // TODO should this list be made user-extensible?
-  val supportedCodecs = immutable.Seq(Gzip)
+  val supportedCodecs = immutable.Seq(Gzip, Identity)
```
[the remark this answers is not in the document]
timw: It's primarily to allow a client to specify `grpc-encoding: identity` in the detect path, which is (at least implicitly what the browser clients do, as gzip has poor performance in JS).

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  /** A trailer (status headers) frame in a gRPC protocol stream */
+  case class TrailerFrame(trailers: List[HttpHeader]) extends Frame
+
+  /**
+   * Implements the encoding of a stream of gRPC Frames into a physical/transport layer.
+   *
+   * This maps the logical gRPC frames into a stream of chunks that can be handled by the HTTP/2 or HTTP/1.1 transport layer.
+   */
+  case class GrpcProtocolMarshaller(
```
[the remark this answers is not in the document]
timw: I thought this would be contentious (and didn't 100% like Marshaller myself, since that's quite overloaded in the various gRPC frameworks already).

This type is more than just the frame encoder though (that's 2 of the attributes on the class).
The others are the negotiated content type, and the negotiated message encoding.

runtime/src/main/scala/akka/grpc/GrpcWeb.scala
timw: I had toyed with them being optionally enabled in the service handler builder, but in the end kept is simple. Only the default CORS settings need a home in the user-facing API, and they could live in the service handler.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
timw: My bad - got a bit carried away optimising imports, and should have left the patch minimal. Will revert those bits.

runtime/src/main/scala/akka/grpc/Grpc.scala
timw: I mostly agree - in this case grpc defines most of the protocol used by grpc-web variants.
Could possibly split the shared/common stuff out into a separate object and have the variants only have the specifics for their concrete protocols.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -0,0 +1,117 @@
...
+import akka.http.javadsl.{ model => jmodel }
+import akka.http.scaladsl.model.{ ContentType, HttpHeader }
+import akka.http.scaladsl.model.HttpEntity.ChunkStreamPart
+import akka.stream.scaladsl.Flow
+import akka.util.ByteString
+
+import scala.util.Try
+
+/**
+ * A variant of the gRPC protocol - e.g. gRPC and gRPC-Web
```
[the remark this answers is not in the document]
timw: I flipflopped a bit here as the structure evolved, but I think that could work well.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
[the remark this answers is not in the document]
timw: Maybe `GrpcProtocolReader` and `GrpcProtocolWriter` (reader would be used to read requests in the server handler, and responses in the client etc.)

lightbend-cla-validator:

runtime/src/main/scala/akka/grpc/Codecs.scala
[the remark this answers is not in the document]
timw: Have fixed this and added a thorough spec.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
timw: Have reverted this change.

timw:
I've added some minor fixes from review comments, and resolved merge conflicts with master - the latter contains more changes than I'd like, but the introduction of `GrpcEntityHelpers` created some issues with my changes and I ended up having to fix the implicits in the generated marshallers and introduced a `GrpcRequestHelpers` to tidy up the tests a bit.

lightbend-cla-validator:

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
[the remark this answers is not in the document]
raboof: I think it would be fine to have the content type and message encoding as properties of the encoding, but I'm OK with reader/writer as well.

runtime/src/main/scala/akka/grpc/Grpc.scala
raboof: > Could possibly split the shared/common stuff out into a separate object and have the variants only have the specifics for their concrete protocols

Yes, I think it would be helpful to have shared/common stuff in `Grpc` and the parts that are specific for the native protocol in something like `GrpcProtocolNative` (next to `GrpcProtocolWeb` etc).

raboof:
OK TO TEST

lightbend-cla-validator:

runtime/src/main/scala/akka/grpc/Grpc.scala
timw: I've separated out the common parts as much as sensible (I hope) and moved the protocol definitions into .internal.

runtime/src/main/scala/akka/grpc/GrpcWeb.scala
timw: Have moved these into .internal - there's no reason for them to be in .grpc at present I think.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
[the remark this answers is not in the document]
timw: Have renamed as discussed in this and other comments to GrpcProtocol*

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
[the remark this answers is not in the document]
timw: Have renamed to reader/writer.

timw:
> Nice improvements. I don't think I'm a fan of the new builder API in `ServiceHandler` yet, so that might be something to iterate on a bit.

I'm not attached to it - I think I envisaged the config being more complex than it ended up being, so it's overkill.
We could probably achieve the same result with a method (or ServiceHandler apply for the Scala dsl) with all named/default args, along the lines of `HttpMesage` etc. Can do that if it sounds like a better approach to you as well...

raboof:
> We could probably achieve the same result with a method (or ServiceHandler apply for the Scala dsl) with all named/default args, along the lines of HttpMessage etc

Let's think a bit about how we expect this functionality to be used, that might help in designing the API. Do we think people will typically serve both native gRPC and gRPC-web on the same port? And only apply CORS to gRPC-web? Would anyone ever want to do gRPC-web without CORS (having that handled by some frontend proxy perhaps)?

timw:
CORS for gRPC probably isn't a thing - CORS is primarily used in web browsers, and gRPC/HTTP2 can't be implemented with current browser APIs.
gRPC-Web without CORs would imply a non-browser client, which is quite feasible (and could actually be quite useful at the moment with AWS not supporting gRPC/HTTP2 in load balancers). I don't know of a non JS gRPC-Web client though, so we'd only be supporting possible usage of the existing JS/TS clients through node.js etc.

Serving through a single port does simplify our deployment a bit (one frontend LB for web assets and gRPC-Web) so it's something we're playing with. Not a catastrophe if that doesn't work, but I don't see any reason to prevent it (even if it's not something we want to bake in) - i.e. if all we support is construction of partials for gRPC and gRPC-Web services people can construct their top level handler as they want.

The existing generated handlers (the concrete handlers) are a bit interesting - they don't provide a 415 response on unknown media type, and there are a lot of overloads (at least in the Scala versions) that could be avoided. 

I think it would also be great if whatever we do supports automagic enablement of the reflection service as well - that would require a combination of the ServiceDescription and a handler to be provided.

Is there any appetite to change the generated Handler APIs for 1.0 to provide a combined `ServiceDescription` + handler type from `.partial`?

timw:
A case class companion to the existing `ServiceHandler` object works quite well.

```
case class ServiceHandler(
    description: ServiceDescription,
    handler: PartialFunction[HttpRequest, Future[HttpResponse]])
```

raboof:
> Serving through a single port does simplify our deployment a bit (one frontend LB for web assets and gRPC-Web) so it's something we're playing with. Not a catastrophe if that doesn't work, but I don't see any reason to prevent it (even if it's not something we want to bake in) - i.e. if all we support is construction of partials for gRPC and gRPC-Web services people can construct their top level handler as they want.

Yes, that sounds sensible: the 'typical use' would be to have native gRPC on one port and gRPC-Web with CORS on another (possibly even a different list of services, or somewhat different implementations).

I see you already plugin-tester-scala/src/main/scala/example/myapp/CombinedServer.scala - could you extend that one to implement this scenario, so also bind a gRPC-Web service (to a different port)? 

> there are a lot of overloads (at least in the Scala versions) that could be avoided.

Yes, perhaps we went a bit overboard with those in an effort to allow various customisations but not require too many parameters. We avoided default arguments here because they tend to be hard to evolve in a binary compatible way. However, thinking about this a bit more, I guess the handlers will be usually used directly and not via some library, so it might be fine to not promise binary compatibility for those.

> Is there any appetite to change the generated Handler APIs for 1.0 to provide a combined ServiceDescription + handler type from .partial?

I'd definitely like to experiment a bit with different options. To not let this PR get too big, however, perhaps we should keep the `ServiceHandler.concatOrNotFound` and add a similar function for gRPC-Web there, and postpone introducing the builder machinery (and other experiments) for another PR?

raboof:
> I see you already plugin-tester-scala/src/main/scala/example/myapp/CombinedServer.scala - could you extend that one to implement this scenario, so also bind a gRPC-Web service (to a different port)?

> perhaps we should keep the `ServiceHandler.concatOrNotFound` and add a similar function for gRPC-Web there, and postpone introducing the builder machinery (and other experiments) for another PR?

Also I'm afraid merging #129 caused some conflicts. I think this is pretty close to be in a mergable state. Do you want me to help out with those last bits or would you prefer to pick them up yourself?

raboof:
Also to make the build green the CLA needs to be signed (https://www.lightbend.com/contribute/cla) and some MiMa filters need to be added. I can take care of the latter if you want?

timw:
> Also I'm afraid merging #129 caused some conflicts. I think this is pretty close to be in a mergable state. Do you want me to help out with those last bits or would you prefer to pick them up yourself?

I've merged master again and resolved conflicts. `akka.grpc.internal.AkkaDiscoveryNameResolverSpec` will need a patch-up as there was a non-portable exception description test in there.

> Also to make the build green the CLA needs to be signed (https://www.lightbend.com/contribute/cla) and some MiMa filters need to be added. I can take care of the latter if you want?

I've signed the CLA. I can do if you tell me which buttons to whack - haven't done MiMa before. Otherwise happy for you to do it and I'll shadow your changes to learn it.

> I'd definitely like to experiment a bit with different options. To not let this PR get too big, however, perhaps we should keep the ServiceHandler.concatOrNotFound and add a similar function for gRPC-Web there, and postpone introducing the builder machinery (and other experiments) for another PR?

Have backed `ServiceHandler` off to a more basic state - I agree we can work on that later.
The Java DSL was giving me conniptions - took a while to get my head around it, but have it sorted now I think - will push update soon.

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
```diff
@@ -4,26 +4,146 @@
...
+     */
+    def withCorsSettings(settings: CorsSettings): this.type = {
+      corsSettings = settings
+      this
+    }
+
+    /**
+     * Produces a partial function that will service the provided gRPC and gRPC-web services:
+     *   - `application/grpc` requests will be served by the specified [[withGrpc gRPC services]].
+     *   - `application/grpc` and `application-grpc-web` requests will be served by the specified [[withGrpcWeb gRPC-web services]].
```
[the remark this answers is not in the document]
timw: This has been backed out, but thanks for the catch.

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
```diff
@@ -4,26 +4,146 @@
...
+      } else None, if (grpcWebServices.nonEmpty) {
+        val servicesHandler = concat(grpcWebServices: _*)
+        // HACK: Cors is only exposed as a Directive, and Route.asyncHandler produces a sealed route
+        //       so make sure that handler is only entered if the underlying grpc-web services would serve the request
+        val grpcWebHandler = Route.asyncHandler(cors(corsSettings) { handleWith(servicesHandler) })
+        Some({
+          case r if (isGrpcWebRequest(r) || isCorsPreflightRequest(r)) && servicesHandler.isDefinedAt(r) =>
+            grpcWebHandler(r)
+        }: PartialFunction[HttpRequest, Future[HttpResponse]])
+      } else None).flatten: _*)
```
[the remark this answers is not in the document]
timw: This has been backed out.

project/Dependencies.scala
```diff
@@ -81,6 +82,7 @@ object Dependencies {
         Compile.akkaHttp,
         Compile.akkaHttp2Support,
         Compile.akkaDiscovery,
+        Compile.akkaHttpCors,
```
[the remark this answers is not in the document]
timw: I'll defer to the maintainers on this one - the CORS dependency feels like it should be optional, so I would tend to agree with making it `provided`.

raboof:
> I've merged master again and resolved conflicts. 

Have you pushed that yet?

lightbend-cla-validator:

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -0,0 +1,112 @@
...
+   */
+  def newReader(codec: Codec): GrpcProtocolReader
+}
+
+/**
+ * Core definitions for gRPC protocols.
+ */
+object GrpcProtocol {
+
+  private val grpcVariants: Seq[GrpcProtocol] = Seq(GrpcProtocolNative, GrpcProtocolWeb, GrpcWebTextProtocol)
```
ennru: The protocol classes could have more consistent names (eg. `GrpcNativeProtocol`, `GrpcWebProtocol`)

runtime/src/main/scala/akka/grpc/internal/Grpc.scala
```diff
@@ -0,0 +1,109 @@
...
+import akka.stream.scaladsl.Flow
+import akka.util.ByteString
+import akka.NotUsed
+import io.grpc.{ Status, StatusException }
+
+/**
+ * Functionality common to all gRPC protocols.
+ */
+@InternalApi
+object Grpc {
```
ennru: I dislike this object name as it is too generic. Should it be `GrpcProtocolHelpers`? It mostly is a companion to `AbstractGrpcProtocol`, right?

runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * The `application/grpc-web-text+proto` variant of gRPC.
+ *
+ * This is the same as `application/grpc-web+proto`, but with each chunk of the frame encoded gRPC data also base64 encoded.
+ */
+object GrpcWebTextProtocol extends GrpcWebVariant("grpc-web-text") {
+
+  override def postEncode(framed: ByteString): ByteString = ByteString(Base64.getEncoder.encode(framed.toByteBuffer))
+
+  override def preDecode(frame: ByteString): ByteString = ByteString(Base64.getDecoder.decode(frame.toByteBuffer))
```
ennru: Soon available natively on ByteString! See https://github.com/akka/akka/pull/28663

viktorklang: @ennru And much more efficient than those intermediate copies. :)

project/Dependencies.scala
```diff
@@ -81,6 +83,7 @@ object Dependencies {
         Compile.akkaHttp,
         Compile.akkaHttp2Support,
         Compile.akkaDiscovery,
+        Compile.akkaHttpCors % "provided",
```
raboof: Hmm I'm not sure if this'll work, since the ServiceHandler class is used for native gRPC as well and might not be loadable if it refers to classes that are not on the classpath. Perhaps neat to move that to its own class anyway?

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
timw: Yeah, apologies for that - IntelliJ kept breaking during rename refactors and I missed that.

project/Dependencies.scala
timw: Has been split in your latest commits.

runtime/src/main/scala/akka/grpc/internal/Grpc.scala
timw: It is effectively the companion to `AbstractGrpcProtocol` yes. Happy to make it formal if you concur.

ennru: Yes, that would be great.

raboof: I think this is the last bit, so I'll pick it up and merge! Almost there!

PR #834, opened 2020-02-21
Add server reflection to 'combined' test server

raboof:


raboof:
> Should the docs say something about the `reflectionService` here?

I don't think so - it would distract on the 'walkthrough' page, and the index on the left already links to https://doc.akka.io/docs/akka-grpc/snapshot/server/reflection.html

Issue #577, opened 2019-05-08
Support for richer error responses / google.rpc.Status


ctoomey:
Currently there's no standard way to return rich error responses, just the code and optional description of `io.grpc.Statu` via `GrpcServiceException`.

Google has introduced and uses [`google.rpc.Status`](https://cloud.google.com/apis/design/errors#error_model) and is developing client library support for it, per [here](https://stackoverflow.com/questions/48748745/pattern-for-rich-error-handling-in-grpc).

Richer error responses are something our team would like to be able to support in a standard way. Is adding support for this on the akka-grpc roadmap?

raboof:
This sounds great!

rockywarren-vertex:
@raboof Is this on the roadmap? It's preventing us from moving to `akka-grpc` since we utilize rich errors so they're machine-readable by clients.

raboof:
Hi @rockywarren-vertex! thanks for expressing your interest in this feature, that is useful to prioritize it.

We're currently working towards 1.0.0 (#827) and we're not planning to work on this feature for that release. That said, it does seem like a very useful feature, so a contribution in this area would definitely be welcome.

If I understand correctly, those detailed error responses are encoded as custom metadata fields in the trailing headers.

One simple thing we could do is change the error handler from `Throwable => Status` to `Throwable => List[HttpHeader]`, that way you could implement detailed error responses in your own error handler.

When we want to support this more natively we can introduce some new exception types, and extend the default error handler to turn those into the standard rich error types. Since there's already code generation in the akka-grpc-runtime subproject that might not be hard anymore.

drmontgomery:
Hi, all. I just submitted a pull request that implements support for returning custom headers from GrpcExceptionHandler (and adds them to GrpcServiceException as well).

I tried to keep the changes fairly minimal. Unfortunately, there was a fair amount of incidental complexity due to the need to support both javadsl and scaladsl HttpHeaders in the interfaces.

Things that I think could be improved:
1. It feels like returning HttpHeaders is a little low-level. It also means that user code is going to have to duplicate the Base64 encoding functionality to send binary data to clients. Perhaps we should be returning a Map of String and Binary metadata entries instead, and handling the encoding internally?
[...]

PR #838, opened 2020-02-25
Adding support for custom headers #577

drmontgomery:
References #577

Hi there!

I took a stab at adding support for returning custom HttpHeaders from a GrpcExceptionHandler.

I modified the GrpcExceptionHandler to return a case class (GrpcErrorResponse) that pairs an io.grpc.Status with an optional Seq of headers. These headers are then added to the trailing chunk by GrpcResponseHelpers.

Most of the complexity stems from needing to provide Java and Scala versions. I tried to follow the existing patterns, but I haven't done a lot of combined Java/Scala library development, so it's entirely possible I've missed something.

I also added HttpHeaders to GrpcServiceException. Unfortunately, because HttpHeader is defined in a language-specific DSL for akka.http, this required moving GrpcServiceException from akka.grpc into the language-specific DSLs as well. That feels like a pretty big API change for feature like this. Is there a better way to handle this?

Anyway, definitely open to feedback, but wanted to try to contribute to such a useful project.

Thanks,
-David


lightbend-cla-validator:
Hi @drmontgomery,

raboof:
OK TO TEST

raboof:
> I took a stab at adding support for returning custom HttpHeaders from a GrpcExceptionHandler. definitely open to feedback, but wanted to try to contribute to such a useful project.

Thanks a lot! 

> I modified the GrpcExceptionHandler to return a case class (GrpcErrorResponse) that pairs an io.grpc.Status with an optional Seq of headers. These headers are then added to the trailing chunk by GrpcResponseHelpers.

I'm on the fence here. The alternative would be to have the exception handler to return a collection of headers. Since the status also turns into a header, this would avoid introducing the `GrpcErrorResponse` concept. On the other hand, introducing `GrpcErrorResponse` makes it more clear that the status is required. Since bringing in custom error handling is a bit of a poweruser feature, I'm not sure what's best.

> I also added HttpHeaders to GrpcServiceException. Unfortunately, because HttpHeader is defined in a language-specific DSL for akka.http, this required moving GrpcServiceException from akka.grpc into the language-specific DSLs as well. Is there a better way to handle this?

The alternative would be to keep GrpcServiceException language-agnostic, by giving it both Scala and Java API's. That is possible because of 2 facts: 1) the scaladsl HttpHeader extends from the javadsl HttpHeader (so each scaladsl HttpHeader can be safely used as javadsl header) and 2) it is impossible to construct a javadsl HttpHeader that is not also a scaladsl HttpHeader. While '2' is not something we want to expose to the end-user API, it does mean we can convert between the two in our implementation.

> That feels like a pretty big API change for feature like this.

I'm not so concerned about the size of the API change (we are pre-1.0.0, after all), I think we should choose the best solution. I'm leaning towards keeping one `GrpcServiceException` and giving it both a Java and a Scala API. What do you think?

drmontgomery:
Thanks for the quick feedback!

Re GrpcErrorResponse:
I agree that this is a bit of a poweruser feature, but returning a collection of headers feels a bit too low-level to me, for two reasons. First, it requires the library user to understand and reimplement the gRPC status encoding logic, which seems bad in terms of code duplication and overall library ergonomics. Second, it provides a mechanism by which the user could accidentally produce responses that don't follow the gRPC-HTTP2 spec (i.e. if they accidentally omit the grpc-status header), which seems like a core library responsibility.

I actually wonder if returning HttpHeaders at all is too low-level. From the perspective of a gRPC user, everything is metadata, not http headers. Using HttpHeaders requires them to understand the details of binary metadata encoding, etc. Should the exception handler be returning a status and a collection of (String, MetadataEntry) tuples instead?

Re GrpcServiceException:
I'm not sure I completely agree with the assertion that it is impossible to construct a javadsl HttpHeader that is not also a scaladsl HttpHeader.

The javadsl HttpHeader is an abstract class with no base class, which scaladsl HttpHeader subclasses. In practice, most of the useful javadsl header classes (like CustomHeader or RawHeader) derive from the scaladsl HttpHeader, but I don't see anything (like a DoNotInherit annotation) to prevent users from creating their own javadsl HttpHeader subclasses with no relationship to the scaladsl HttpHeader.  I tried to take advantage of this in the GrpcExceptionHelper.asScala method, by attempting to cast the incoming javadsl HttpHeader as a scaladsl HttpHeader, and falling back to wrapping it in a RawHeader, but I don't think the cast is guaranteed to succeed.

When you envision the language-agnostic GrpcServiceException, are you thinking of something like the following?
class GrpcServiceException(val status: Status, val headers: Seq[sHttpHeader] = Nil) {
  def this(status: Status, headers: jIterable[jHttpHeader]) {
    this(status, headers.asScala.map(asScala).toSeq)
  }
}

Looking a bit more closely at Akka HTTP, it looks like the standard pattern for Java/Scala interop is to define a Java interface with a Scala implementation that derives it, and then use factory methods in the javadsl to create instances from the scaladsl. I'm violating that pattern with this change, since there is no relationship between the two GrpcServiceExceptions (or GrpcErrorResponse) for that matter. Would it make sense to move in that direction? The main quirk from a user perspective is that we'd have to disallow calling the constructor directly for javadsl classes, but it would simplify the internals slightly.

drmontgomery:
I went ahead and made GrpcServiceException language-agnostic. Please take a look and confirm that my change matches your intent.

raboof:
# GrpcErrorResponse

> returning a collection of headers feels a bit too low-level to me. it requires the library user to understand and reimplement the gRPC status encoding logic, which seems bad in terms of code duplication and overall library ergonomics

I agree we should not force users to reimplement that encoding - instead we could provide convenient helpers that produce headers based on a `Status` and further custom headers.

> I actually wonder if returning HttpHeaders at all is too low-level. From the perspective of a gRPC user, everything is metadata, not http headers. Using HttpHeaders requires them to understand the details of binary metadata encoding, etc. Should the exception handler be returning a status and a collection of (String, MetadataEntry) tuples instead?

This could also largely be solved with helpers that take either `String` or `ByteString` data and produce the corresponding `HttpHeader`. I see the appeal of using the `MetadataEntry` concept since that is already introduced on the client side (though only in a 'corner' of the API there, as well...)

I'm not completely against introducing something like `GrpcErrorResponse(Status, Map[String, MetadataEntry])`, but having convenient helper methods that create plain `HttpHeader`s seems quite a bit simpler, and acceptable for an 'advanced' feature like this. I'm not sure the benefits of adding a `GrpcErrorResponse` runtime concept justify the added complexity.

# GrpcServiceException

> I don't see anything (like a DoNotInherit annotation) to prevent users from creating their own javadsl HttpHeader subclasses with no relationship to the scaladsl HttpHeader

It is suggested in the javadoc ("All actual header values will be instances of one of the subtypes defined in the `headers` packages. Unknown headers will be subtypes of `RawHeader`."), though I agree it would be helpful to clarify. Added https://github.com/akka/akka-http/pull/2982.

> I went ahead and made GrpcServiceException language-agnostic. Please take a look and confirm that my change matches your intent.

Yes, I think this looks good.

interop-tests/src/test/scala/akka/grpc/scaladsl/GrpcExceptionHandlerSpec.scala
```diff
@@ -114,6 +137,8 @@ class GrpcExceptionHandlerSpec
       statusHeader.map(_.value()) should be(Some("3"))
       val statusMessageHeader = lastChunk.trailer.find { _.name == "grpc-message" }
       statusMessageHeader.map(_.value()) should be(Some("No name found"))
+      val richErrorHeader = lastChunk.trailer.find { _.name == RichErrorHeader.name }
```
raboof: I think if you make it a `case class` you could even:

```
val richErrorHeaderValue = lastChunk.trailer.collectFirst { case RichErrorHeader(value) => value }
richErrorHeaderValue should be(Some("test-data"))
```
Haven't tested though.

In any case, nice test!

runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
```diff
@@ -4,8 +4,38 @@
...
 
+import java.lang.{ Iterable => jIterable }
+
+import akka.grpc.internal.GrpcExceptionHelper
+import akka.http.javadsl.model.{ HttpHeader => jHttpHeader }
+import akka.http.scaladsl.model.HttpHeader
 import io.grpc.Status
 
-class GrpcServiceException(val status: Status) extends RuntimeException(status.getDescription) {
+class GrpcServiceException(val status: Status, val headers: Seq[HttpHeader])
```
raboof: Seq is immutable or mutable depending on whether you're on Scala 2.12 or 2.13. It might be nicer to explicitly use `immutable.Seq` or perhaps go for `List`.

runtime/src/main/scala/akka/grpc/internal/GrpcEntityHelpers.scala
```diff
@@ -52,8 +51,8 @@ object GrpcEntityHelpers {
       system: ActorSystem) =
     e.map(m.serialize).via(Grpc.grpcFramingEncoder(codec)).map(bytes => HttpEntity.Chunk(bytes)).concat(trail)
 
-  def trailer(status: Status): LastChunk =
-    LastChunk(trailer = statusHeaders(status))
+  def trailer(status: Status, headers: Seq[HttpHeader] = Nil): LastChunk =
+    LastChunk(trailer = statusHeaders(status) ++ headers)
```
raboof: :+1:

runtime/src/main/scala/akka/grpc/javadsl/GrpcErrorResponse.scala
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (C) 2018-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.javadsl
+
+import java.lang.{ Iterable => jIterable }
```
raboof: while there indeed also exists an `Iterable` in Scala, I think that's used rarely enough that you can just use `java.lang.Iterable` without an alias like this

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
```diff
@@ -4,48 +4,56 @@
 
 package akka.grpc.javadsl
 
+import java.lang.{ Iterable => jIterable }
 import java.util.concurrent.CompletionException
+import java.util.{ ArrayList => jArrayList }
```
raboof: (I think this can also be just `ArrayList`)

interop-tests/src/test/scala/akka/grpc/scaladsl/GrpcExceptionHandlerSpec.scala
drmontgomery: Done!

runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
drmontgomery: Updated to List.

runtime/src/main/scala/akka/grpc/internal/GrpcEntityHelpers.scala
drmontgomery: Updated to List.

runtime/src/main/scala/akka/grpc/javadsl/GrpcErrorResponse.scala
drmontgomery: I like using the j-prefix when referencing Java types in a Scala source file just to indicate that something slightly unusual is going on.

That said, I've removed the prefix everywhere except in GrpcExceptionHelper, where there is enough Java/Scala conversion code that it's nice to have consistent prefixes on the parameter and return types.

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
drmontgomery: Done!

drmontgomery:
I've also created two new branches in my repo to explore some of the options we've been discussing.

In the wip-rich-error-response-headers branch, I removed GrpcErrorResponse and replaced it with a list of http headers.

In the wip-rich-error-response-metadata branch, I modified GrpcErrorResponse to include a String -> MetadataEntry map. I also moved GrpcErrorResponse from the language-specific packages to the akka.grpc package, along with MetadataEntry, StringEntry and BytesEntry.

Is there a good way for me to package these up for you to review, or is mentioning them here enough? (I could create a pull request, I suppose, but that seems like it risks fragmenting the discussion.)

ignasi35:
> # GrpcErrorResponse
> (not quoting everything) 

I think there's room for a new type but instead of `GrpcErrorResponse(...)` we could call it [`Trailers(Status, Metadata)`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses) (where `Metadata` is the [already available `Metadata` type including a collection of headers(https://github.com/akka/akka-grpc/blob/992f19f213fb1c267c091d261ac59e419f994ea7/runtime/src/main/scala/akka/grpc/scaladsl/Metadata.scala#L20)). Then, the errorHandler would be `PartialFunction[Throwable, Trailers]`. I'm in favor of a more strongly typed approach.

Still, I think a collection of helpers for users to easily build the headers for that `Trailers` instance is more valuable (or even necessary) since `Trailers` encapsulates `Metadata` which encapsulates `List[HttpHeader]` after all.

drmontgomery:
I've updated the branch based on our discussion. (I also rebased it onto the current master branch.)

I introduced a MetadataBuilder class to handle constructing immutable Metadata instances on the server side. (The client-side metadata creation code was coupled to the RequestBuilder interface.)

I ended up refactoring the Metadata implementation classes pretty heavily. I think it ended up in a good place, but I'd be happy to walk it back a bit if you think the changes are too much.

raboof:
> I ended up refactoring the Metadata implementation classes pretty heavily. I think it ended up in a good place, but I'd be happy to walk it back a bit if you think the changes are too much.

I see what you mean - I think it's OK, much of the 'tricky' stuff is internal anyway so that's OK.

I think this is pretty close to being mergable. mimaReportBinaryIssues is reporting some incompatibilities - since we are still before 1.0.0 that's fine and you can just add the `ProblemFilter` entries to a backwards-excludes file as described in CONTRIBUTING.md (https://travis-ci.org/akka/akka-grpc/jobs/657617901#L531). I also suspect we need to run `sbt scalafmtAll` to make sure all the formatting is consistent.

Let me know if you want any help with that!

drmontgomery:
Thanks for info about mimaReportBinaryIssues!

I also ran scalafmtAll, but nothing changed.

raboof:
> Thanks for info about mimaReportBinaryIssues!

Sure!

> I also ran scalafmtAll, but nothing changed

Then everything was formatted correctly already :+1:

runtime/src/main/mima-filters/0.7.3.backwards.excludes/838-custom-headers.backwards.excludes
```diff
@@ -0,0 +1,40 @@
+# Replaced javadsl.MetadataImpl with internal.JavaMetadataImpl adapter.
+ProblemFilters.exclude[MissingClassProblem]("akka.grpc.javadsl.MetadataImpl")
+
+# Replaced scaladsl.MetadataImpl with internal.HeaderMetadataImpl.
+ProblemFilters.exclude[MissingClassProblem]("akka.grpc.scaladsl.MetadataImpl")
+ProblemFilters.exclude[MissingClassProblem]("akka.grpc.scaladsl.MetadataImpl$")
+
+# Modified GrpcExceptionHandler to return Trailers instead of io.grpc.Status.
+ProblemFilters.exclude[IncompatibleSignatureProblem]("akka.grpc.javadsl.GrpcExceptionHandler.standard")
```
raboof: you can even use wildcards here but this is :+1:

raboof:
Great contribution, thanks!

PR #824, opened 2020-02-17
Less Materializer in the API

raboof:
To make it more natural to use with 2.6

~Not for merge until Akka 2.5.30 is out, because of
https://github.com/akka/akka/pull/28538~

Issue #761, opened 2019-12-18
Binary compatibility promise


raboof:
We are gearing up for Akka 1.0 early 2020.

One thing we need to decide on is the level of binary compatibility we want to promise.

In contrast to for example Akka and Akka HTTP, we mainly see Akka gRPC used 'directly' in non-library projects. A notable exception is https://doc.akka.io/docs/alpakka/current/google-cloud-pub-sub-grpc.html .

There are 3 forms of binary compatibility that are potentially interesting:

* If you have an application that uses a library that uses Akka gRPC X to generate code, and also another library that uses Akka gRPC X+1 to generate code, should you be able to run these side-by-side in a project using Akka gRPC X+1 (or later)?
* If you have a 'user' library that has a dependency of a 'deeper' library that was built by Akka gRPC X, and there is a newer version of that 'deeper' library that was built with Akka gRPC X+1, should an application be allowed to upgrade the 'deeper' library without having to recompile/upgrade the 'user' library?
* If you are using a build system plugin that uses akka-grpc (such as [play-grpc](https://github.com/playframework/play-grpc), should it be possible to update akka-grpc in your build without also updating play-grpc?

In technical terms, these translate to:
* binary compatibility of the akka-grpc-runtime library
* binary compatibility of the generated code across versions of Akka gRPC
* binary compatibility of the build system plugin components

I think the second and third forms of binary compatibility do not seem necessary for Akka gRPC.

The first we should likely uphold to a certain degree, though perhaps we can still identify parts of the runtime API that do not need such strong guarantees.

ennru:
The rules should point to [Akka's page](https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html) about it.

ignasi35:
> I think the second and third forms of binary compatibility do not seem necessary for Akka gRPC.

I think I agree. IIUC, this means a user of `play-grpc` stuck on version `akka-grpc`'s `codegen` version `X` would be able to use the generated code with `runtime` X+1.

If that's the case you had in mind, then I think we agree.

ignasi35:
Hi @raboof, I've noticed https://github.com/akka/akka-grpc/pull/900 and wondered what's left... Do you have an open list of areas that need reviewing?

raboof:
I think we've been working towards this steadily over the last few releases. I don't have any particular area's that I know need work, so I think there's 4 things we can do:

[...]

PR #842, opened 2020-02-28
Document binary compatibility

raboof:
Document how we want to approach this from 1.0.0 going forward

Refs #761 but does not close it yet, we want to apply #824,
review the API's and add @InternalAPI/@ApiMayChange where
applicable before that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,10 +98,11 @@ Akka gRPC is still in experimental mode, so this section does not apply yet.
 However, once we declare the project stable, we will adhere to the following
 rules:
 
-Binary compatibility rules and guarantees are described in depth in the [Binary Compatibility Rules
-](https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html) section of the documentation.
+Our binary compatibility guarantees are described in depth in the
+[Binary Compatibility](https://doc.akka.io/docs/akka-grpc/current/binary-compatibility.html)
+section of the documentation.
 
-Akka HTTP uses MiMa (which is short for [Lightbend Migration Manager](https://github.com/lightbend/migration-manager)) to
+Akka gRPC uses MiMa (which is short for [Lightbend Migration Manager](https://github.com/lightbend/migration-manager)) to
 validate binary compatibility of incoming Pull Requests. If your PR fails due to binary compatibility issues, you may see 
 an error like this:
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -98,10 +98,11 @@ Akka gRPC is still in experimental mode, so this section does not apply yet.
...
 rules:
 
-Binary compatibility rules and guarantees are described in depth in the [Binary Compatibility Rules
-](https://doc.akka.io/docs/akka/current/common/binary-compatibility-rules.html) section of the documentation.
+Our binary compatibility guarantees are described in depth in the
+[Binary Compatibility](https://doc.akka.io/docs/akka-grpc/current/binary-compatibility.html)
+section of the documentation.
 
-Akka HTTP uses MiMa (which is short for [Lightbend Migration Manager](https://github.com/lightbend/migration-manager)) to
+Akka gRPC uses MiMa (which is short for [Lightbend Migration Manager](https://github.com/lightbend/migration-manager)) to
```
ennru:
```suggestion
Akka gRPC uses [Lightbend MiMa](https://github.com/lightbend/mima) to
```

docs/src/main/paradox/binary-compatibility.md
```diff
@@ -0,0 +1,37 @@
...
+Internal API's (designated by the `akka.grpc.internal` package or with the `@InternalApi` annotation) and API's that are still marked `@ApiMayChange` are not guaranteed to remain binary compatible.
+
+Libraries that use such methods may not work in applications that depend on a newer version of Akka gRPC.
+
+## Downstream libraries
+
+We depend on a number of downstream libraries that don't formally maintain binary compatibility,
+such as [ScalaPB](https://scalapb.github.io/) (when generating Scala code) and 
+[grpc-java](https://github.com/grpc/grpc-java/). Updates to those libraries might introduce
+incompatibilities that are outside the scope of Akka gRPC's binary compatibility.
```
ennru: An upgrade of those might require a new major release of Akka gRPC, right?

raboof: Yes, it might, but I don't think it must. When grpc-java makes a change that breaks any old code from working with new code, then yes. However, when for instance scalapb makes a change that only breaks code that uses some particular extension, then I think while this is 'strictly binary incompatible' it might make sense decide how we let this impact versioning on a case-by-case basis. What do you think?

ennru: Yes, that's what I want.
To me "might introduce incompatibilities that are outside the scope of Akka gRPC's binary compatibility" sounds a bit like: "We don't care if these upgrades could break anything".
So maybe write that we will need to decide depending on what those libraries change.

PR #843, opened 2020-02-28
Simplify ClientState

raboof:
As rediscovering/retrying is now the responsibility of the netty client

/cc @tayvs since you worked on retry logic, that should be intact and not rely on the retry logic in CiientState AFAICS

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
@@ -9,21 +9,20 @@ import java.util.concurrent.atomic.AtomicInteger
 
 import scala.concurrent.{ Await, Future }
 import scala.concurrent.duration._
-
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.discovery.{ Lookup, ServiceDiscovery }
 import akka.discovery.ServiceDiscovery.{ Resolved, ResolvedTarget }
 import akka.grpc.GrpcClientSettings
+import akka.grpc.internal.ClientConnectionException
 import akka.http.scaladsl.Http
-import akka.http.scaladsl.Http.ServerBinding
 import akka.stream.scaladsl.Source
-
 import org.scalatest.wordspec.AnyWordSpec
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.BeforeAndAfterAll
-
 import example.myapp.helloworld.grpc.helloworld._
+import io.grpc.Status.Code
+import io.grpc.StatusRuntimeException
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.time.Span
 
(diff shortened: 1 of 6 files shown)


tayvs:
@raboof So you remove channel recreation logic? Now recreation would work on subchannel level and if 'super channel' failed, then `monitorChannel` would catch failure and close client, Am I right?

raboof:
> So you remove channel recreation logic?

I removed the recreation logic in ClientState, because in #811 we switched things around so that discovery is also done 'inside' the Channel. This means when discovery fails it no longer prevents creation of the Channel, so it doesn't need to be retried.

> Now recreation would work on subchannel level and if 'super channel' failed, then monitorChannel would catch failure and close client, Am I right?

I think the 'super channel' will now never fail anymore when there is a recoverable failure - do you know of a scenario where it would?

tayvs:
@raboof I investigated about cases where 'super channel' would fail and it seems that it would not fail. If subchannel fail then it would be rediscovered and recreated.

raboof:
Great, thanks for validating!

interop-tests/src/test/scala/akka/grpc/scaladsl/LoadBalancingIntegrationSpec.scala
```diff
@@ -133,6 +133,28 @@ class LoadBalancingIntegrationSpec extends AnyWordSpec with Matchers with Before
       service1.greetings.get should be(requestsPerServer)
       service2.greetings.get should be(requestsPerServer)
     }
+
+    "select the right endpoint among invalid ones" in {
```
ennru: Maybe add a test case without a valid address to validate failing discovery?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -9,6 +9,7 @@ import java.util.concurrent.atomic.AtomicReference
 
 import akka.Done
 import akka.annotation.InternalApi
+import akka.annotation.InternalStableApi
 import akka.event.LoggingAdapter
 import akka.grpc.GrpcClientSettings
 import akka.pattern.Patterns
```
ennru: Some imports can go now.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -22,138 +23,35 @@ import scala.util.Failure
...
+    this(settings, log, NettyClientUtils.createChannel(settings, log))
 
   mat match {
     case m: ActorMaterializer =>
       m.system.whenTerminated.foreach(_ => close())(ex)
     case _ =>
   }
 
-  // used from generated client
+  // TODO #761 consider not leaking ManagedChannel to the generated code?
```
ennru: This seems important. Is there a reason why users would need the "Managed" channel?

raboof: I guess it was convenient at the time, but let's see if we can refactor that (in a different PR).

PR #856, opened 2020-03-09
Autolabel more things

ennru:


ennru:
You were right, what is set up is the [PR labeler action](https://github.com/TimonVS/pr-labeler-action), so I tried to modify the wrong file.
Updated now and enabled [Autolabeler Probot](https://github.com/mithro/autolabeler).

PR #861, opened 2020-03-11
Don't generate unidoc for codegen projects

raboof:


ennru:
A `scalafmtSbt` is required.

PR #864, opened 2020-03-12
Mark more API's internal or ApiMayChange

raboof:
First step towards 1.0.0 - currently we don't promise Java/Scala API stability
yet anyway.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientResponse.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientResponse.scala
@@ -6,7 +6,7 @@ package akka.grpc
 
 import java.util.concurrent.CompletionStage
 
-import akka.annotation.DoNotInherit
+import akka.annotation.{ ApiMayChange, DoNotInherit }
 
 import scala.concurrent.Future
 
@@ -16,6 +16,7 @@ import scala.concurrent.Future
  * Not for user extension
  */
 @DoNotInherit
+@ApiMayChange
 trait GrpcResponseMetadata {
 
   /**
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -5,7 +5,7 @@
 package akka.grpc
 
 import akka.actor.ActorSystem
-import akka.annotation.{ ApiMayChange, InternalApi }
+import akka.annotation.{ ApiMayChange, DoNotInherit, InternalApi }
 import akka.discovery.{ Discovery, ServiceDiscovery }
 import akka.discovery.ServiceDiscovery.{ Resolved, ResolvedTarget }
 import akka.grpc.internal.HardcodedServiceDiscovery
@@ -23,6 +23,7 @@ import com.typesafe.sslconfig.ssl.{
 import io.grpc.CallCredentials
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
 import javax.net.ssl.SSLContext
+
 import scala.collection.immutable
 import scala.concurrent.duration.{ Duration, _ }
 
@@ -190,6 +191,8 @@ object GrpcClientSettings {
       None
 }
 
+@ApiMayChange
+@DoNotInherit
 final class GrpcClientSettings private (
     val serviceName: String,
     val serviceDiscovery: ServiceDiscovery,
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -5,7 +5,7 @@
 package akka.grpc
 
 import akka.actor.ActorSystem
-import akka.annotation.{ ApiMayChange, DoNotInherit, InternalApi }
+import akka.annotation.{ ApiMayChange, InternalApi }
 import akka.discovery.{ Discovery, ServiceDiscovery }
 import akka.discovery.ServiceDiscovery.{ Resolved, ResolvedTarget }
 import akka.grpc.internal.HardcodedServiceDiscovery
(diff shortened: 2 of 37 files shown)


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -190,6 +191,8 @@ object GrpcClientSettings {
       None
 }
 
+@ApiMayChange
+@DoNotInherit
```
jrudolph: Do we add `@DoNotInherit` to final classes?

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -21,27 +22,34 @@ import scala.util.Try
 trait GrpcProtocol {
```
jrudolph: Shouldn't the whole class be `@DoNotInherit` and/or `@InternalApi` when it can only be implemented internally?

runtime/src/main/scala/akka/grpc/scaladsl/Metadata.scala
```diff
@@ -29,10 +30,12 @@ import akka.util.ByteString
   /**
    * @return The metadata as a map.
    */
+  @ApiMayChange
```
jrudolph: Do those need extra annotations if they are inside of an `@ApiMayChange` trait?

runtime/src/main/scala/akka/grpc/internal/Identity.scala
```diff
@@ -1,8 +1,4 @@
-/*
- * Copyright (C) 2018-2020 Lightbend Inc. <https://www.lightbend.com>
- */
-
-package akka.grpc
+package akka.grpc.internal
```
jrudolph: Copyright now missing? Also in other moved files.

Do we put `@InternalApi` on classes in an `internal` package? For me, the package name seems to be a strong enough indication.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
raboof: Not sure yet, I suspect 'GrpcProtocol' will show up in public API's (even if the specific implementations don't and users can't provide their own) - but maybe it can become internal entirely.

runtime/src/main/scala/akka/grpc/GrpcClientResponse.scala
```diff
@@ -16,6 +16,7 @@ import scala.concurrent.Future
  * Not for user extension
  */
 @DoNotInherit
+@ApiMayChange
```
ennru: Is there an issue to point to?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -190,6 +191,7 @@ object GrpcClientSettings {
       None
 }
 
+@ApiMayChange
```
ennru: To mark the settings is a bit broad, API-wise there is mostly always a way forward. But settings might become irrelevant, of course.

runtime/src/main/scala/akka/grpc/Trailers.scala
```diff
@@ -4,16 +4,19 @@
...
 package akka.grpc
 
+import akka.annotation.ApiMayChange
 import io.grpc.Status
 import akka.grpc.internal.JavaMetadataImpl
 import akka.grpc.scaladsl.{ Metadata, MetadataBuilder }
 import akka.grpc.javadsl.{ Metadata => jMetadata }
 
+@ApiMayChange
 case class Trailers(status: Status, metadata: Metadata) {
```
ennru: Could this be turned into a regular class to ease bin-compatible evolvement?

runtime/src/main/scala/akka/grpc/javadsl/MetadataBuilder.scala
```diff
@@ -6,19 +6,20 @@ package akka.grpc.javadsl
...
 import akka.util.ByteString
-
 import akka.grpc.scaladsl
 import akka.grpc.internal.JavaMetadataImpl
 
 /**
  * This class provides an interface for constructing immutable Metadata instances.
  */
+@ApiMayChange
 class MetadataBuilder {
```
ennru: make it final and the constructor `private` instead?

runtime/src/main/scala/akka/grpc/scaladsl/MetadataEntry.scala
```diff
@@ -4,13 +4,14 @@
...
+import akka.annotation.ApiMayChange
 import akka.util.ByteString
-
 import akka.grpc.javadsl
 
 /**
  * Represents a entry (value) in a collection of Metadata.
  */
+@ApiMayChange
 sealed trait MetadataEntry extends javadsl.MetadataEntry
```
ennru: Mark the case classes final?

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
raboof: I mostly meant to mark this constructor ApiMayChange, since not doing that would mean adding additional bincompat constructors each time a setting is added.

runtime/src/main/scala/akka/grpc/Trailers.scala
raboof: possibly, but I'm not yet sure about the `Metadata` API's yet - perhaps we could use more of https://grpc.github.io/grpc-java/javadoc/io/grpc/Metadata.html . Since this seems like a poweruser api it seemed low-cost to make the whole class ApiMayChange for now.

raboof:
> I guess we might want to do another round for 1.0.

Definitely!

PR #868, opened 2020-03-13
Update MiMa

raboof:


ignasi35:
The upgrade of `` in 159b819990151539e3a4d31e330fedf42903ba3e produces a generated companion object where `def of(...)` has a different signature. Take `ServerReflectionResponse#of` for example:

```diff
  def of(
    validHost: _root_.scala.Predef.String,
    originalRequest: _root_.scala.Option[grpc.reflection.v1alpha.reflection.ServerReflectionRequest],
    messageResponse: grpc.reflection.v1alpha.reflection.ServerReflectionResponse.MessageResponse,
-    unknownFields: _root_.scalapb.UnknownFieldSet
  ): _root_.grpc.reflection.v1alpha.reflection.ServerReflectionResponse = _root_.grpc.reflection.v1alpha.reflection.ServerReflectionResponse(
    validHost,
    originalRequest,
    messageResponse,
    unknownFields
  )

```

raboof:
Right, this makes sense: this is the change from https://github.com/scalapb/ScalaPB/issues/778, and we didn't notice this in #865 because we didn't have generated classes yet in the previous version, 0.7.3.

So this makes sense, and we don't need to roll back #865 because we're pre-1.0.0 and this is a change we're happy to see. I'll add the appropriate exclusions.

build.sbt
```diff
@@ -41,7 +41,7 @@ lazy val runtime = Project(id = akkaGrpcRuntimeName, base = file("runtime"))
   .settings(
     // We don't actually promise binary compatibility before 1.0.0, but want to
     // introduce the tooling
-    mimaPreviousArtifacts := Set(organization.value %% "akka-grpc-runtime" % "0.7.3"),
+    mimaPreviousArtifacts := Set(organization.value %% "akka-grpc-runtime" % "0.8.0"),
```
ennru: Shouldn't we use `previousStableVersion` here?
```
      mimaPreviousArtifacts := Set(
          organization.value %% name.value % previousStableVersion.value
            .getOrElse(throw new Error("Unable to determine previous version"))
        ),
```

Maybe always excluding internal packages would be attractive?
```
mimaBinaryIssueFilters += ProblemFilters.exclude[Problem]("akka.grpc.internal.*")
```

raboof: Updated to use `previousStableVersion`, wasn't aware of it.

I don't think we want to exclude internal packages, as they might still have `InternalStableApi`

Issue #647, opened 2019-07-26
 Java code generation requires that "option java_package" be defined


orendain:
### Versions used 

akkaGrpcVersion = 0.7.0
protobufVersion = 3.4.0

In case it's relevant, I'm using the Gradle plugin - no SBT in this project.

### Expected Behavior

A protobuf file without the optional `option java_package = "my.java.package"` should fallback to using the protobuf `package my.protobuf.package`.

### Actual Behavior

Without setting `option java_package`, the generated Java files have no package.

Generated files will look like:
```java
// Generated by Akka gRPC. DO NOT EDIT.
package ;
...
```

References to other classes will have a leading period without package names:
```java
java.util.concurrent.CompletionStage<.MyProtobufService>
// Notice the period before "MyProtobufService"
```

### Reproducible Test Case

See the gradle project here: https://github.com/orendain/akka-grpc/tree/eo/example-for-issue-647/issue-647-example

Instructions:
`gradle build` to generate the files.  Notice the errors generated.  Also look at `build/generated/source/proto/main/akkaGrpc/com.orendainx/` to see the generated files and their missing packages.

orendain:
This may be related to, but is a different bug than, #619.

PR #873, opened 2020-03-14
codegen: fallback to protobuf package when no java_package is defined

bjaglin:
Follow the behavior of:
- protobuf-java: https://github.com/protocolbuffers/protobuf/blob/2228af6/src/google/protobuf/compiler/java/java_helpers.cc#L243-L257
- scalapb: https://github.com/scalapb/ScalaPB/blob/d5e3c33/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala#L831-L835

The Scala codegen was already working without a `java_package` (since it uses the scalapb descriptor implicits), but I added a scripted test anyway.


raboof:
OK TO TEST

raboof:
This fixes #647

PR #874, opened 2020-03-17
server reflection: support service/message definitions across files

bjaglin:
Clients using server reflection were failing to inspect messages declared in another protobuf file that the one containing the service(s).

This aligns the `akka-grpc` implementation with the reference implementation: https://github.com/grpc/grpc-java/blob/3a39b81/services/src/main/java/io/grpc/protobuf/services/ProtoReflectionService.java#L309.

I haven't found an easy way to add a test - ideally `grpc_cli ls` should run in the scripted test, but that's not trivial.


raboof:
OK TO TEST

raboof:
PLS BUILD

PR #876, opened 2020-03-19
Properly camel-case names with dashes

raboof:


codegen/src/main/scala/akka/grpc/gen/javadsl/Service.scala
```diff
@@ -60,6 +61,20 @@ object Service {
 
   private[javadsl] def toCamelCase(name: String): String = {
     if (name.isEmpty) ""
-    else name.head.toUpper + "_[a-z]".r.replaceAllIn(name.tail, s => s.group(0)(1).toUpper.toString)
+    else toCamelCaseRec(name.tail, new StringBuffer(name.head.toUpper.toString), false)
+  }
+
+  @tailrec
+  def toCamelCaseRec(in: String, out: StringBuffer, capNext: Boolean): String = {
```
ennru: `private`

codegen/src/test/scala/akka/grpc/gen/javadsl/ServiceSpec.scala
```diff
@@ -11,6 +11,7 @@ class ServiceSpec extends AnyWordSpec with Matchers {
   "The Service model" should {
     "correctly camelcase strings" in {
       Service.toCamelCase("foo_bar") should be("FooBar")
+      Service.toCamelCase("grpc-example") should be("GrpcExample")
```
ennru: Add a test with digits, please.

interop-tests/src/main/protobuf/crdt-example.proto
```diff
@@ -0,0 +1,65 @@
+syntax = "proto3";
```
ennru: Why is this file in this PR?

raboof: Without the change in this PR, this .proto would generate gRPC javadsl service code that referred to a `Crdt-example` outer class, which wouldn't compile, reproducing the original problem.

PR #877, opened 2020-03-19
Correctly deal with conflicting class names

raboof:
On top of #876


codegen/src/main/scala/akka/grpc/gen/javadsl/Service.scala
```diff
@@ -59,6 +55,22 @@ object Service {
   private[javadsl] def basename(name: String): String =
     name.replaceAll("^.*/", "").replaceAll("\\.[^\\.]*$", "")
 
+  private[javadsl] def outerClass(t: FileDescriptor) =
```
ennru: Could be regular `private` so that Java users don't see it? Otherwise, `@InternalApi` might be good.

raboof: It's used from `Method` as well so it can't be regular `private`.

Good point, though in this case this code is *generating* the javadsl, so I don't think it'll ever be used from Java.

Issue #628, opened 2019-07-17
Including PB.externalIncludePath in PB.protoSources leads to redundant class generation


adkafka:
### Versions used 

Scala version: 2.12.8
Akka version: 2.5.23
Akka-gRPC version: 0.6.2

### Expected Behavior

When defining a `.proto` file that relies on a google type (`import "google/protobuf/wrappers.proto";`), the code generation should not regenerate these files, and it should use the compiled classes included in the runtime jar (`scalapb-runtime`).

### Actual Behavior

Due to [this line](https://github.com/akka/akka-grpc/blob/master/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L118) (`PB.protoSources += PB.externalIncludePath.value`), all files in the `externalIncludePath` will be codegen'ed through ScalaPB. This results in regenerating the code for the google wrapper types, despite them being included in the runtime jar.

### Relevant logs
SBT logs that these external schemas are compiled:

```
[info] Compiling schema /app/target/protobuf_external/google/protobuf/compiler/plugin.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/descriptor.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/struct.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/any.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/duration.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/field_mask.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/source_context.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/type.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/empty.proto
[info] Compiling schema /app/target/protobuf_external/scalapb/scalapb.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/timestamp.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/wrappers.proto
[info] Compiling schema /app/target/protobuf_external/google/protobuf/api.proto
```

### Example
* Start with the [akka-grpc-quickstart-scala](https://developer.lightbend.com/guides/akka-grpc-quickstart-scala/) project
* Add an import to the `src/main/protobuf/helloworld.proto` file
  * `import "google/protobuf/wrappers.proto"`
[...]

PR #878, opened 2020-03-19
Don't generate code for external proto's by default

raboof:
Refs #628

### DIFF
--- a/docs/src/main/paradox/buildtools/sbt.md
+++ b/docs/src/main/paradox/buildtools/sbt.md
@@ -85,6 +85,20 @@ that contain proto definitions to your build:
 libraryDependencies += "com.example" %% "my-grpc-service" % "1.0.0" % "protobuf"
 ```
 
+This by default assumes this artifact also contains the correct generated classes for this API.
+If you want to generate your own, either because the upstream artifact contains no generated
+files or they were generated in a way that is incompatible with your intended use, you can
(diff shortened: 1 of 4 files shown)


docs/src/main/paradox/buildtools/sbt.md
```diff
@@ -85,6 +88,18 @@ that contain proto definitions to your build:
...
+This by default assumes this artifact also contains the correct generated classes for this API.
+If you want to generate your own, either because the upstream artifact contains no generated
+files or they were generated in a way that is incompatible with your intended use, you can
+add add them with:
+
+@@snip[build.sbt](/sbt-plugin/src/sbt-test/gen-scala-server/00-interop/build.sbt) { #external }
+
+For a discussion around a more fine-grained control over how artifacts with `.proto` definitions
+can be used see [this sbt-protoc issue](https://github.com/thesamet/sbt-protoc/issues/144).
+In the short term there is nothing wrong with duplicating the `.proto` files of the protoco
```
bjaglin:
```suggestion
In the short term there is nothing wrong with duplicating the `.proto` files of the protocol
```

raboof:
> happy to see the behavior going into this direction!

Thanks - I still think I *prefer* external artifacts to contain just .proto files and not generated code, but I agree this belongs in sbt-protoc rather than akka-grpc

Issue #238, opened 2018-06-18
Support JSON encoding on the server


johanandren:
https://developers.google.com/protocol-buffers/docs/proto3#json

For easier ad-hoc testing from the command line for example.

raboof:
Interesting, didn't know about that mapping.

In the mean time https://github.com/njpatel/grpcc seems like a useful tool.

johanandren:
Additional inspiration: https://www.envoyproxy.io/docs/envoy/latest/api-v1/http_filters/grpc_json_transcoder_filter

raboof:
See also https://github.com/cloudstateio/cloudstate/blob/master/protocols/example/shoppingcart/shoppingcart.proto#L48, https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala

raboof:
@tdrozdowski this could be an interesting contribution, I don't know of anyone planning to take it on at this point.

PR #881, opened 2020-03-19
WIP: JSON serialization support

timw:
Initial implementation of `+json` serialization in all gRPC protocols.

References #238

I implemented this to see how much more work would be required to support JSON in the protocol/negotiation stack after it was extended to support gRPC-Web.

This is mostly functionally complete, but has a few issues.

Functional issues:
- There's no way I can see to fully wire this into the client at the moment, so I've stubbed it through the client as far as possible to prove the approach. You can control the marshallers externally, but the `io.grpc` client hard-wires the request content type so things wouldn't work too well in practice.
- There aren't any functional gRPC clients I know of that implement `+json` now, so it'd primarily useful in debugging until then, but then tools like `grpcurl` already allow you to use JSON input and output at the client level. The gRPC-Web roadmap does list JSON (or JSON like) encoding as something on the roadmap to make the (de)serialization more performant in browsers.

Commit issues:
- I wasn't sure where to put test protobuf definitions (which project and which folder - couldn't convince sbt to build them from `src/test/protobuf`) so I've put in one spec testing the JSON serialization for now and haven't tried to minimise the service spec, and that only for scala. Will possibly need to go into the interop or plugin-tester projects to test both the javadsl and scaladsl paths.
- Doc updates required


johanandren:
This has been silent for a couple of years and conflicts with pretty much every file in the repo, I'll close it for now, feel free to re-open in the future if you want to continue the work @timw

Issue #886, opened 2020-03-24
Substream Source  cannot be materialized more than once in GrpchMarshalling


marekzebrowski:
I have a very simple request-response grpc service that started to fail after upgrade to akka-grpc 0.8.1
actual service implementation is rather boring, but something is wrong in marshalling part.
generated code that matches following stacktrace looks like:
```scala
  GrpcMarshalling.unmarshal(request.entity.dataBytes)(AddEmojiRequestSerializer, mat, reader)
                  .flatMap(implementation.addEmoji(_))
                  .map(e => GrpcMarshalling.marshal(e, eHandler)(AddEmojiResponseSerializer, writer, system))
```
so nothing fancy here, yet it fails

### Versions used 
Akka version: 2.5.30
akka-http: 10.1.11
akka-grpc: 0.8.1
Scala 2.13.1
### Expected Behavior
Response is serialized properly - that worked fine up to akka-grpc 0.7.3

### Relevant logs
```
14:34:12.223 [emojis-akka.actor.default-dispatcher-5] ERROR akka.actor.ActorSystemImpl - Unhandled error: [Substream Source(substream-out-3) cannot be materialized more than once].
java.lang.IllegalStateException: Substream Source(substream-out-3) cannot be materialized more than once
	at akka.stream.impl.fusing.SubSource$$anon$13.createMaterializedTwiceException(StreamOfStreams.scala:830)
	at akka.stream.impl.fusing.SubSource$$anon$13.<init>(StreamOfStreams.scala:801)
	at akka.stream.impl.fusing.SubSource.createLogic(StreamOfStreams.scala:797)
	at akka.stream.stage.GraphStage.createLogicAndMaterializedValue(GraphStage.scala:107)
	at akka.stream.stage.GraphStageWithMaterializedValue.createLogicAndMaterializedValue(GraphStage.scala:51)
	at akka.stream.impl.GraphStageIsland.materializeAtomic(PhasedFusingActorMaterializer.scala:675)
	at akka.stream.impl.PhasedFusingActorMaterializer.materialize(PhasedFusingActorMaterializer.scala:492)
	at akka.stream.impl.PhasedFusingActorMaterializer.materialize(PhasedFusingActorMaterializer.scala:451)
	at akka.stream.impl.PhasedFusingActorMaterializer.materialize(PhasedFusingActorMaterializer.scala:444)
	at akka.stream.scaladsl.RunnableGraph.run(Flow.scala:629)
[...]

PR #888, opened 2020-03-25
Fix isDefinedAt bug in concatOrNotFound

raboof:
Refs #886

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/CountingGreeterServiceImpl.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/CountingGreeterServiceImpl.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright (C) 2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.scaladsl
+
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.concurrent.Future
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+
+import example.myapp.helloworld.grpc.helloworld._
+
+class CountingGreeterServiceImpl extends GreeterService {
+  var greetings = new AtomicInteger(0);
+
+  def sayHello(in: HelloRequest): Future[HelloReply] = {
+    greetings.incrementAndGet()
+    Future.successful(HelloReply(s"Hi ${in.name}!"))
+  }
+
+  def itKeepsReplying(in: HelloRequest): Source[HelloReply, NotUsed] =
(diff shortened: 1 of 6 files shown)


jrudolph:
Without digging in more deeply, it's hard to understand what was wrong and how you fixed it. Could you give a short explanation?

interop-tests/src/test/scala/akka/grpc/scaladsl/CountingGreeterServiceImpl.scala
```diff
@@ -0,0 +1,32 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+import scala.concurrent.Future
+
+import akka.NotUsed
+import akka.stream.scaladsl.Source
+
+import example.myapp.helloworld.grpc.helloworld._
+
+class CountingGreeterServiceImpl extends GreeterService {
```
jrudolph: Is this one covering the bug?

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
```diff
@@ -29,31 +27,13 @@ object ServiceHandler {
...
-    handler(handlers: _*)
-
-  /**
-   * Creates a `HttpRequest` to `HttpResponse` handler for gRPC services that can be used in
-   * for example `Http().bindAndHandleAsync` for the generated partial function handlers:
-   *  - The generated handler supports the `application/grpc` media type.
-   *  - If the request is for a handled service, but with an invalid media type, then a _415: Unsupported Media Type_ response is produced.
-   *  - Otherise if the request is not handled by one of the provided handlers, a _404: Not Found_ response is produced.
-   */
-  def handler(handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*): HttpRequest => Future[HttpResponse] = {
```
jrudolph: Is that one only internal so that it can just be removed?

raboof:
The previous implementation used `servicesHandler.isDefinedAt`, but depending on the implementation of the `PartialFunction` (including in this case), `isDefinedAt` works by actually calling the function and seeing if it returns...

interop-tests/src/test/scala/akka/grpc/scaladsl/CountingGreeterServiceImpl.scala
raboof: (no, sorry, that was a cleanup I made while reproducing the bug in a testcase, but I ended up testing it with another project..)

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
ennru: No, this it public API in 0.8

PR #896, opened 2020-03-31
bump sbt-protoc & fix compatibility with other sbt-protoc plugins

bjaglin:
This is a second take at https://github.com/akka/akka-grpc/pull/696, which had silently regressed because of https://github.com/akka/akka-grpc/pull/702.

As described in the commit messages:
* cc1c166 fixes and enforce the compatibility with other plugins affecting sbt-protoc keys in `Compile` (what should have been done in https://github.com/akka/akka-grpc/pull/696)
* 415866b bumps sbt-protoc (which does not need any change)
* 66850c0 makes use of https://github.com/thesamet/sbt-protoc/blob/master/CHANGELOG.md#09931 to extend the support of cc1c166 to `Test`
* 2bd5204 is a bit unrelated, but makes more explicit what is needed for `IntegrationTest` to work as expected (as already tested by https://github.com/akka/akka-grpc/blob/v0.8.2/sbt-plugin/src/sbt-test/gen-scala-server/03-test-config/build.sbt#L10-L14)

I chose to group all these commits in the same PR as the sbt-protoc bump is mostly useful for configuration improvements.

### DIFF
--- a/docs/src/main/paradox/buildtools/sbt.md
+++ b/docs/src/main/paradox/buildtools/sbt.md
@@ -22,11 +22,9 @@ you have to inherit the `.proto` definitions from `Compile` over to `Test`:
 
 @@snip[build.sbt](/sbt-plugin/src/sbt-test/gen-scala-server/03-test-config/build.sbt) { #test }
 
-If you have other configurations with `.proto` sources (for example `IntegrationTest`), you can enable the plugin on them:
+If you have other configurations with `.proto` sources (for example `IntegrationTest`), you must first declare them and enable the plugin on them:
 
-```
-.settings(AkkaGrpcPlugin.configSettings(IntegrationTest))
-```
+@@snip[build.sbt](/sbt-plugin/src/sbt-test/gen-scala-server/03-test-config/build.sbt) { #it }
 
 ### Generating server "power APIs"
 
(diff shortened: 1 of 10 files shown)


docs/src/main/paradox/buildtools/sbt.md
```diff
@@ -25,7 +25,9 @@ you have to inherit the `.proto` definitions from `Compile` over to `Test`:
 If you have other configurations with `.proto` sources (for example `IntegrationTest`), you can enable the plugin on them:
 
 ```
-.settings(AkkaGrpcPlugin.configSettings(IntegrationTest))
+configs(IntegrationTest)
+Defaults.itSettings
+AkkaGrpcPlugin.configSettings(IntegrationTest))
```
raboof, line 30: Good improvement. Perhaps this should be a snippet from a scripted test?

raboof:
OK TO TEST

docs/src/main/paradox/buildtools/sbt.md
bjaglin: Indeed, I was not familiar with the syntax but it looks easy - I'll amend.

bjaglin: cf92528

bjaglin: 2bd5204

Issue #898, opened 2020-04-01
Regression in 0.8.0 - client sends wrong host header


fredfp:
Since switching to 0.8.[012], using akka-grpc as client doesn't work when going through a reverse proxy relying on host header to forward connection to the right grpc server.

The load balancer decodes the requested host as `static` instead of the host given in `GrpcClientSettings.connectToServiceAt("host.lb.com", 80)`.

I think the regression comes from this line in a recent commit https://github.com/akka/akka-grpc/commit/96e15516f6ce78b3ac0530491c1da104952e5710#diff-977da58f66c832d603f6ad9ac553d99dR48 where `settings.serviceName` seems to be the name of the service discovery method (`static`) instead of the targeted hostname.

raboof:
Yes, I just ran into the same issue. As a workaround I *think* you should be able to set the `override-authority` client setting, but we should definitely fix the default here.

raboof:
I _think_ this should be fixed by setting https://github.com/akka/akka-grpc/blob/master/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala#L111 to the host+port, but that needs validating/testing. I'll look into it tomorrow, unless of course someone beats me to it :smile:

fredfp:
I gave it a try setting service-name to host only, not sure about the host+port part you mentioned as it's not part of service-name with other discovery methods.

raboof:
Should be fixed with 0.8.4, feel free to reopen (or open a separate issue) when the lack of a port number causes problems.

pavlo-t:
Lack of port number causes problems with linkerd proxy in front of a service on a custom port after switching to 1.0.1.
For now this seems to work:
```scala
def staticClientSettings(clientName: String): GrpcClientSettings = {
  val config = system.settings.config.getConfig(s"akka.grpc.client.\"$clientName\"")
  val host = config.getString("host")
  val port = config.getInt("port")
  GrpcClientSettings
    .usingServiceDiscovery(
      s"$host:$port",
      new HardcodedServiceDiscovery(Resolved(host, immutable.Seq(ResolvedTarget(host, Some(port), None))))
    )
    .withDefaultPort(port)
}
```
But it would be nice to use `GrpcClientSettings.fromConfig("my-service")` again

fredfp:
@pavlo-t I would create a new issue and reference this one. A PR is always appreciated!

PR #899, opened 2020-04-02
Use correct service-name when using static service-discovery mechanism

fredfp:
references #898

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -106,9 +106,8 @@ object GrpcClientSettings {
       case "static" | "grpc-dns" =>
         val host = clientConfiguration.getString("host")
         require(host.nonEmpty, "host can't be empty when service-discovery-mechanism is set to static or grpc-dns")
-        // Required by the Discovery infrastructure, even when we use static discovery.
-        if (serviceName.isEmpty)
-          serviceName = "static"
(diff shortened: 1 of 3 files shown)


raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -106,9 +106,8 @@ object GrpcClientSettings {
       case "static" | "grpc-dns" =>
         val host = clientConfiguration.getString("host")
         require(host.nonEmpty, "host can't be empty when service-discovery-mechanism is set to static or grpc-dns")
-        // Required by the Discovery infrastructure, even when we use static discovery.
-        if (serviceName.isEmpty)
-          serviceName = "static"
+        // Required by the Discovery infrastructure, set to host as we use static or grpc-dns discovery.
+        serviceName = host
```
raboof, line 110: should we make this `host:port` when the port is non-standard, or does netty take care of that for us?

fredfp: checking...

fredfp: I have no clue, sorry this is deeper than I can dig right now

raboof: :+1: - this is already a good improvement, thanks!

PR #900, opened 2020-04-02
Adapt public API's to accept either typed or classic systems

raboof:
This is part of #761 but does not close the issue entirely

### DIFF
--- a/codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
@@ -13,6 +13,7 @@ import akka.grpc.scaladsl.{ GrpcExceptionHandler, GrpcMarshalling }
 import akka.grpc.Trailers
 
 import akka.actor.ActorSystem
+import akka.actor.ClassicActorSystemProvider
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse, StatusCodes }
 import akka.http.scaladsl.model.Uri.Path
 import akka.http.scaladsl.model.Uri.Path.Segment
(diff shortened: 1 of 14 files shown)


codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -95,7 +96,7 @@ import akka.stream.SystemMaterializer
             }
             case m => scala.concurrent.Future.failed(new NotImplementedError(s"Not implemented: $m"))
           })
-          .recoverWith(GrpcExceptionHandler.from(eHandler(system))(system, writer))
+          .recoverWith(GrpcExceptionHandler.from(eHandler(system.classicSystem))(system, writer))
```
bjaglin, line 99: @raboof if I understand well, users with an akka 2.6.x classpath (in my case brought by Play 2.8.1) need to wait for 2.6.5 to come out with https://github.com/akka/akka/pull/28847 and manually declare it, right?

raboof: Ach, you are right of course. This change in Akka is binary compatible, but not source compatible. Since we're generating sources here that impacts us.

Until 2.6.5 is out we can work around it in Akka gRPC by moving the conversion from the generated code to the akka-grpc-runtime library.

Do you want to or shall I?

fredfp: That's what I experienced this morning, generated code doesn't compile when using akka-grpc 0.8.3 with akka 2.6.4

bjaglin: > Until 2.6.5 is out we can work around it in Akka gRPC by moving the conversion from the generated code to the akka-grpc-runtime library.

Makes sense! I'm AFK today, so I won't get the chance to look at it.

I don't have an urgent need for this, so my question was more related to how getting 2.6.5+, when available, would have been possible without the client adding it manually. But if we move the accessor to the right package, this becomes unnecessary obviously.

raboof: I'll cut a 0.8.4 #905 with #904

PR #904, opened 2020-04-03
Extract classic actor system via helper; improve generated API docs

ennru:
Follow-up on #900 to avoid using Akka 2.6.0-4 internal API.

See https://github.com/akka/akka-grpc/pull/900/files#r402617802

Makes the generated API docs links work in IDEs and Java/scaladoc.


codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -96,7 +99,7 @@ import akka.stream.SystemMaterializer
             }
             case m => scala.concurrent.Future.failed(new NotImplementedError(s"Not implemented: $m"))
           })
-          .recoverWith(GrpcExceptionHandler.from(eHandler(system.classicSystem))(system, writer))
+          .recoverWith(GrpcExceptionHandler.from(eHandler(GrpcExceptionHandler.toClassic(system)))(system, writer))
```
ennru, line 102: This is the important change.

runtime/src/main/scala/akka/grpc/scaladsl/GrpcExceptionHandler.scala
```diff
@@ -21,6 +21,11 @@ object GrpcExceptionHandler {
   private val INVALID_ARGUMENT = Trailers(Status.INVALID_ARGUMENT)
   private val UNIMPLEMENTED = Trailers(Status.UNIMPLEMENTED)
 
+  /**
+   * Helper method to access an Akka internal method (Until Akka 2.6.5).
+   */
+  def toClassic(system: ClassicActorSystemProvider): ActorSystem = system.classicSystem
```
raboof, line 27: Should we mark this helper 'internal'? I'd say when 2.6.5 has been out for a while we can eventually drop it, and specify when on 2.6 you need at least 2.6.5.

ennru: Added a note in the comment.

PR #907, opened 2020-04-03
Fix scala version settings

ignasi35:
(blocked by https://github.com/sbt/sbt/pull/5494 as some regressions seem to have been introduced after `1.3.4` wrt behavior of `+`). 

A simple `+publishLocal` should determine what needs and doesn't need to be published. Then, for the sbt plugin project (`sbt-akka-grpc`) there's a dependency to `sbt-ptotoc` which must be resolved using the correct `scalaBinVersion` (that is, `2.12`).


ignasi35:
Commit 8bb39013691527d2565b327d2975c5b44db48078 should be enough to complete this PR.

Commit 43219121f060e663d2f2f47341c8747ce7d3b014 is going a bit overboard and even forcing `sbt` to `1.3.4` to avoid the `+`regressions mentioned in the description.

raboof:
> Commit 8bb3901 should be enough to complete this PR.

`sbt clean +compile` doesn't work on that commit for me?

> Commit 4321912 is going a bit overboard and even forcing sbt to 1.3.4

I don't think we want to downgrade 1.3.4

ignasi35:
> `sbt clean +compile` doesn't work on that commit for me?

Sorry, I was too ambiguous, let me rephrase:

- This PR has two commits but only Commit 8bb3901 is really necessary. Once the sbt cross-task issue is fixed, we could redo this PR and drop Commit 4321912 since I was just trying out other stuff.

😄

raboof:
So this is now waiting for sbt 1.3.11 (or 1.4.0).

We can then also update .travis.yml to use `+` for publishing again.

build.sbt
```diff
@@ -32,12 +31,16 @@ lazy val codegen = Project(id = akkaGrpcCodegenId, base = file("codegen"))
     mainClass in assembly := Some("akka.grpc.gen.Main"),
     assemblyOption in assembly := (assemblyOption in assembly).value.copy(prependShellScript =
         Some(sbtassembly.AssemblyPlugin.defaultUniversalScript(shebang = true))),
-    crossScalaVersions -= scala213))
+    crossScalaVersions := Dependencies.Versions.CrossScalaForPlugin,
+    scalaVersion := scala212))
```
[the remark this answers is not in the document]
raboof: I think `scalaVersion := scala212` looks nicer to me here - after all, we want it to explicitly be 2.12 since that's what sbt expects. For projects that aren't (part of) an sbt plugin indeed `crossScalaVersions.head` seems nicer

ignasi35:
With `master` already using sbt `1.3.12`, this PR is only doing some code refactor.

I'll mark it ready for review but I'm happy to close it too.

project/Dependencies.scala
```diff
@@ -6,8 +6,16 @@ import buildinfo.BuildInfo
...
 object Dependencies {
   object Versions {
-    val scala212 = "2.12.10"
-    val scala213 = "2.13.1"
+    val scala212 = "2.12.11"
+    val scala213 = "2.13.2"
+
+    // Libs and plugins must share a version. The root project must use that (and set the crossScalaVersions
+    // as empty list)so each sub-project can then decide which scalaVersion and crossCalaVersions they use.
+    val commonScalaVersion = scala212
```
raboof, line 14: Is this still needed/used?

ignasi35: In the PR as it is now it's dead code.

But I just remembered why I added it in the first place: I wanted to add the comment. The rules and restrictions `sbt` uses to decide the default `scalaVersion` were unclear to me when working on this. 

So, for example, the PR now uses `scala212` in a couple of places on `build.sbt` which could be considered as `magic number`'s. If, instead, we used `commonScalaVersion` in `root`'s `scalaVersion := ...` on `build.sbt` it'd be more clear why `root` needs to use scala212.

I'll move the comment to `build.sbt` and remove the `val`.

project/Dependencies.scala
```diff
@@ -6,8 +6,12 @@ import buildinfo.BuildInfo
 
 object Dependencies {
   object Versions {
-    val scala212 = "2.12.10"
-    val scala213 = "2.13.1"
+    val scala212 = "2.12.11"
```
raboof, line 9: small update sneaking in :+1:

ignasi35: Sorry, I forgot to highlight it.

Issue #887, opened 2020-03-25
Filter API


jroper:
There are a number of use cases where a filter API would be useful in Akka gRPC.

Of course, currently you can implement a filter by composing the request handler function with something that works with the Akka HTTP `HttpRequest` and `HttpResponse`. But there's a number of limitations with this:

* If in the filter you want to return an error, eg, because the filter does authentication and the authentication failed, you need to manually encode this as a gRPC status message, by returning a 200 HTTP response and then encoding the gRPC status in a trailer. This is made significantly more complex if you want to support other gRPC encodings, such as grpc-web, or grpc HTTP transcoding, which both encode errors differently, so you would need to detect which is in use for the request and encode manually.
* If you want to do anything with the response, status, for example, if you want a filter that logs requests and want to include the response status, you need to manually decode this from the trailers in the response entity (again, this is further complicated if supporting multiple grpc encodings).
* If you want to do anything with the request or response body (not sure on use cases here, maybe a filter that modifies the messages for the purpose of migrations/backwards compatibility? doesn't seem like a good thing to do...), you have to manually decode/re-encode.

There's a few options for what such a filter API might look like. It could be as simple as:

```
trait GrpcFilter {
  def filter(
    method: MethodDescriptor,
    metadata: Metadata,
    in: Source[ByteString, NotUsed],
    handler: (Metadata, Source[ByteString, NotUsed]) => Future[(Metadata, Source[ByteString, NotUsed])]
  ): Future[(Metadata, Source[ByteString, NotUsed])]
}
```

In the above, it would be assumed that filters could inspect the outgoing status by intercepting errors on the returned `Future` and `Source`, which could either be application specific handled by an application provided exception handler, or an Akka `GrpcException`. It may be nicer though to support unary and streamed calls differently, eg:

```
trait GrpcFilter {
  def filterUnary(
    method: MethodDescriptor,
    metadata: Metadata,
    in: ByteString,
    handler: (Metadata, ByteString) => Future[(Metadata, ByteString)]
  ): Future[(Metadata, ByteString)]
  
  def filterClientStreamed(
    method: MethodDescriptor,
    metadata: Metadata,
    in: Source[ByteString, NotUsed],
    handler: (Metadata, Source[ByteString, NotUsed]) => Future[(Metadata, ByteString)]
  ): Future[(Metadata, ByteString)]
  
  def filterServerStreamed(
    method: MethodDescriptor,
    metadata: Metadata,
    in: ByteString,
[...]

PR #910, opened 2020-04-07
Makes concat public

Marcus-Rosti:
I want to be able to concatenate the Service and the Service Reflection for reuse.

For instance, it's handy to have a class like:
```scala

class GrpcServer(
  grpcPort: Int,
  grpcServices: PartialFunction[HttpRequest, Future[HttpResponse]]*
)(
  override implicit val ec: ExecutionContext,
  override implicit val system: ActorSystem,
  override implicit val materializer: ActorMaterializer
) extends Server {

  override val port: Int = grpcPort

  override val services: Service =
    ServiceHandler.concatOrNotFound(grpcServices: _*).asInstanceOf[Service]

}
```

And then a separate one that does authentication
```Scala
class AuthenticatedGrpcServer(
  grpcPort: Int,
  authenticator: Authenticator,
  authedServices: PartialFunction[HttpRequest, Future[HttpResponse]]*
)(
  override implicit val ec: ExecutionContext,
  override implicit val system: ActorSystem,
  override implicit val materializer: ActorMaterializer
) extends Server {
  override val port: Int = grpcPort

  override val services: Service = Route
    .asyncHandler(
      authenticator
        .authenticationDirective { ctx =>
          ServiceHandler
            .concatOrNotFound(authedServices: _*)(ctx.request)
            .map(RouteResult.Complete)
        }
    )

}
```

So then we can host it like this
```Scala

  val unauthedServer: Server = new GrpcServer(
    grpcPort,
    service1,
    service2,
    service3,
  )

  val authedServer: Server = new AuthenticatedGrpcServer(
    authGrpcPort,
    authenticator,
    service1,
    service3,
  )
```

But that now means I have to pass in reflection separate into each of those services. 
I believe it's easiest to concat them at the top then
```Scala
  val service1 = ServiceHandler.concat(
    Service1.partial(new Service1Impl)),
    ServerReflection.partial(List(Service1))
)
```

### DIFF
--- a/runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
+++ b/runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
@@ -31,7 +31,7 @@ object ServiceHandler {
       handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*): HttpRequest => Future[HttpResponse] =
     concat(handlers: _*).orElse { case _ => notFound }
 
-  private[scaladsl] def concat(handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*)
+  def concat(handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*)
       : PartialFunction[HttpRequest, Future[HttpResponse]] =
     handlers.foldLeft(PartialFunction.empty[HttpRequest, Future[HttpResponse]]) {
       case (acc, pf) => acc.orElse(pf)


lightbend-cla-validator:

raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
```diff
@@ -31,7 +31,7 @@ object ServiceHandler {
       handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*): HttpRequest => Future[HttpResponse] =
     concat(handlers: _*).orElse { case _ => notFound }
 
-  private[scaladsl] def concat(handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*)
+  def concat(handlers: PartialFunction[HttpRequest, Future[HttpResponse]]*)
```
raboof, line 34: I agree it would be useful to open this up, your use case is very sensible.

When we open this up, perhaps we should reconsider whether 'varargs' is the way to go here, or that it would be better to explicitly expect a collection, such as `immutable.Seq`. What do you think?

ignasi35:
The authenticate use case looks more like a filter (see #887).

Marcus-Rosti:
@ignasi35 it most likely is. I think we're probably coming up with convergent solutions

runtime/src/main/scala/akka/grpc/scaladsl/ServiceHandler.scala
Marcus-Rosti: Would you want that to be a change to concatOrNotFound too?

raboof: Ah, makes sense to keep those consistent, so this is :+1:

raboof:
The commit has "Marcus Rosti <marcus.rosti@sony.com>", but that email address doesn't appear to be added to your github account - could you either attach it or change the address?

lightbend-cla-validator:
Hi @Marcus-Rosti,

lightbend-cla-validator:

raboof:
@Marcus-Rosti with which email address would you like this commit to show up? Is that email address already linked to your github user as described at https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user ?

Marcus-Rosti:
Okay I clear it all up @raboof had to just verify some things!

Issue #352, opened 2018-08-15
Document how to setup TLS


chbatey:

raboof:
This may include:
* how to set up TLS in general, https://doc.akka.io/docs/akka-http/current/server-side/server-https-support.html
* how to load the ALPN agent for Akka https://doc.akka.io/docs/akka-http/current/server-side/http2.html#application-layer-protocol-negotiation-alpn-
* how to load the ALPN agent for Play (see also the play scala quickstart https://github.com/playframework/akka-grpc-play-quickstart-scala/blob/master/README.md#running)

ignasi35:
@raboof other than cross-linking to sections of the docs you mentioned, do you think any extra documentation is necessary?

I was thinking that a new section in https://doc.akka.io/docs/akka-grpc/current/deploy.html or https://doc.akka.io/docs/akka-grpc/current/troubleshooting.html could be enough to detail each problem and a brief description of the solution and link to the corresponding existing docs.

ignasi35:
@raboof One of the reasons I'd rather keep this minimal is that both `ConnectionContext.https()` and `UseHttp2.Always` are deprecated so I'd rather keep the contents of these new docs minimal so users find the appropriate ref docs for 10.1 or 10.2.

raboof:
Good point! I agree pointing to the Akka HTTP docs (and perhaps improving those a bit? or are they OK?) should be sufficient. We can point to 10.1 now and update both the dependency and the documentation reference to 10.2.0 when that comes out

ignasi35:
Here's some [sample code](https://github.com/ignasi35/akka-grpc-tls-scala) I used while investigating the work requried for this issue.

ignasi35:
After #912 I did a review of what's already documented to setup TLs on the client-side and found https://doc.akka.io/docs/akka-grpc/current/client/configuration.html contains all the necessary information.

I think we can mark this issue as completed.

esamson:
Sorry to jump in here. I was lead to this issue by this section of the docs: https://doc.akka.io/docs/akka-grpc/current/server/walkthrough.html#serving-the-service-with-akka-http

I'm hoping that's the part that needs updating for this issue to be closed.

ignasi35:
Good catch @esamson! I'll remove that TODO and cross-link the walkthrough to the `deployment.md` where TLS is now documented. Thanks!

PR #921, opened 2020-04-16
[docs TLS] Remove TODO and cross-reference to correct section

ignasi35:
References #352

### DIFF
--- a/docs/src/main/paradox/server/walkthrough.md
+++ b/docs/src/main/paradox/server/walkthrough.md
@@ -198,11 +198,7 @@ In the example this was done from the `main` method, but you could also do this 
 
 @@@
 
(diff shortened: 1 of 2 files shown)


docs/src/main/paradox/server/walkthrough.md
```diff
@@ -200,9 +200,7 @@ In the example this was done from the `main` method, but you could also do this
 
 The above example does not use TLS and is configured to only serve HTTP/2. 
 To allow HTTP and HTTP/2 and gRPC on the same port TLS must be used.
```
raboof, line 202: Wait, this is no longer true... perhaps this whole paragraph needs rewriting? (though linking to the TLS deploy docs from here still seems like a good idea

PR #925, opened 2020-04-20
Make "default" private

ignasi35:
References #761

I think exposing `GrpcExceptionHandler.default` is because:

- `default` is a reserved work in Java8+
- users will probably use `GrpcExceptionHandler.defaultMapper` instead (invoking `MySvcHandlerFactory.create(new MySvcImpl(mat), GrpcExceptionHandler.defaultMapper(), system);`)

If we want to keep it public, we probably need to change the argument type to `ClassicActorSystem.Provider` anyway.


ignasi35:
I setup MiMa in the wrong place.

ignasi35:
Oh, TravisCI get your networking together! (failure running `git clone` and dowloading jabba)

PR #926, opened 2020-04-20
Missing private modifiers

ignasi35:
References #761

A couple of methods annotated as internal and documented as such (but still not `private`). I'm not 100% sure this is the correct fix (or the annotation and docs should be removed instead).

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
@@ -6,6 +6,7 @@ package akka.grpc
 
 import akka.NotUsed
 import akka.annotation.InternalApi
+import akka.annotation.InternalStableApi
 import akka.grpc.GrpcProtocol.{ GrpcProtocolReader, GrpcProtocolWriter }
 import akka.grpc.internal.{ Codec, Codecs, GrpcProtocolNative, GrpcProtocolWeb, GrpcProtocolWebText }
 import akka.http.javadsl.{ model => jmodel }
@@ -33,8 +34,8 @@ trait GrpcProtocol {
    * Constructs a protocol writer for writing gRPC protocol frames for this variant
    * @param codec the compression codec to encode data frame bodies with.
    */
-  @InternalApi
-  private[grpc] def newWriter(codec: Codec): GrpcProtocolWriter
+  @InternalStableApi
+  def newWriter(codec: Codec): GrpcProtocolWriter
 
   /**
    * INTERNAL API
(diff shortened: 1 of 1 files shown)


runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -34,7 +34,7 @@ trait GrpcProtocol {
    * @param codec the compression codec to encode data frame bodies with.
    */
   @InternalApi
-  def newWriter(codec: Codec): GrpcProtocolWriter
+  private[grpc] def newWriter(codec: Codec): GrpcProtocolWriter
```
raboof, line 37: Actually having this public-but-marked-InternalApi came in handy at https://github.com/cloudstateio/cloudstate/pull/211/files#diff-738f4776d7ff0cc32b60cc61f1676218R444 - might be OK to keep it?

ignasi35: Should we annotate as `InternalStableApi` then?

raboof: Good question. The cloudstate proxy is not a library that gets included in downstream projects, but a 'leaf' application. That means there is no reason cloudstate would have a different version of Akka gRPC on its classpath than the one it directly depends on. That means if we make any changes to Akka gRPC, there is no risk of binary incompatibilities in cloudstate: when cloudstate updates to a newer version of Akka gRPC, we can just fix the compile errors in that PR.

So for that reason `InternalStableApi` is less important here. On the other hand, it might still be useful to somehow signal that we don't want to make these API's entirely inaccessible because they have a use case there... I'm thinking perhaps a comment, but I'm not sure what that comment should say :).

ignasi35: Hmmm, I'd rather use the annotation than a comment (on top of the `INTERNAL API`).

I think the cloudstate use case fits quite well the description in [InternalStableApi](https://github.com/akka/akka/blob/master/akka-actor/src/main/java/akka/annotation/InternalStableApi.java).

raboof: OK!

ignasi35: I mean, I read that description and think of `cinnamon` or Lagom and not so much about cross-library compat.

ignasi35:
This is ready to be merged (TravisCI checks are not being updated, but the builds have succeeded).

PR #929, opened 2020-04-21
More InternalApi annotations and apidocs.

ignasi35:
References #761

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
@@ -7,6 +7,7 @@ package akka.grpc.javadsl
 import java.util.concurrent.{ CompletableFuture, CompletionStage }
 
 import akka.annotation.ApiMayChange
+import akka.annotation.InternalApi
 import akka.grpc.scaladsl.{ ServiceHandler => sServiceHandler }
 import akka.http.javadsl.model.{ HttpRequest, HttpResponse, StatusCodes }
 // using japi because bindAndHandleAsync expects that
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -17,9 +18,17 @@ import scala.annotation.varargs
...
+  @InternalApi
   private[javadsl] val notFound: CompletionStage[HttpResponse] =
     CompletableFuture.completedFuture(HttpResponse.create().withStatus(StatusCodes.NOT_FOUND))
 
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
   private[javadsl] val unsupportedMediaType: CompletionStage[HttpResponse] =
     CompletableFuture.completedFuture(HttpResponse.create().withStatus(StatusCodes.UNSUPPORTED_MEDIA_TYPE))
```
ignasi35, line 33: `notFound` and `unsupportedMediaType` are `CompletionStage` specific here because this is `javadsl`. In the `scaladsl` version of `ServiceHandler` there is also a `notFound` and a `unsupportedMediaType` but those are `Future[HttpRespomse]`, ofc. Note I'm only annotating `InternalApi` the `javadsl` copies on purpose.

(off-topic) Each `MyServiceHandlerFactory.{java.scala}` class generated using the [Java](https://github.com/akka/akka-grpc/blob/817aa17258604838ae93f809f52a5aff8fd251a6/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt#L36-L40)/[Scala](https://github.com/akka/akka-grpc/blob/817aa17258604838ae93f809f52a5aff8fd251a6/codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt#L29-L30) `Handler.scala.txt` templates, also has its own pair of `notFound` and `unsupportedMediaType` constants.

This sounds like Akka-gRPC (or even Akka-HTTP) could use a collection of already completed `CS/Future` responses for `notFound` and `unsupportedMediaType` which we could then use from generated code of `ServiceHandler`.

NOTE: the copies in `ServiceHandler` were added in #829.

ignasi35: raised #930

PR #934, opened 2020-04-22
Make Trailers a non-case class

ignasi35:
Prefer non-case classes for public API.

References #761


runtime/src/main/scala/akka/grpc/Trailers.scala
```diff
@@ -11,12 +11,11 @@ import akka.grpc.scaladsl.{ Metadata, MetadataBuilder }
 import akka.grpc.javadsl.{ Metadata => jMetadata }
 
 @ApiMayChange
-case class Trailers(status: Status, metadata: Metadata) {
+class Trailers(val status: Status, val metadata: Metadata) {
   def this(status: Status) {
     this(status, MetadataBuilder.empty)
   }
 
-  @ApiMayChange
```
ignasi35, line 19: The class is already annotated

raboof:
Travis is failing due to missing MiMa filters

runtime/src/main/mima-filters/0.8.4.backwards.excludes/avoid-case-class-on-trailers.backwards.excludes
```diff
@@ -0,0 +1,14 @@
+ProblemFilters.exclude[MissingTypesProblem]("akka.grpc.Trailers")
+ProblemFilters.exclude[DirectMissingMethodProblem]("akka.grpc.Trailers.unapply")
```
raboof, line 2: `ProblemFilters.exclude[DirectMissingMethodProblem]("akka.grpc.Trailers.*")` also would have been fine :)

PR #943, opened 2020-04-30
Avoid shipping classes already in scalapb-runtime

raboof:


sbt-plugin/src/sbt-test/gen-scala-server/01-gen-basic-server/test
```diff
@@ -4,3 +4,7 @@ $ exists target/scala-2.12/src_managed
 $ exists target/scala-2.12/src_managed/main/example/myapp/helloworld/grpc/HelloRequest.scala
 $ exists target/scala-2.12/src_managed/main/example/myapp/helloworld/grpc/HelloworldProto.scala
 $ exists target/scala-2.12/src_managed/main/example/myapp/helloworld/grpc/GreeterService.scala
+
+# This will for example detect when we publish generated code that is
+# already in scalapb-runtime
```
jrudolph, line 9: Because there would be exactly the same files at the same path now clashing?

raboof: Yes - actually not *exactly* the same, scalapb generates additional methods (they enable the `javaConversions` option)

PR #948, opened 2020-05-01
Don't include .proto source into artifacts

raboof:
This was originally added in https://github.com/akka/akka-grpc/pull/149
for dev-mode in play/lagom. Since we don't want .proto definitions to
appear in our output artifacts we have to remove this connection here,
though we might possibly add it back in play-grpc.

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -100,7 +100,6 @@ object AkkaGrpcPlugin extends AutoPlugin {
       (if (config == Compile || config == Test) Seq() // already supported by sbt-protoc by default
        else sbtprotoc.ProtocPlugin.protobufConfigSettings) ++
       Seq(
-        unmanagedResourceDirectories ++= (resourceDirectories in PB.recompile).value,
         watchSources in Defaults.ConfigGlobal ++= (sources in PB.recompile).value,
         akkaGrpcGenerators := {
           generatorsFor(


hepin1989:
I think this is fine as a reference.

raboof:
> I think this is fine as a reference, I suggest to keep them.

I see your point. However, if you are using sbt-assembly, it will fail on you when you are depending on different variations of the same resources. Including the .proto in the artifact makes it likely that you'd see such errors on .proto definitions, which is not very nice..

ignasi35:
Should this change also document the new behavior or edit the edit the [docs](https://doc.akka.io/docs/akka-grpc/current/buildtools/sbt.html#loading-proto-files-from-artifacts)?

I'm still divided about the convenience of this change.

ihostage:
Hmmm.... Does Cloudstate not use that? I don't remember. 😞

ennru:
The docs should how to include the `.proto` files if that is desired.
When using sbt-assembly, you are normally getting good at excluding duplicated files.

viktorklang:
I believe the proper way of publishing protos on the JVM is to include them
in a separate jar with the ”protobuf” scope/qualifier:
https://github.com/cloudstateio/cloudstate/blob/master/build.sbt#L419
-- 
Cheers,
√

raboof:
> I believe the proper way of publishing protos on the JVM is to include them in a separate jar

I think that would be a great approach, but...

> https://github.com/cloudstateio/cloudstate/blob/master/build.sbt#L419

that's not what's happening here: the `"protobuf"` configuration tells sbt-protoc to add the proto files from this artifact to the external protobuf include path (and the `"protobuf-src"` configuration adds them to protobuf source path for which code will be generated). It doesn't download a different jar. See also https://doc.akka.io/docs/akka-grpc/current/buildtools/sbt.html

So there is some precedent for including proto's and generated classes in the same jar, even though IMHO it's problematic :).

raboof:
Given the comments above I'll close this until we have more fully-fledged advice on how to best approach this.

Issue #946, opened 2020-05-01
Multiple test failures using AdoptOpenJDK 8u252 (hs)


ignasi35:
### Versions used 

`master` on commit `52f006d6269ff7867d1abe39e92acb2d6c8486bb`

Using SDKman I installed and setup the 8u252 JDK (`sdk use java 8.0.252.hs-adpt`).

Note that I can't reproduce any of the failures when rolling back to 8u242.

### Expected Behavior

All tests pass when running `sbt test`

### Actual Behavior

There are multiple test failures (consistently):

```
[error] Failed tests:
[error] 	akka.grpc.interop.GrpcInteropAkkaJavaWithAkkaScalaSpec
[error] 	akka.grpc.interop.GrpcInteropIoWithAkkaJavaSpec
[error] 	akka.grpc.interop.GrpcInteropAkkaJavaWithAkkaJavaSpec
[error] 	akka.grpc.interop.GrpcInteropAkkaScalaWithAkkaScalaSpec
[error] 	akka.grpc.interop.GrpcInteropAkkaScalaWithAkkaJavaSpec
[error] 	akka.grpc.interop.GrpcInteropIoWithAkkaScalaSpec
```

ignasi35:
Following is a test result matrix with more information.

`GrpcInteropIoWithAkkaJavaSpec` is represented as `Server: Io / Client: Java` so the information in the description of the issue is translated into the `master` column of the matrix. `master` actually represents the commit with SHA-1 52f006d6.

The variants `master` + `agent <version>` represent changing only the JettyAgent version in the [project extensions](https://github.com/akka/akka-grpc/blob/master/project/ProjectExtensions.scala#L18).

The variant `master`+"no agent" requires adding the setting `Test / fork := true` so interop tests run forked. The reason is that the `JavaAgent` plugin will fork tests under the covers but [only if there's an agent in the classpath](https://github.com/sbt/sbt-javaagent/blob/master/src/main/scala/com/lightbend/sbt/javaagent/JavaAgent.scala#L81).

|Server|Client|`master`|`master` +  agent `2.0.10`| `master` + no agent|
|-|-|-|-|-|
|IO|IO|✅|✅|✅|
|IO|java|🔴|🔴|🔴|
|IO|Scala|🔴|🔴|🔴|
|java|IO|✅|✅|🔴|
|java|java|🔴|🔴|🔴|
|java|Scala|🔴|🔴|🔴|
|Scala|IO|✅|✅|🔴|
[...]

PR #964, opened 2020-05-11
Support JDK 1.8.0_252

raboof:
References #946

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -12,17 +12,9 @@ import akka.grpc.internal.HardcodedServiceDiscovery
 import akka.util.Helpers
 import akka.util.JavaDurationConverters._
 import com.typesafe.config.{ Config, ConfigValueFactory }
-import com.typesafe.sslconfig.akka.util.AkkaLoggerFactory
-import com.typesafe.sslconfig.ssl.{
-  ConfigSSLContextBuilder,
-  DefaultKeyManagerFactoryWrapper,
-  DefaultTrustManagerFactoryWrapper,
-  SSLConfigFactory,
-  SSLConfigSettings
-}
 import io.grpc.CallCredentials
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
-import javax.net.ssl.SSLContext
+import javax.net.ssl.TrustManager
 
 import scala.collection.immutable
 import scala.concurrent.duration.{ Duration, _ }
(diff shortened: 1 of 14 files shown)


runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -52,8 +49,11 @@ object NettyClientUtils {
     if (!settings.useTls)
       builder = builder.usePlaintext()
     else {
-      builder = settings.sslContext
-        .map(javaCtx => builder.negotiationType(NegotiationType.TLS).sslContext(nettyHttp2SslContext(javaCtx)))
+      builder = settings.trustManager
+        .map(trustManager => {
+          val nettySslContext = GrpcSslContexts.forClient().trustManager(trustManager).build()
```
ignasi35, line 54: Then, we'd also invoke `.keyManager` on the builder here...

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -92,30 +92,6 @@ object NettyClientUtils {
...
-    val nettySslContext: JdkSslContext =
-      GrpcSslContexts.configure(GrpcSslContexts.forClient, SslProvider.JDK).build.asInstanceOf[JdkSslContext]
-
-    // Patch the SSLContext value inside the JdkSslContext object
-    val nettySslContextField: Field = classOf[JdkSslContext].getDeclaredField("sslContext")
-    nettySslContextField.setAccessible(true)
-    nettySslContextField.set(nettySslContext, javaSslContext)
-
-    nettySslContext
-  }
```
ignasi35, line 117: So good to get rid of this!

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -164,32 +157,6 @@ object GrpcClientSettings {
...
-  @InternalApi
-  private def getSSLContext(config: Config)(implicit actorSystem: ActorSystem): SSLContext = {
-    val sslConfigSettings: SSLConfigSettings = SSLConfigFactory.parse(config)
-    val sslContext: SSLContext = new ConfigSSLContextBuilder(
-      new AkkaLoggerFactory(actorSystem),
-      sslConfigSettings,
-      new DefaultKeyManagerFactoryWrapper(sslConfigSettings.keyManagerConfig.algorithm),
-      new DefaultTrustManagerFactoryWrapper(sslConfigSettings.trustManagerConfig.algorithm)).build()
-    sslContext
-  }
```
ignasi35, line 181: Maybe we can still use this code but, instead of returning `SSLContext` return a `trustManager` and a `keyManager` build using `ssl-config`'s `Default{Key,Trust}ManagerFactoryWrapper`.

ignasi35:
@raboof WDYT about `keyManagers` so we support mTLS? Worth adding it here or should we just merge as is (LGTM) and eventually implement support for client-side TLS when the need arises?

raboof:
> @raboof WDYT about `keyManagers` so we support mTLS? Worth adding it here or should we just merge as is (LGTM) and eventually implement support for client-side TLS when the need arises?

Sounds useful to add, but let's merge this first

PR #970, opened 2020-05-14
Don't use github milestones

raboof:
The last few releases using release-drafter to have an overview of the changes
in a release worked pretty well, and avoided the overhead of using a milestone.
Maybe let's not require using milestones anymore?

### DIFF
--- a/docs/release-train-issue-template.md
+++ b/docs/release-train-issue-template.md
@@ -14,16 +14,12 @@ Variables to be expanded in this template:
 
 ### before the release
 
-- [ ] Create a new milestone for the [next version](https://github.com/akka/akka-grpc/milestones)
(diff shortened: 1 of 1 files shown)


ihostage:
For me (as external watcher) milestone is more about the road map and planning of release scope than about release notes.
Release Drafter not describe Akka Team plans. 😄

raboof:
Right, and we can still use them for that: for example we have marked some issues as '1.0.0' that we want/plan to resolve before releasing 1.0.0.

That is relatively uncommon though: typically, we would use a 'pick next' label for things we'd like to consider for the 'next' release, and not plan ahead further than that.

Usually we assign milestones only after-the-fact when merging the PR, which is convenient for producing release notes, but now that we have release-drafter that no longer seems so imporant.

ihostage:
> Right, and we can still use them for that: for example we have marked some issues as '1.0.0' that we want/plan to resolve before releasing 1.0.0.

Exactly what I need 👌 
> Usually we assign milestones only after-the-fact when merging the PR, which is convenient for producing release notes, but now that we have release-drafter that no longer seems so imporant.

Agree 👍

PR #973, opened 2020-05-15
Allow ssl-provider to be configured

raboof:


runtime/src/main/resources/reference.conf
```diff
@@ -30,6 +30,9 @@ akka.grpc.client."*" {
   # Location on the classpath of CA PEM to trust
   trusted = ""
   use-tls = true
+  # SSL provider to use:
+  # leave empty to auto-detect, or configure 'jdk' or 'openssl'.
+  ssl-provider = ""
```
ignasi35, line 35: Add `openssl_refcnt` on the comment? Not sure, TBH, the internal implementation is identical to `openssl`.

Issue #978, opened 2020-05-21
Restore withSSLContext in client settings


ignasi35:
In #964 we refactorred `withSSLContext` to `withTrustManager` in `GrpcClientSettings` to allow the underlying implementation (Netty/grpc-java) to choose a different implementation (BoringSSL, OpenSSL, JDK).

As a consequence of removing `withSSLContext` usages where only an `SSLContext` is available (and not the `TrustManager`) are locked out.

The motivation for the change in #964 was to avoid the lock to the JDK SSL provider unblocking the usage of JDK8u252. Bringing back `withSSLContext` next to `withTrustManager` means that users may:

 - build a `GrpcClientSettings` for PLAINTEXT connections
 - build a `GrpcClientSettings` providing only a `TrustManager` (using BoringSSL/OpenSSL) and execute on any JDK8 or JDK11 
 - build a `GrpcClientSettings` providing only a `SSLContext` (using JDK Provider) and execute  on JDK8 < u250 or JDK11

PR #979, opened 2020-05-22
Re-introduce (but discourage) withSslContext

raboof:
Refs #978

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
@@ -5,6 +5,7 @@
 package akka.grpc.internal
 
 import java.util.concurrent.TimeUnit
+import javax.net.ssl.SSLContext
 
 import akka.Done
 import akka.annotation.InternalApi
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -51,20 +52,25 @@ object NettyClientUtils {
...
+              val context = provider match {
+                case None =>
+                  GrpcSslContexts.configure(SslContextBuilder.forClient())
+                case Some(sslProvider) =>
+                  GrpcSslContexts.configure(SslContextBuilder.forClient(), sslProvider)
+              }
+              builder.sslContext((tm match {
+                case None               => context
+                case Some(trustManager) => context.trustManager(trustManager)
+              }).build())
```
ignasi35, line 72: Hm, this gives precedence to `SSLContext` over `TrustManager`. I _naïvely_ expected the opposite would be preferred but I have no argument backing the idea.

raboof: I'll throw an error when both are set.

Issue #253, opened 2018-06-27
Expected generated sources directory incorrect


johanandren:
Again, supposed to be `target/scala-2.xy/src_managed/{test|main}/a/b/c` but `main` part is missing now, so IntelliJ doesn't find the generated classes out of the box anymore.

johanandren:
Ups, sorry, I take that back, was a leftover of some kind

PR #990, opened 2020-06-02
Force jdk8u242 in travis (instead of latest)

ignasi35:
References #253


ignasi35:
This change was meant for https://github.com/playframework/play-grpc 🤦🏼‍♂️

Issue #1018, opened 2020-06-16
Snapshot versions publish unreachable gradle plugins.


ignasi35:
https://github.com/gradle/gradle/issues/13445

ennru:
Fixed with #1019

PR #1019, opened 2020-06-16
Avoid using + in version

ignasi35:
References #1018

Will produce the version id: `1.0.0-M1-40-d1e58d73`


build.sbt
```diff
@@ -9,6 +9,9 @@ val akkaGrpcRuntimeName = "akka-grpc-runtime"
 
 lazy val mkBatAssemblyTask = taskKey[File]("Create a Windows bat assembly")
 
+// gradle plugin infrast
```
ennru, line 12:
```suggestion
// gradle plugin compatibility (avoid `+` in snapshot versions)
```

PR #1021, opened 2020-06-17
Mark akka-grpc 1.0.0 as 'Supported'

raboof:
Let's update the README after we actually released 1.0.0


raboof:
> overview.md

Thx, missed that one

> README.md

Yeah, not necessary to avoid a 'race condition' there :smile: , updated.

Issue #1024, opened 2020-06-17
Gradle plugin support for scala 2.13


ignasi35:
The Akka-gRPC plugin for Gradle uses a hardcoded [artifact](https://github.com/akka/akka-grpc/blob/master/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L143). Because of that users building their project using gradle can only use scala 2.12.

There are two options:

- release different plugins for each scala major akka-grpc supports
- remove the [automatic dependency management](https://github.com/akka/akka-grpc/blob/master/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L143) from the gradle plugin so users need to add the plugin _and_ add the dependency to `akka-grpc-runtime_2.xx` 

PS. Note the gradle plugin [uses other artifacts](https://github.com/akka/akka-grpc/blob/master/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L52-L61) hardcoded to 2.12. That's fine because those are only used at source-generation time.

eshepelyuk:
@ignasi35 @raboof - got a comment about #1025 

Althought proposed patch opens the possibility to use Scala 2.13, but it also makes a plugin application more complicated, i..e. forces users to explicitly declare akka grpc runtime dependency.

While plugin should simplify such things, declare dependencies itself, without forcing users to care about exact dependencies and their application via `build.gradle`.

I propose a different approach to unblock Scala 2.13 usage - to introduce a configuration parameter that will be holding versino of Scala, it could be overwritten and it will have default value.

So, `build.gradle` could look like 

```
akkaGrpc {
  scalaVersion = '2.13'
}
```

By default value of `scalaVersion` could be `2.12` - so this change will not introduce compatibility break for existing users of the plugin.

This is quite common approach in Gradle plugins working with Scala libraries. 
I did it for my own plugin and I'm willing to contribute the code with the test if there's no objections from your side.

ignasi35:
This is a great idea. Would you like to work on it? I am fine dropping the exploratory work on the PR.

eshepelyuk:
@ignasi35 sure I do , as mentioned above i will contribute this code.
Maybe during a weekend.

eshepelyuk:
@raboof plz feedback :)
I am also going to introduce tests for my functionality 
[...]

PR #1025, opened 2020-06-17
Support 2.13 in gradle

ignasi35:
References #1024

This approach removes the automatic addition of `akka-grpc-runtime` as a compile dependency on user projects that include the akka-grpc plugin. Instead, users will need to add the dependency to runtime explicitly:

```
plugins {
  id "com.lightbend.akka.grpc.gradle" version "${akkaGrpcVersion}"
}

dependencies {
  implementation 'com.lightbend.akka.grpc:akka-grpc-runtime_${scalaVersion}:${akkaGrpcVersion}'
}
```

### DIFF
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -140,7 +140,7 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
     @Override
     void beforeResolve(ResolvableDependencies resolvableDependencies) {
         def compileDeps = project.getConfigurations().getByName("compile").getDependencies()
-        compileDeps.add(project.getDependencies().create("com.lightbend.akka.grpc:akka-grpc-runtime_2.12:${pluginVersion}"))
+//        compileDeps.add(project.getDependencies().create("com.lightbend.akka.grpc:akka-grpc-runtime_2.13:${pluginVersion}"))
         // TODO #115 grpc-stub is only needed for the client. Can we use the 'suggestedDependencies' somehow?
         compileDeps.add(project.getDependencies().create("io.grpc:grpc-stub:${grpcVersion}"))
         project.gradle.removeListener(this)
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -6,6 +6,7 @@ import org.apache.commons.lang.SystemUtils
 import org.gradle.api.GradleException
 import org.gradle.api.Plugin
 import org.gradle.api.Project
+import org.gradle.api.artifacts.Dependency
 import org.gradle.api.artifacts.DependencyResolutionListener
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
```diff
@@ -140,7 +141,9 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
     @Override
     void beforeResolve(ResolvableDependencies resolvableDependencies) {
         def compileDeps = project.getConfigurations().getByName("compile").getDependencies()
-        compileDeps.add(project.getDependencies().create("com.lightbend.akka.grpc:akka-grpc-runtime_2.12:${pluginVersion}"))
+        boolean foundRuntimeDep = compileDeps.any { dep -> dep.group == "com.lightbend.akka.grpc" && dep.group.startsWith("akka-grpc-runtime") && dep.version == pluginVersion }
+        if(!foundRuntimeDep)
+            throw new RuntimeException("""Missing dependency:   implementation 'com.lightbend.akka.grpc:akka-grpc-runtime_\${scalaVersion}:${pluginVersion}' """)
```
ignasi35, line 146: I've tested the removal of the dependency but I haven't tested the verification + exception.

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
```diff
@@ -140,7 +141,9 @@ class AkkaGrpcPlugin implements Plugin<Project>, DependencyResolutionListener {
     @Override
     void beforeResolve(ResolvableDependencies resolvableDependencies) {
         def compileDeps = project.getConfigurations().getByName("compile").getDependencies()
-        compileDeps.add(project.getDependencies().create("com.lightbend.akka.grpc:akka-grpc-runtime_2.12:${pluginVersion}"))
+        boolean foundRuntimeDep = compileDeps.any { dep -> dep.group == "com.lightbend.akka.grpc" && dep.group.startsWith("akka-grpc-runtime") && dep.version == pluginVersion }
```
raboof, line 144: I wonder if we could find the `scala-library` dependency in a similar way and add the right runtime based on that?

ignasi35: Yeah. On my way to the office I was thinking maybe adding an argument to the plugin so things are a bit less magic. 

I'm a bit unsure about how the scala version is set on a gradle build. Unlike `sbt` in the [gradle plugin for scala](https://docs.gradle.org/current/userguide/scala_plugin.html) there doesn't seem to be a need (or even a mechanism?) to explicitly set the desired scala version. There is a recommendation to add the dependency to `'org.scala-lang:scala-library:2.x.y'` but I have a pet project without the explicit dep that compiles and runs tests successfully.

eshepelyuk:
@raboof @ignasi35 I left a comment with a slightly different approach to solve the issue in #1024 
Could you please take a look ?

eshepelyuk:
#1037 is opened for preview - I thinks this can be closed for now.

ignasi35:
> #1037 is opened for preview - I thinks this can be closed for now.

Sure thing! Thanks @eshepelyuk

PR #1032, opened 2020-06-22
Update to ScalaPB 0.10.6 and support running in a sandboxed classloader.

thesamet:
This change uses a new functionality of protocbridge which enables
code generators to be loaded via a sandboxed classloader. This eliminates the
possibility of binary compatibility problems with SBT, protobuf-java or ScalaPB
itself. This would also enable moving forward to protobuf-java 3.12.x as it is
incompatible with the version of protobuf-java presently shipped with SBT.


raboof:
OK TO TEST

Issue #1050, opened 2020-07-06
Snapshots are uploaded with timestamp


raboof:
https://travis-ci.org/github/akka/akka-grpc/jobs/705323771 uploads snapshots as `1.0.0-17-3dcade13-20200706-0920` - apparently the directory has become dirty.

raboof:
akka-grpc-runtime_2.12 is still fine but akka-grpc-runtime_2.13 was published with the timestamp

ignasi35:
So this was a side-effect of #1049? How did that happen? (we check the scalafmt on PRs 🤔)

raboof:
I guess 2 #1049 was merged in parallel with another PR

PR #1051, opened 2020-07-06
Add a 'git status' to diagnose the dirty repo

raboof:
To diagnose #1050


ignasi35:
This is not meant to be merged, is it? I mean, will the job where the issue manifests already run in the PR?

raboof:
> This is not meant to be merged, is it? I mean, will the job where the issue manifests already run in the PR?

We only publish from master, so this will have to be merged to see the diagnostics

ignasi35:
> > This is not meant to be merged, is it? I mean, will the job where the issue manifests already run in the PR?
> 
> We only publish from master, so this will have to be merged to see the diagnostics

How abou hacking the conditions so the PR build runs a `publishLocal` could save us the extra noise in `master`? Maybe too much work and potential for a false positive.

Merging.

ignasi35:
> Merging.

... when green. 😅

Issue #289, opened 2018-07-12
Empty.java in Gradle Scala project


patriknw:
For unknown reason the Gradle complains about no java sources. The workaround was to add an `Empty.java` file in src/main/java, but would be nice to make it work without that.

Should also be fixed in Quickstarts

style95:
Any updates on this issue?
I want to remove annoying `Empty.java` file from my project.

raboof:
No - I'm not very familiar with Gradle so I'd be very interested if someone can shed light on this!

eshepelyuk:
This is side-effect of protobuf plugin. Probably could be hacked. Will try to figure out.

raboof:
Thanks again!

PR #1056, opened 2020-07-07
Disable Java compilation when a Gradle project only has Scala sources

eshepelyuk:
References #289 

@raboof another fix :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ jobs:
       name: Test Maven Scala
     - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
       name: Test Gradle Java
-    - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd plugin-tester-scala && ./gradlew clean test --info --stacktrace -DakkaGrpcTest.pluginVersion=`git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8` -PakkaGrpcTest.pluginVersion=`git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8`
+    - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd gradle-plugin && ./gradlew clean publishToMavenLocal --info --stacktrace && cd ../plugin-tester-scala && ./gradlew clean test --info --stacktrace -PakkaGrpcTest.pluginVersion=`git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8`
       name: Test Gradle Scala
     - script: cd benchmark-java && sbt -jvm-opts .jvmopts-travis test:compile
       name: Compile benchmarks
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ jobs:
       name: Test Maven Scala
     - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
       name: Test Gradle Java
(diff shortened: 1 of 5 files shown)


raboof:
OK TO TEST

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
```diff
@@ -125,6 +125,10 @@ class AkkaGrpcPlugin implements Plugin<Project> {
         }
 
         project.afterEvaluate { Project p ->
+            if (akkaGrpcExt.scala) {
+                p.compileJava.enabled = false
```
raboof, line 129: Aaah interesting. I guess this would interfere with hybrid java/scala projects - but I'm not sure how well-supported those are in Gradle in the first place?

eshepelyuk: Yeah, you're right ! Probably I had to do a better, real fix and not that stupid as above

eshepelyuk: Well, the possibility of having mixed scala \ java project also have to be handled in plugin itself.
Since now there're different options are set to the protoc compiler.
What do you think ?

raboof: I think for maven/scala you're allowed to mix java and scala, but you can only generate either java or scala gRPC infrastructure. Would that be possible for gradle as well?

eshepelyuk: Well, not sure what do you mean. It's 100% one can setup mixed java \ scala project in Gradle and classes will be properly compiled.

raboof: Right - but with this change such a project would no longer compile the java classes, right? Or am I misunderstanding that?

Perhaps the condition could be changed so compileJava is only disabled when there are no java files?

eshepelyuk: I am working on the fix right now.
But I asked smeth completely different  aboout mixed projects.

raboof:
Hmm, the travis gradle plugin-tester-scala run seems to trip over this change now: https://travis-ci.org/github/akka/akka-grpc/jobs/705808720#L934

eshepelyuk:
> Hmm, the travis gradle plugin-tester-scala run seems to trip over this change now: https://travis-ci.org/github/akka/akka-grpc/jobs/705808720#L934

Yes I am trying to figure it out :((

.travis.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       name: Test Maven Scala
     - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd plugin-tester-java && ./gradlew --include-build ../gradle-plugin clean compileJava
       name: Test Gradle Java
-    - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd plugin-tester-scala && ./gradlew clean test -DakkaGrpcTest.pluginVersion=`git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8`
+    - script: sbt -jvm-opts .jvmopts-travis publishM2 && cd gradle-plugin && ./gradlew clean publishToMavenLocal --info --stacktrace && cd ../plugin-tester-scala && ./gradlew clean test --info --stacktrace -PakkaGrpcTest.pluginVersion=`git describe --tags | sed -e "s/v\(.*\)-\([0-9][0-9]*\).*/\\1-\\2-/"``git rev-parse HEAD | head -c8`
```
raboof, line 40: wow, weird that we didn't notice this before! Should we also update the Gradle Java test?

eshepelyuk: @raboof I'll be working on this - to remove a need for `publishToMavenLocal`. Will try to leverage composite build with `--include-build` flag. Plz need more time on this.

raboof: No rush! Thanks!

PR #1058, opened 2020-07-07
Fix some deprecation warnings

ennru:


ennru:
> Shouldn't those warnings be fatal?

Not in Java code.

ennru:
Travis is happy, but didn't tell anyone.

PR #1059, opened 2020-07-08
Avoid generating duplicate imports in scala clients

### LINKED ISSUE #1009: Unused imports in generated code
The Scala client code generation will import all `akka.grpc.scaladsl.*RequestBuilder`, even if they are not all needed. Would be nicer to generate only the imports that are actually used.

raboof:
fixes #1009


codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -34,7 +34,7 @@ import akka.grpc.internal.ClientState
     }
 
     statements
-  }.mkString("\n")
+  }.distinct.mkString("\n")
```
ennru, line 37: :-)

PR #1063, opened 2020-07-09
chore: rework ci

eshepelyuk:
- use workspaces to share compiled sources and published plugins
- rearrange stages
- unify java and scala gradle plugin testing

@raboof could you please check ? this is a follow up to #289 

References #289


raboof:
OK TO TEST

eshepelyuk:
Seems Travis CI build is not starting :(

raboof:
> Seems Travis CI build is not starting :(

Not sure what's going on there, seeing it on other repo's as well...

eshepelyuk:
@raboof got a question. From logs of jobs I see that `sbt test` causes compilation and running tests from `plugin-tester-java` and `plugin-tester-scala`.
Is it expected behaviour ?

raboof:
> From logs of jobs I see that sbt test causes compilation and running tests from plugin-tester-java and plugin-tester-scala. Is it expected behaviour ?

Yes, those subprojects are ran from the parent project for sbt, and 'independently' for gradle and maven.

.travis.yml
```diff
@@ -25,39 +25,76 @@ env:
 before_install: curl -Ls https://git.io/jabba | bash && . ~/.jabba/jabba.sh
 install: jabba install "$TRAVIS_JDK" && jabba use "$_" && java -Xmx32m -version
 
+stages:
+  - name: prepare
+    if: NOT tag =~ ^v
+  - name: test
```
raboof, line 31: Splitting this in 'prepare' and 'test' makes travis a bit slower: it means the 12-minute 'sbt scripted' job will not start before the 5-minute 'prepare' phase is done.

I'm OK with that though, we can always optimize further later.

eshepelyuk: I will put `scripted` to preparte

eshepelyuk:
> > From logs of jobs I see that sbt test causes compilation and running tests from plugin-tester-java and plugin-tester-scala. Is it expected behaviour ?
> 
> Yes, those subprojects are ran from the parent project for sbt, and 'independently' for gradle and maven.

Could you explain further ? isn't it a slowdown for the build ? 
Those repos are not required for compilation and pubishing and for tests, why are they stil lincluded ?
I think they should excluded from `sbt test` and only be used in Maven \ Gradle tests itself.

Am I missing smth ?

eshepelyuk:
@raboof I'd also want to introduce Gradle plugin testing both for Scala 2.12 and 2.13 - is it OK to do within this PR ?
Or you prefer another one ?

raboof:
> Could you explain further ? isn't it a slowdown for the build ? Those repos are not required for compilation and pubishing and for tests, why are they stil lincluded ?

While it does take *some* time, they test some of the sbt plugin code generation, which is useful. You're right that this has some overlap with the 'scripted' tests, but especially during development it's worth having early feedback here I think.

raboof:
> I'd also want to introduce Gradle plugin testing both for Scala 2.12 and 2.13 - is it OK to do within this PR ? Or you prefer another one ?

Whatever you prefer :+1:

eshepelyuk:
> While it does take _some_ time, they test some of the sbt plugin code generation, which is useful. You're right that this has some overlap with the 'scripted' tests, but especially during development it's worth having early feedback here I think.

Does it mean, that those repos are used to _test sbt plugin_ as well ?

raboof:
> Does it mean, that those repos are used to test sbt plugin as well ?

Correct!

eshepelyuk:
> > Does it mean, that those repos are used to test sbt plugin as well ?
> 
> Correct!

Thanks for clarification.

.travis.yml
eshepelyuk: Hmm, it seems not that good idea now, sinc tests will wait for scripted job.
Is there a chance to split `scripted` job into several smaller jobs ?

eshepelyuk: Another option for `scripted` job will be to use pre compiled classes saved from previous stage
does `scripted` use a single scala and JDK version ( according to logs - i suppose, yes)

eshepelyuk:
@raboof I'm done with this PR, tried to optimize duration as much as possible.

raboof:
(LGTM but not merging since still marked 'draft')

eshepelyuk:
@raboof I still have some question regarding the build, could you please clarify.

1. Above, you'd written that `plugin-tester-java` and `plugin-tester-scala` are used for testing Sbt plugin itself. But as far as I see `sbt-plugin` only uses scripted tests and those tests has own sources. So, could you point out exact place where those testar repositories are used for the tests different from Gradle and Maven. Supposedly those repos could be excluded from compilation and this can improve the overall build speed.

1. Another question is about `interop-tests` - those tests are definately used for `scripted` tests, but are those sourcess needed for other projects ? Since this repo take a quite long to compile, if this repo is not needed for `sbt test` - it should be excluded and this also improve the overall duration of the build.

raboof:
> Above, you'd written that `plugin-tester-java` and `plugin-tester-scala` are used for testing Sbt plugin itself. But as far as I see sbt-plugin only uses scripted tests and those tests has own sources. So, could you point out exact place where those testar repositories are used for the tests different from Gradle and Maven

During a regular 'sbt test' the code generation is triggered for these subprojects via `ReflectiveCodeGen` (https://github.com/akka/akka-grpc/blob/master/build.sbt#L208).

> Another question is about `interop-tests` - those tests are definately used for `scripted` tests, but are those sourcess needed for other projects ? Since this repo take a quite long to compile, if this repo is not needed for sbt test - it should be excluded and this also improve the overall duration of the build

You are correct that some of these tests are invoked twice: once while testing the `interop-test` project during a regular `sbt test`, and once via the scripted tests.

Both have their advantages: the regular 'sbt test' tests have a much quicker turnaround time, so during development those are much more useful. However, they do rely on the `ReflectiveCodeGen` mentioned above - which is useful, but not as realistic as a scripted test.

I think it's a reasonable idea if we skipped those duplicated tests in `sbt test` on travis, since they're executed as a `scripted` test anyway. I think there's some additional tests in `interop-tests` that aren't part of the `scripted` test, though. A first step might be to keep compiling `interop-tests` and just skip the 'duplicated' tests. As a follow-up we could port the other tests to the `runtime` project and disable the `interop-tests` subproject entirely when doing a `sbt test` on travis.

eshepelyuk:
1. Regarind plugin-tester-java and plugin-tester-scala.
what if we introduce a new stage that will only invoke code generation via sbt for those projects, and just simply exclude them from mainline `sbt compilation` ?

2. regarding `interop-tests` - I need your help on this, since by sbt knowledge is almost zero.

raboof:
> what if we introduce a new stage that will only invoke code generation via sbt for those projects, and just simply exclude them from mainline sbt compilation?

I'm not sure which step that would be intended to improve?

> regarding `interop-tests` - I need your help on this, since by sbt knowledge is almost zero.

On second thought, having those part of the regular tests also makes sure they're ran both on all combinations of JDK8/JDK11 and Scala2.12/Scala2.13, so perhaps it's still useful to just keep them?

eshepelyuk:
> > what if we introduce a new stage that will only invoke code generation via sbt for those projects, and just simply exclude them from mainline sbt compilation?
> 
> I'm not sure which step that would be intended to improve?

it should improve speed of 

* 4 steps calling `sbt test`  ( jdk8, jdk11, scala 2.12, scala 2.13 combination)
* 2 scripted tests steps
* "compile benchmarks" step



> > regarding `interop-tests` - I need your help on this, since by sbt knowledge is almost zero.
> 
> On second thought, having those part of the regular tests also makes sure they're ran both on all combinations of JDK8/JDK11 and Scala2.12/Scala2.13, so perhaps it's still useful to just keep them?

As for me personally, I'd rather create 4 dedicated steps for that, than keeping it inside `sbt test`. But it's ony my personal view. Current behaviour is quite fine now.

PR #1065, opened 2020-07-13
Remove the unused AkkaGrpcClientFactory

ignasi35:
A leftover from the pre-1.0.0 cleanup.


runtime/src/main/scala/akka/grpc/internal/AkkaGrpcClientFactory.scala
```diff
@@ -4,24 +4,25 @@
...
 
 object AkkaGrpcClientFactory {
-  def create[T <: AkkaGrpcClient: ClassTag](
-      settings: GrpcClientSettings)(implicit mat: Materializer, ex: ExecutionContext): T = {
+  def create[T <: AkkaGrpcClient: ClassTag](settings: GrpcClientSettings)(
+      implicit sys: ClassicActorSystemProvider): T = {
     // this reflection requires:
     //    object @{service.name}Client {
-    //      def apply(GrpcClientSettings)(Materializer, ExecutionContext): @{service.name}Client
+    //      def apply(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider): @{service.name}Client
```
ignasi35, line 19: When changing the codegen we didn't change the reflective code here.

This code is no longer exercised in this repo, when we split the code for play and lagom away from this repo we left this orphan code here. I believe this is the right repo for this `AkkaGrpcClientFactory, but not untested.

ignasi35: PS: I'm writing the test now. Don't merge yet :-)

raboof: > I believe this is the right repo for this

I suspect the reason is that you'd want to avoid accidentally breaking compatibility between Akka gRPC and Play/Lagom gRPC.

There are generally 2 ways to achieve that:
* Move this code to play-grpc, and make sure the assumptions in this code are upheld in Akka gRPC
* Keep this code here and keep this code binary compatible for Play/Lagom gRPC's sake.

The first option seems 'cleaner' since it's a bit strange to have non-public API that is not used in Akka gRPC itself but is relied upon in Play/Lagom. On the other hand, the second option has the advantage that it gives us the option to change the structure of the generated code, as long as we evolve this code along with it.

Was this indeed the reasoning? If so, I think we should double-check that the assumptions in this code aren't already set in stone by https://doc.akka.io/docs/akka-grpc/current/binary-compatibility.html . If so, then I think that's a strong case for option '1'. If not, then indeed I'm fine with option '2'.

ignasi35: Good point.

Actually, it makes a lot more sense to move the `AkkaGrpcClientFactory` to `play-grpc` it's an implementation detail of the testing tools there. It's a convenience tool that we can evolve there each time the `akka-grpc` dependency is bumped.

This class is in an internal package (but we didn't mark it `@InternalApi`) so we can remove it.

I'll redo this PR to remove the class and move it to `play-grpc`.

Issue #1074, opened 2020-07-17
Gradle plugin should add maven repositories.


eshepelyuk:
### Short description

Gradle plugin should add mavenCentral repository when applied.

### Details

Since plugin need some repository to be configured to resolve dependencies, it is valid if plugin will add those repositories itself, without need for a end user to explicitely configure repositories in `build.gradle`.

When plugin detects that SNAPSHOT version is used, it should add akka snapshots repository as well.

eshepelyuk:
@raboof i am going to contribute this.

PR #1072, opened 2020-07-16
Upgrade protobuf-gradle-plugin to 0.8.12

raboof:
https://github.com/google/protobuf-gradle-plugin/releases/tag/v0.8.12

PR #1075, opened 2020-07-17
Gradle plugin should add maven repositories.

eshepelyuk:
References #1074


raboof:
OK TO TEST

eshepelyuk:
@raboof Maybe it worth to wait for #1072 to be merged and then rebase this PR ?

vincentye38:
In here https://github.com/akka/akka-grpc/blob/c26005397a311f1568002205f8d395ed38639ac7/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L44-L51 , should mavenContral() being inside the if statement. It adds mavenContral to the project that uses this plugin. In my case, I can't use mavenContral within my company.

PR #1069, opened 2020-07-15
fix: scala version autodetect

eshepelyuk:
References #1067

Issue #1067, opened 2020-07-15
Gradle Scala autodetect doesn't support non-scala package names with underscores


pmitura-lucid:
Hi, I've experienced a Gradle plugin issue on the latest revision (`1.0.0-33-7465b18b`). The autodetection of scala version parses it from all dependencies, taking the part after the last underscore:

https://github.com/akka/akka-grpc/blob/7465b18b65df9783af3ee87986867cbf38945479/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy#L164

which, whether it's standardized or not, creates an issue with non-scala package names containing underscores. In my case,`akka-stream-alpakka-kinesis_2.13:2.0.1` pulls in `com.google.errorprone:error_prone_annotations:2.3.4`, which results in a gradle sync error:

```
> com.lightbend.akka.grpc.gradle requires a single major.minor version of all scala libraries in compileClasspath.
  Found [2.13, annotations]
```
Thanks for any help, or workarounds!

eshepelyuk:
@pmitura-lucid no workaround for now, but I'll start working on the fix. Shouldn't be too long to deliver.

eshepelyuk:
@raboof @pmitura-lucid created PR #1069

eshepelyuk:
#1069 is merged into master

raboof:
@pmitura-lucid could you see if the `1.0.0-34-a75ded46` snapshot indeed fixes the problem for you?

pmitura-lucid:
@eshepelyuk @raboof That was quick, thank you so much :) Tested the new snapshot, and it's building seamlessly.

Just wondering, do you store snapshot versions of runtime and codegen in some public repo? Maven central doesn't have them, so for now I just set up an override with:
```
dependencies {
    implementation group: 'com.lightbend.akka.grpc', name: 'akka-grpc-runtime_2.13', version: '1.0.0'
}

protobuf {
    plugins {
        akkaGrpc {
            artifact = 'com.lightbend.akka.grpc:akka-grpc-codegen_2.13:1.0.0:assembly@jar'
        }
    }
}
```
totally not an issue though, I can simply go with this until a tagged release comes.

eshepelyuk:
@pmitura-lucid plz take a look at 
https://github.com/akka/akka-grpc/tree/master/plugin-tester-scala

especially 
https://github.com/akka/akka-grpc/blob/master/plugin-tester-scala/build.gradle#L9-L15

[...]

PR #1079, opened 2020-07-22
fix: implicit scala-library version

eshepelyuk:
References #1069 #1067 

@raboof this is a fix for advanced usages of `Gradle`  - while dependency versions are not declared explicitly, but instead constraints are used to align version across multi build projects \ BOMs etc.


raboof:
OK TO TEST

eshepelyuk:
@raboof I would like to know if it will be possible to release this fix in next patch release any time soon ?
if yes, we would like to wait until released version comes out, otherwise we will use snapshot repos in our code.

raboof:
@eshepelyuk we don't currently have a next release planned yet, so it might be good to use a snapshot for the time being

Issue #1099, opened 2020-08-14
Add  test code generation support to the Maven plugin


samantmaharaj:
Currently the `akka-grpc-maven-plugin` only supports generating files to `target/generated-sources`.

It would be useful to have a `test-generate` mojo that reads from `src/test/proto` and `src/test/protobuf` and outputs to `target/generated-test-sources`. The mojo would also add the target directory to the project's test source roots.

I am using this plugin for code generation but have to use the protobuf maven plugin with plain `scalapb` for unit tests.

PR #1100, opened 2020-08-14
Add TestGenerateMojo to the akka-grpc-maven-plugin #1099

samantmaharaj:
References #1099

### DIFF
--- a/maven-plugin/src/main/maven/plugin.xml
+++ b/maven-plugin/src/main/maven/plugin.xml
@@ -64,6 +64,13 @@
           <description>Where to look for proto files</description>
         </parameter>
         <parameter>
+          <name>outputDirectory</name>
+          <type>string</type>
+          <required>false</required>
+          <editable>true</editable>
+          <description>Where to write the generated files</description>
+        </parameter>
+        <parameter>
           <name>protocVersion</name>
           <type>string</type>
           <required>false</required>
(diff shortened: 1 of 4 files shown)


lightbend-cla-validator:
Hi @samantmaharaj,

raboof:
OK TO TEST

maven-plugin/src/main/maven/plugin.xml
```diff
@@ -79,6 +86,90 @@
...
+      </parameters>
+      <configuration>
+        <project implementation="" default-value="${project}"/>
+        <language implementation="enum" default-value="JAVA">${akka-grpc.language}</language>
+        <generateClient implementation="boolean" default-value="true">${akka-grpc.generate-client}</generateClient>
+        <generateServer implementation="boolean" default-value="true">${akka-grpc.generate-server}</generateServer>
+        <generatorSettings implementation="java.util.List" default-value=""/>
+        <extraGenerators implementation="java.util.List" default-value=""/>
+        <protoPaths default-value="src/test/proto,src/test/protobuf">${akka-grpc.protoPaths}</protoPaths>
+        <outputDirectory default-value="target/generated-test-sources">${akka-grpc.outputDirectory}</outputDirectory>
```
raboof, line 172: Hmm, that's quite some duplication.

Is this the way other code-generating plugins do it as well?

maven-plugin/src/main/scala/akka/grpc/maven/TestGenerateMojo.scala
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright (C) 2018-2020 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.maven
+
+import javax.inject.Inject
+import org.sonatype.plexus.build.incremental.BuildContext
+
+class TestGenerateMojo @Inject() (buildContext: BuildContext) extends AbstractGenerateMojo(buildContext) {
```
raboof, line 10: :+1:

maven-plugin/src/main/maven/plugin.xml
samantmaharaj: If you develop using maven, the `plugin.xml` is generated based on annotations in the code so it's not a problem that normally comes up as you don't need to maintain it.

I could factor the common XML out and use a templating tool but I'm not that familiar with SBT so I'd appreciate some help on how to do it.

raboof: Aah makes sense. Adding templating would be possible for example with sbt-twirl, but to be honest I'm not sure the added complexity is worth it. Let's live with the duplication for now ;)

PR #1108, opened 2020-08-31
Adding grpc Scalapb option

dkichler:
Allows pass-through of the `grpc` ScalaPB generator option.  The intent is to allow the ServiceNameGrpc.scala class to be generated for uses outside the scope of an Akka gRPC client or server.  The case we ran into was consuming the native io.grpc service descriptor in the context of a Gatling test (via [this library](https://github.com/phiSgr/gatling-grpc)).  The option must be explicitly included (defaults to false in the context of `akka-grpc`).

I looked for tests I could extend to cover this but didn't find any, please do point me in the right direction if I managed to overlook them.

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/ProtocSettings.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/ProtocSettings.scala
@@ -16,5 +16,6 @@ object ProtocSettings {
     "single_line_to_proto_string",
     "ascii_format_to_string",
     "no_lenses",
-    "retain_source_code_info")
+    "retain_source_code_info",
+    "grpc")
 }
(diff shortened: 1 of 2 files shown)


lightbend-cla-validator:

raboof:
OK TO TEST

raboof (approved):
> The intent is to allow the ServiceNameGrpc.scala class to be generated for uses outside the scope of an Akka gRPC client or server

It seems we were lucky enough that this doesn't lead to any naming clashes. 

I don't think we should explicitly support having the Akka gRPC and the built-in ScalaPB gRPC code generation enabled side-by-side, but if it happens to work I'm happy to allow it.

If in the future there is some problem preventing having both generators enabled side-by-side, we should consider whether we should fix the problem or provide another way of making the io.grpc descriptors available.

> I looked for tests I could extend to cover this but didn't find any, please do point me in the right direction if I managed to overlook them.

It would be possible to add a 'scripted' test under `sbt-plugin/src/sbt-test`, but I don't think this is needed in this case.

Issue #1081, opened 2020-07-28
Client connection terminates unexpectedly on first call


ryanhanks:
### Versions used 
akka-grpc 1.0.1, akka-grpc-quickstart project

### Expected Behavior

The request completes successfully.

### Actual Behavior

Akka-grpc server terminates stream unexpectedly on first call for new connection.

A full description with steps to reproduce can be found [here](https://github.com/ryanhanks/akka-dart-grpc-connection-bug)

The client being used is a dart client. It's fair to assume part (or all) of the issue resides in that codebase, so I will cross-post this issue to the dart-grpc project. However, it's worth noting that I wasn't able to reproduce this with the same client/test against a python backend.

I was able to reproduce this with both the java and scala quickstart projects. In both cases, the only modification made was to use an HTTP connection instead of HTTPS. I'm not sure if this reproduces with HTTPS connections.

### Relevant logs

I used the akka-grpc-quickstart project to reproduce. 

### Reproducible Test Case

Please see https://github.com/ryanhanks/akka-dart-grpc-connection-bug

ryanhanks:
Cross-post to grpc-dart: https://github.com/grpc/grpc-dart/issues/334

raboof:
Thank you for this excellent report with reproducer.

I have looked into this somewhat, and at first glance it indeed seems to be a problem on the Akka gRPC/Akka HTTP side.

For me it's not consistently the 'first call' that fails, it's just failing now and then inconsistently - I suspect a race condition.

Looking at the exchange in wireshark, for each request, the dart code sends 1 data frame with a protobuf object, and then 1 'empty' data frame signalling the end of the stream. That's fine.

It appears that after reading the protobuf object, Akka gRPC cancels the stream. That's not necessarily wrong (it's just a way to signal that Akka gRPC is no longer interested in any further data), but perhaps it would be nicer to avoid it.

However, after that a couple of problematic things appear to happen (more verification needed):
* Due to this cancellation, Akka HTTP resets the stream with code 'CANCEL', moving the whole stream (both the reading and the writing side) to Closed. That seems premature.
* It then also sends a GOAWAY frame for the whole connection (that seems wrong to me, it was just this one stream that got cancelled)
* And finally it sends the reply, even though the stream was already closed.

Those are things we should look into and fix.

Possibly we could work around these particular problems in Akka gRPC by not cancelling the request, but that seems to ignore the 'real' issues.

jrudolph:
[...]

PR #1111, opened 2020-09-02
Don't cancel the incoming stream after reading the parameter

raboof:
Should fix one of the issues identified in #1081. Unfortunately even with this change, the problem still reproduces, which I didn't expect - so this still needs a closer look.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
@@ -11,12 +11,12 @@ import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.actor.ClassicActorSystemProvider
 import akka.grpc._
-import akka.grpc.internal.{ CancellationBarrierGraphStage, GrpcResponseHelpers, MissingParameterException }
+import akka.grpc.internal._
 import akka.grpc.GrpcProtocol.{ GrpcProtocolReader, GrpcProtocolWriter }
 import akka.http.javadsl.model.{ HttpRequest, HttpResponse }
 import akka.japi.Function
 import akka.stream.Materializer
-import akka.stream.javadsl.{ Sink, Source }
+import akka.stream.javadsl.Source
 import akka.util.ByteString
 
 import com.github.ghik.silencer.silent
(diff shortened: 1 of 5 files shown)


jrudolph:
I found out that the sbt project in the reproducer doesn't update the classpath between different runs so changes from recompilation might not be applied. Not sure how you are testing so not sure if it applies to your problem as well.

runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
```diff
@@ -39,10 +39,7 @@ object GrpcMarshalling {
       u: ProtobufSerializer[T],
       mat: Materializer,
       reader: GrpcProtocolReader): CompletionStage[T] =
-    data.via(reader.dataFrameDecoder).map(u.deserialize).runWith(Sink.headOption[T], mat).thenCompose[T] { opt =>
-      if (opt.isPresent) CompletableFuture.completedFuture(opt.get)
-      else failure(new MissingParameterException())
-    }
+    data.via(reader.dataFrameDecoder).map(u.deserialize).runWith(SingleParameterSink.create[T](), mat)
```
jrudolph, line 42: I just used `Sink.head` here instead and then just acted on the first element. We we only expect a single element, maybe that would be the simpler solution to `require` that the head just returns a single element.

jrudolph: E.g. like this:

```suggestion
    data.via(reader.dataFrameDecoder).map(u.deserialize).runWith(Sink.seq[T], mat).thenCompose[T] { opt =>
      if (!opt.isEmpty) {
        require(opt.size == 1, s"Expected only a single parameter but got ${opt.size}")
        CompletableFuture.completedFuture(opt.get(0))
      }
      else failure(new MissingParameterException())
    }
```

jrudolph: This seems to fix it for me.

raboof: Hmm, it still reproduces for me even with this change (and `clean`ing), after a few tries.

raboof: Oops, I incorrectly updated the reproducer - both this PR and the `Sink.seq` approach seem work (since both are pulling until completion instead of cancelling after the first (and only) element).

jrudolph: Ah, great to have that mystery resolved at least :)

ennru (commented):
If the `Sink.seq` with reading the seq is enough, why introduce the `SingleParameterStage`?

raboof:
> If the `Sink.seq` with reading the seq is enough, why introduce the `SingleParameterStage`?

It seemed a bit simpler. I'm fine with replacing it with a `Sink.seq`-based implementation, though.

ennru (approved):
It keeps this special sink handling to one place, otherwise to mapping would be duplicated for the Java DSL and the Scala DSL.
LGTM.

raboof:
> It keeps this special sink handling to one place, otherwise to mapping would be duplicated for the Java DSL and the Scala DSL.
> LGTM.

works for me, we can always switch when this doesn't turn out well

Issue #1096, opened 2020-08-12
sbt: fix and/or document usage of akkaGrpcExtraGenerators


bjaglin:
Custom generators declared and compiled as part of the metabuild could be used as `akkaGrpcExtraGenerators` until Akka-gRPC 1.0.0. This is no longer possible as of 1.0.1 (`java.lang.ClassNotFoundException`), since https://github.com/akka/akka-grpc/pull/1032 can only load generators packaged as part of the Akka-gRPC JARs: https://github.com/akka/akka-grpc/blob/2d1242466e42821b70347ce329c1bd8a1f0db78d/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L193-L205.

`akkaGrpcExtraGenerators` is not documented nor tested via a scripted, so maybe it was not meant to be used this way?

The following workaround seems to work:
```scala
      Seq(Compile, Test).flatMap(inConfig(_) {
        Seq(
          PB.artifactResolver := {
            val prev = PB.artifactResolver.value
            val metaBuildClasses =
              (ThisBuild / baseDirectory).value / "project/target/scala-2.12/sbt-1.0/classes/"
            artifact: protocbridge.Artifact => prev(artifact) :+ metaBuildClasses
          },
        )
      })
```

I'll try to open a PR with a failing scripted to demonstrate this.

raboof:
> since #1032 can only load generators packaged as part of the Akka-gRPC JARs

Ah, I didn't realize that (though in retrospect it makes sense). That was not intentional: while indeed we haven't promoted akkaGrpcExtraGenerators much so far, it was definitely intended to allow downstream projects to add additional generators.

bjaglin:
https://github.com/bjaglin/akka-grpc/commits/akkaGrpcExtraGenerators is a repo with 2 commits:
- https://github.com/bjaglin/akka-grpc/commit/4e7bc05e4acb8c4131563bf57009faa3a6c47ce1 describes the usage that works on 1.0.0 in a scripted
- https://github.com/bjaglin/akka-grpc/commit/bf01972796b6607c698cf0f9fe796f7bd1058c8c merges 1.0.1 on top of the previous commit, making the scripted regress (`ava.lang.ClassNotFoundException: generators.CustomGenerator$`)
- https://github.com/bjaglin/akka-grpc/commit/e49553e8045f511e80f31d9624f84122ba6ededb "valides" the workaround documented above

[...]

PR #1115, opened 2020-09-07
Expose helpers to bridge external generators to sbt

raboof:
Possible solution to #1096, this way the project registering
the additional generator can decide for itself whether to use
sandboxing.

From here the step to just providing something directly to
PB.targets is pretty small as well.

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -7,7 +7,7 @@ package akka.grpc.sbt
 import akka.grpc.gen.CodeGenerator.ScalaBinaryVersion
 import akka.grpc.gen.scaladsl.{ ScalaClientCodeGenerator, ScalaServerCodeGenerator, ScalaTraitCodeGenerator }
 import akka.grpc.gen.javadsl.{ JavaClientCodeGenerator, JavaInterfaceCodeGenerator, JavaServerCodeGenerator }
-import akka.grpc.gen.{ Logger => GenLogger, BuildInfo, ProtocSettings }
+import akka.grpc.gen.{ ProtocSettings, Logger => GenLogger }
 import protocbridge.Generator
 import sbt.Keys._
 import sbt.{ GlobFilter, _ }
@@ -111,7 +111,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
             akkaGrpcGeneratedLanguages.value,
             ScalaBinaryVersion(scalaBinaryVersion.value),
             generatorLogger) ++ akkaGrpcExtraGenerators.value.map(g =>
-            toGenerator(g, ScalaBinaryVersion(scalaBinaryVersion.value), generatorLogger))
+            GeneratorBridge.sandboxedGenerator(g, ScalaBinaryVersion(scalaBinaryVersion.value), generatorLogger))
         },
         // configure the proto gen automatically by adding our codegen:
         PB.targets ++=
@@ -162,7 +162,8 @@ object AkkaGrpcPlugin extends AutoPlugin {
       scalaBinaryVersion: ScalaBinaryVersion,
       logger: GenLogger): Seq[protocbridge.Generator] = {
     import AkkaGrpc._
-    def toGen(codeGenerator: akka.grpc.gen.CodeGenerator) = toGenerator(codeGenerator, scalaBinaryVersion, logger)
+    def toGen(codeGenerator: akka.grpc.gen.CodeGenerator) =
+      GeneratorBridge.sandboxedGenerator(codeGenerator, scalaBinaryVersion, logger)
     // these two are the model/message (protoc) generators
     def ScalaGenerator: protocbridge.Generator = scalapb.gen.SandboxedGenerator
     // we have a default flat_package, but that doesn't play with the java generator (it fails)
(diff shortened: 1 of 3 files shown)


sbt-plugin/src/main/scala/akka/grpc/sbt/GeneratorBridge.scala
```diff
@@ -0,0 +1,70 @@
...
+  def jvmGenerator(
+      codeGenerator: akka.grpc.gen.CodeGenerator,
+      scalaBinaryVersion: ScalaBinaryVersion,
+      logger: akka.grpc.gen.Logger): protocbridge.Generator = {
+    val adapter = new PlainProtocBridgeSbtPluginCodeGenerator(codeGenerator, scalaBinaryVersion, logger)
+    protocbridge.JvmGenerator(codeGenerator.name, adapter)
+  }
+
+  /** INTERNAL API */
+  class PlainProtocBridgeSbtPluginCodeGenerator(
```
bjaglin, line 34: could these implementations be completely hidden?
```suggestion
  private class PlainProtocBridgeSbtPluginCodeGenerator(
```

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -190,42 +191,13 @@ object AkkaGrpcPlugin extends AutoPlugin {
...
-    override def run(request: Array[Byte]): Array[Byte] =
-      runMethods.invoke(module, request, reflectiveLogger).asInstanceOf[Array[Byte]]
-    override def toString = s"ProtocBridgeSbtPluginCodeGenerator(${className})"
-  }
+  def sandboxedGenerator(codeGenerator: akka.grpc.gen.CodeGenerator): Def.Initialize[protocbridge.Generator] =
+    Def.setting {
+      GeneratorBridge.sandboxedGenerator(codeGenerator, ScalaBinaryVersion(scalaBinaryVersion.value), generatorLogger)
+    }
 
+  def jvmGenerator(codeGenerator: akka.grpc.gen.CodeGenerator): Def.Initialize[protocbridge.Generator] =
```
bjaglin, line 199: It might be a bias from knowing the sbt-protoc wording, but I find it slightly confusing to use `jvm` alone, as `sandboxedGenerator` also exposes a generator defined in a JVM bytecode. Not sure it's better, but to better emphasize the difference with `sandboxedGenerator`, what about 
```suggestion
  def sharedJvmGenerator(codeGenerator: akka.grpc.gen.CodeGenerator): Def.Initialize[protocbridge.Generator] =
```

bjaglin (commented):
looking good! I agree `akkaGrpcExtraGenerators` should be deprecated (it was not well documented anyway), but maybe this PR can maintain backward compatibility by wrapping the `akkaGrpcExtraGenerators` in `jvmGenerator` ?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
bjaglin: Actually the wording is the same for that https://github.com/scalapb/protoc-bridge/blob/f0e09cecfdef9f352b17acd94b9d87435f0633b1/bridge/src/main/scala/protocbridge/Generator.scala#L28-L32, but the JVM is also present in the sandboxed version.

raboof: I guess the 'jvm' is a bit silly here because we don't support non-JVM generators (if you want to use those, you can just add them to PB.targets).

I agree `jvmGenerator` is a bit obscure - perhaps `plainGenerator` would be better? Or `unsandboxed`? I generally dislike negations in names, but in this case it might be justified since it is a 'somewhat dangerous'. WDYT?

raboof:
> maybe this PR can maintain backward compatibility by wrapping the `akkaGrpcExtraGenerators` in `jvmGenerator`?

I guess there's no harm in that :+1:

sbt-plugin/src/main/scala/akka/grpc/sbt/GeneratorBridge.scala
```diff
@@ -0,0 +1,70 @@
...
+  def plainGenerator(
+      codeGenerator: akka.grpc.gen.CodeGenerator,
+      scalaBinaryVersion: ScalaBinaryVersion,
+      logger: akka.grpc.gen.Logger): protocbridge.Generator = {
+    val adapter = new PlainProtocBridgeSbtPluginCodeGenerator(codeGenerator, scalaBinaryVersion, logger)
+    protocbridge.JvmGenerator(codeGenerator.name, adapter)
+  }
+
+  /**
+   * Uses reflection to load a [[akka.grpc.gen.CodeGenerator]] and turns it into protocbridge required type without sandboxing.
```
bjaglin, line 34: this does not use reflection, does it?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
bjaglin: :+1:

sbt-plugin/src/main/scala/akka/grpc/sbt/GeneratorBridge.scala
raboof: Oops ;)

ennru (approved):
LGTM.

Issue #138, opened 2018-03-29
Circuit Breaking


ignasi35:
Akka-gRPC client should allow setting up circuit breaking. Related to #137

ignasi35:
Some mental notes:

 A gRPC client has a `Channel` instance connecting the client to a single remote host and that `Channel` is shared across several calls.

Lagom's `CircuitBreakingServiceLocator` offers a different circuit breaker instance for each call on a remote service. Different approaches could be: (1) per remote service or (2) per remote instance. Maybe akka-grpc could consider supporting all 3 alternatives in the future.


Akka provides general purpose [circuit breakers](https://doc.akka.io/docs/akka/2.5/common/circuitbreaker.html) so users can manually setup circuit breaking around their `GrpcService` calls if they want to.

Lagom provides a general purpose [circuit breaker panel](https://github.com/lagom/lagom/issues/1248) and _also_ promotes `ServiceLocator` implementors to extend [`CircuitBreakingServiceLocator`](https://www.lagomframework.com/documentation/1.4.x/scala/api/com/lightbend/lagom/scaladsl/client/CircuitBreakingServiceLocator.html) which keeps circuit breaking a bit hidden but still exposed to users via Settings. `CircuitBreakingServiceLocator` looks up a service instance and then runs userland code wrapped on a circuit breaker.

yahor-filipchyk:
gRPC project has this approved [proposal](https://github.com/grpc/proposal/blob/master/A6-client-retries.md) for retries that includes retry throttling. I think it'd be nice if `akka-grpc` client could use the configuration and semantics written in the proposal, opening the door for future full support of gPRC standard, including service config discovery via the DNS or other service discovery mechanisms.

raboof:
I agree this is an area where the Akka gRPC client can be improved! We should definitely look at the linked proposal, but also remember to look for the best ways to translate this into Akka ecosystem concepts (such as akka-discovery and eventually akka-http)

ignasi35:
After digging a bit more on the topic, I've found [a discussion](https://github.com/grpc/grpc-java/issues/7281) hinting the grpc-java client can support circuit breaking via `Interceptor`s (the discussion links to a [PoC in Kotlin using resilience4J](https://gist.github.com/eungju/226274b3dacb3203de3514bcf1c54505#file-circuitbreakerclientinterceptor-kt)).

[...]

PR #1118, opened 2020-09-09
(Draft) Interceptor implementing circuit breaking

ignasi35:
References #138

A [recommended way to include Circuit Breakers on a `grpc-java` client](https://github.com/grpc/grpc-java/issues/7281#issuecomment-668785443) is using `Interceptor`'s. Based on a [linked sample](https://gist.github.com/eungju/226274b3dacb3203de3514bcf1c54505#file-circuitbreakerclientinterceptor-kt) this PR provides an `io.grpc.Interceptor` that uses an Akka Circuit Breaker.

There's a [sandbox project](https://github.com/ignasi35/akka-grpc-circuit-breaker-sandbox) to play with this code a bit.

TODO: test, document, API,...

### DIFF
--- a/runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
+++ b/runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
@@ -1,0 +1,57 @@
+package akka.grpc.netty
+
+import akka.pattern.CircuitBreaker
+import io.grpc.CallOptions
+import io.grpc.Channel
+import io.grpc.ClientCall
+import io.grpc.ClientInterceptor
+import io.grpc.ClientInterceptors
+import io.grpc.ForwardingClientCallListener
+import io.grpc.Metadata
+import io.grpc.MethodDescriptor
+import io.grpc.Status
+
+/**
+ * A `io.grpc.ClientInterceptor` that prevents downstream overload wrapping the calls with a circuit breaker.
+ */
+class NettyCircuitBreakerInterceptor(breaker: CircuitBreaker) extends ClientInterceptor {
+
+  private class Listener[RespT](delegate: ClientCall.Listener[RespT])
+      extends ForwardingClientCallListener.SimpleForwardingClientCallListener[RespT](delegate) {
+    override def onClose(status: Status, trailers: Metadata): Unit = {
+      if (status.isOk) {
+        breaker.succeed()
+      } else {
+        breaker.fail()
+      }
+
+      super.onClose(status, trailers)
+    }
+  }
+
+  override def interceptCall[ReqT, RespT](
+      method: MethodDescriptor[ReqT, RespT],
+      callOptions: CallOptions,
+      next: Channel): ClientCall[ReqT, RespT] =
+    new ClientInterceptors.CheckedForwardingClientCall[ReqT, RespT](next.newCall(method, callOptions)) {
+
+      override def checkedStart(responseListener: ClientCall.Listener[RespT], headers: Metadata): Unit = {
+        delegate().start(new Listener(responseListener), headers)
+      }
+
+      override def isReady: Boolean = {
+        if (breaker.isClosed || breaker.isHalfOpen) {
+          super.isReady
+        } else {
+          throw new IllegalStateException("Circuit Breaker is open. Can't send messages.")
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
```diff
@@ -0,0 +1,57 @@
...
+import io.grpc.ClientInterceptors
+import io.grpc.ForwardingClientCallListener
+import io.grpc.Metadata
+import io.grpc.MethodDescriptor
+import io.grpc.Status
+
+/**
+ * A `io.grpc.ClientInterceptor` that prevents downstream overload wrapping the calls with a circuit breaker.
+ */
+class NettyCircuitBreakerInterceptor(breaker: CircuitBreaker) extends ClientInterceptor {
```
ignasi35, line 17: This [interceptor](https://github.com/grpc/grpc-java/blob/master/api/src/main/java/io/grpc/ClientInterceptor.java#L21-L35) would be used like this:

```
val circuitBreakingInterceptor = new NettyCircuitBreakerInterceptor(myBreaker)
GrpcClientSettings.fromConfig("helloworld.GreeterService")
      .withChannelBuilderOverrides{ builder =>
        builder.intercept(circuitBreakingInterceptor)
      }
```

so even if the interceptor seems to operate at call level (only `interceptCall()` is provided), I think there's a single instance per client. 

I am not sure how an Interceptor interferes with the Channel lifecycle.

ignasi35:
After discussing offline, we'll introduce a `withCircuitBreaker` in `GrpcClientSettings` so the user is not exposed to grpc-java API. It'll also gives us the freedom to re-implement differently if we find a more convenient approach, interference with other features or regressions.

runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
```diff
@@ -0,0 +1,57 @@
...
+
+      override def checkedStart(responseListener: ClientCall.Listener[RespT], headers: Metadata): Unit = {
+        delegate().start(new Listener(responseListener), headers)
+      }
+
+      override def isReady: Boolean = {
+        if (breaker.isClosed || breaker.isHalfOpen) {
+          super.isReady
+        } else {
+          throw new IllegalStateException("Circuit Breaker is open. Can't send messages.")
```
patriknw, line 46: use `CircuitBreakerOpenException`?

runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * A `io.grpc.ClientInterceptor` that prevents downstream overload wrapping the calls with a circuit breaker.
+ */
+class NettyCircuitBreakerInterceptor(breaker: CircuitBreaker) extends ClientInterceptor {
+
+  private class Listener[RespT](delegate: ClientCall.Listener[RespT])
+      extends ForwardingClientCallListener.SimpleForwardingClientCallListener[RespT](delegate) {
+    override def onClose(status: Status, trailers: Metadata): Unit = {
+      if (status.isOk) {
+        breaker.succeed()
```
patriknw, line 23: One thing we'll miss when using breaker.succeed/fail is the callTimeout feature. Are there timeout's at the gRCP level for req/response calls?

If we want to include callTimeout we could use `withCircuitBreaker` and create a Promise when call is started and complete/fail that here. That's more overhead than breaker.succeed/fail so only if we think callTimeout is important.

patriknw (commented):
nice

runtime/src/main/scala/akka/grpc/netty/NettyCircuitBreakerInterceptor.scala
ignasi35: Yeah, this is one of the areas where I struggled the most. I think the exception is swallowed and treated differently by the grpc-java internals. I couldn't get to the exception I was throwing from user code so I settled to using a vanilla Java API one. 
If it makes no difference we can switch to `CircuitBreakerOpenException`.

ignasi35:
I don't think we'll be pursuing this.

PR #1127, opened 2020-09-24
Documentation for running a gRPC service on Kubernetes

raboof:


docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+
+An Akka gRPC service can be deployed to a Kubernetes cluster just like any other HTTP-based application,
+but if you want to make sure HTTP/2 is used across the board there are some things to consider.
+
+There are 3 ways to use HTTP/2:
+
+* With TLS
+* Without TLS ('h2c'), using the `Upgrade` mechanism to negotiate between using HTTP/1 and HTTP/2
+* Without TLS ('h2c'), without negotiation (which assumes the client has prior knowledge that the server supports HTTP/2)
+
```
ignasi35, line 11:
```suggestion
* HTTP/2 over TLS
* HTTP/2 over plaintext HTTP ('h2c') with negotiation using the `Upgrade` mechanism to upgrade from HTTP/1 to HTTP/2
* HTTP/2 over plaintext HTTP ('h2c') without negotiation which assumes the client has prior knowledge that the server supports HTTP/2

```

I'm not entirely sure about the parens VS comma on the last sentence.

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+Limitations of this approach are that you are restricted to 1 service per IP, and
+more advanced features such as TLS termination and path-based routing are not available.
+
+## Nginx Ingress
+
+Nginx, by default, will only support h2c via the `Upgrade` mechanism. Because this mechanism is more
+complicated than the other 2 approaches, client support for this mechanism is not ubiquitous. Also, the
+loadbalancer will use HTTP/1 rather than HTTP/2 to the service. While this may happen to work it is
+not optimal.
+
```
ignasi35, line 30:
```suggestion
Of the three ways to use HTTP/2 presented at the beginning of this page, nginx only supports 
`h2c` with negotiation via the `Upgrade` mechanism. Because the `Upgrade` mechanism is more
complicated than the other 2 approaches (aka "TLS", and "prior knowledge"), client support 
for `Upgrade` mechanism is not ubiquitous. Also, the loadbalancer will use HTTP/1 rather 
than HTTP/2 to the service. While this may happen to work it is not optimal.

```

I tried to remove as many pronouns as possible. But the current writing seems to focus too much in some details and doesn't paint `nginx` as a very valid option. I liked the `LoadBalancer Service` section because it listed the disadvantages very clearly.

--------------------

I'm not sure how to rewrite the last bit. 

> Also, the loadbalancer will use HTTP/1 rather than HTTP/2 to the service. While this may happen to work it is not optimal.

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+## GCE Ingress
+
+GCE only supports HTTP/2 over TLS. This means you must configure GCE to use HTTPS both on the
+'outside' and on this 'inside', and you will have to make sure your service is exposed over HTTPS
+internally as well.
```
ignasi35, line 35: Hmm, this section also starts indicating what's supported and what's not. Maybe have a repeated list of what's supported or not or what's the default setting, etc... would be helpful.

Or even move that table to the top of the page:

|   | nginx | GCE |
| --------------- | --------------- | --------------- |
| HTTP/2 over TLS | supported | supported |
| h2c + `Upgrade` | default | not supported |
| h2c + prior knowledge | supported | not supported |

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+  selector:
+    app: metrics
+    department: sales
+  ports:
+  - name: my-grpc-port
+    port: 443
+    targetPort: 8443
+```
+
+By default the loadbalancer will accept any certificate, so your service can be configured with
```
ignasi35, line 58: Sorry, I didn't review this last section

project/plugins.sbt
```diff
@@ -20,7 +20,8 @@ addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.4.0")
 
 // docs
 addSbtPlugin("com.lightbend.akka" % "sbt-paradox-akka" % "0.35")
-addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-dependencies" % "0.2.1")
+//addSbtPlugin("com.lightbend.paradox" % "sbt-paradox-dependencies" % "0.2.1")
```
ignasi35, line 23: Can we remove this plugin, then?

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
```diff
@@ -53,7 +53,7 @@ object NettyClientUtils {
             settings.resolveTimeout))
 
     if (!settings.useTls)
-      builder = builder.usePlaintext()
+      builder = builder.negotiationType(NegotiationType.PLAINTEXT_UPGRADE)
```
ignasi35, line 56: Was this change intended? I guess it's fixing a deprecation warning?

ignasi35 (changes requested):
I've focused on the docs section (the meat of the PR). I think the gist of it is that the `loadbalancer service` section seems to be out of place, though I understand it's necessary to mention it. 

Then, I think a summary of the capabilities of each ingress may be helpful. And, finally, prefer lists over walls of text. 

Please take all the suggestions with a grain of salt, I'm not English native nor a Technical Writer specialist.

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+
+## LoadBalancer Service
+
+By creating a service of type `LoadBalancer` you can expose your services using a direct TCP loadbalancer.
+This means all features supported by the underlying implementation are available.
+
+Limitations of this approach are that you are restricted to 1 service per IP, and
+more advanced features such as TLS termination and path-based routing are not available.
+
+## Nginx Ingress
```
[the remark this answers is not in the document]
ignasi35: I stand corrected. The preferred way seems to be all caps (`NGINX`) ass seen in nginx.com (as opposed to nginx.org).

runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
raboof, line 56: Definitely not :D

```suggestion
      builder = builder.usePlaintext()
```

project/plugins.sbt
raboof: oops, that was unintentional

docs/src/main/paradox/server/kubernetes.md
raboof: While I see the desire for such a table, I'm not sure we can pull it off: all the variations come with their own "ifs and buts", and it is not clear whether this is talking about the connection between the client and the loadbalancer, or the connection between the loadbalancer and the service (or both). Might be interesting to revisit, but I'd like to keep it outside the scope of this PR :).

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,71 @@
...
+
+* With TLS
+* Without TLS ('h2c'), using the `Upgrade` mechanism to negotiate between using HTTP/1 and HTTP/2
+* Without TLS ('h2c'), without negotiation (which assumes the client has prior knowledge that the server supports HTTP/2)
+
+Kubternetes supports many types of [Ingress Controllers](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/),
+but [GCE](https://git.k8s.io/ingress-gce/README.md) and [nginx](https://git.k8s.io/ingress-nginx/README.md)
+seem to be the most widely known.
+
+## LoadBalancer Service
```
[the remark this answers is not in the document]
raboof: Added it there

ennru (approved):
LGTM. Great to have this written down for future reference...

docs/src/main/paradox/server/kubernetes.md
```diff
@@ -0,0 +1,74 @@
...
+Of the three ways to use HTTP/2 presented at the beginning of this page, NGINX by default only supports 
+`h2c` with negotiation via the `Upgrade` mechanism. Because the `Upgrade` mechanism is more
+complicated than the other 2 approaches ("TLS", and h2c with "prior knowledge"), gRPC client support 
+for h2c with the `Upgrade` mechanism is not ubiquitous. Also, the loadbalancer will use HTTP/1 rather 
+than HTTP/2 for the connection from the loadbalancer to the service. While this may happen to work it is not optimal.
+
+## GCE Ingress
+
+GCE only supports HTTP/2 over TLS. This means you must configure GCE to use HTTPS both on the
+'outside' and on this 'inside', and you will have to make sure your service is exposed over HTTPS
```
ignasi35, line 37:
```suggestion
'outside' and on the 'inside', and you will have to make sure your service is exposed over HTTPS
```

ignasi35 (approved):
Minor typo, otherwise LGTM

PR #1132, opened 2020-10-01
Generate version.conf file in runtime jar

ennru:
Add a `version.conf` file to make the Akka gRPC version available outside of class files.
This complements the generated `akka.grpc.gen.BuildInfo` class.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ lazy val codegen = Project(id = akkaGrpcCodegenId, base = file("codegen"))
       val file = assembly.value
       Assemblies.mkBatAssembly(file)
     },
-    buildInfoKeys ++= BuildInfoKey.ofN(organization, name, version, scalaVersion, sbtVersion),
+    buildInfoKeys ++= Seq[BuildInfoKey](organization, name, version, scalaVersion, sbtVersion),
     buildInfoKeys += "runtimeArtifactName" -> akkaGrpcRuntimeName,
     buildInfoKeys += "akkaVersion" → Dependencies.Versions.akka,
     buildInfoKeys += "akkaHttpVersion" → Dependencies.Versions.akkaHttp,
@@ -43,6 +43,7 @@ lazy val codegen = Project(id = akkaGrpcCodegenId, base = file("codegen"))
 
 lazy val runtime = Project(id = akkaGrpcRuntimeName, base = file("runtime"))
   .settings(Dependencies.runtime)
+  .settings(VersionGenerator.settings)
   .settings(
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := Dependencies.Versions.CrossScalaForLib.head)
--- a/project/VersionGenerator.scala
+++ b/project/VersionGenerator.scala
@@ -1,0 +1,24 @@
+import sbt.Keys._
+import sbt._
+
+/**
+ * Generate version.conf file based on the version setting.
+ *
+ * This was adapted from https://github.com/akka/akka/blob/v2.6.8/project/VersionGenerator.scala
+ */
+object VersionGenerator {
+
+  val settings: Seq[Setting[_]] = inConfig(Compile)(
+    Seq(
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -23,7 +23,7 @@ lazy val codegen = Project(id = akkaGrpcCodegenId, base = file("codegen"))
       val file = assembly.value
       Assemblies.mkBatAssembly(file)
     },
-    buildInfoKeys ++= BuildInfoKey.ofN(organization, name, version, scalaVersion, sbtVersion),
+    buildInfoKeys ++= Seq[BuildInfoKey](organization, name, version, scalaVersion, sbtVersion),
```
raboof, line 26: :+1: nice to fix this warning :)

project/VersionGenerator.scala
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Generate version.conf file based on the version setting.
+ *
+ * This was adapted from https://github.com/akka/akka/blob/v2.6.8/project/VersionGenerator.scala
+ */
+object VersionGenerator {
+
+  val settings: Seq[Setting[_]] = inConfig(Compile)(
+    Seq(resourceGenerators += generateVersion(resourceManaged, _ / "version.conf", """|akka.grpc.version = "%s"
```
raboof, line 12: OK I guess :)

raboof (approved):

raboof:
The generated `version.conf` files clash when creating an assembly:

```
[info] [error] java.lang.RuntimeException: deduplicate: different file contents found in the following:
[info] [error] /home/jenkinsakka/.cache/coursier/v1/https/repo1.maven.org/maven2/com/typesafe/akka/akka-actor_2.12/2.5.31/akka-actor_2.12-2.5.31.jar:version.conf
[info] [error] /localhome/jenkinsakka/.ivy2/local/com.lightbend.akka.grpc/akka-grpc-runtime_2.12/1.0.2-4-29bf5f74/jars/akka-grpc-runtime_2.12.jar:version.conf
```

Of course it is possible to fix this by choosing the 'append' strategy, but since apparently AFAICT we're not doing this yet for any of the other sattelite projects yet, perhaps we should consider if we want to let them clash like this?

ennru:
> Of course it is possible to fix this by choosing the 'append' strategy, but since apparently AFAICT we're not doing this yet for any of the other sattelite projects yet, perhaps we should consider if we want to let them clash like this?

I said the same when we added it to Alpakka Kafka.

https://github.com/akka/alpakka-kafka/pull/1188#discussion_r475750876

sbt-plugin/src/sbt-test/gen-scala-server/07-gen-basic-server-with-akka-26/build.sbt
```diff
@@ -4,14 +4,13 @@ enablePlugins(AkkaGrpcPlugin)
...
 dependencyOverrides += "com.typesafe.akka" %% "akka-stream" % "2.6.5"
 
-// Should no longer be needed from Akka 2.6.6, which includes
-// d2afff6bfc29baf0fb03f531fbe3af40a97fd39f
-assemblyMergeStrategy in assembly := {
+assembly / assemblyMergeStrategy := {
   case PathList("google", "protobuf", "field_mask.proto") =>
     // The only difference is in a comment, so we could also just pick
     // one, but we don't need it anyway so better to be deterministic:
     MergeStrategy.discard
```
ennru, line 11: Should I remove this while at it?

raboof (commented):
I'm not really a fan of introducing additional required configuration when creating an assembly.

Where is this version.conf used? Couldn't we choose names so we can avoid name clashes on the classpath?

ennru:
Thanks for pushing back. I should have done that in the Alpakka Kafka case as well.
I chose a library-unique file name now.

The file is used by other tools when a project produces a fat-jar so that the jar manifest files can't be used to detect the originating version of the libraries.

raboof (approved):

PR #1134, opened 2020-10-02
[wip] move to Akka HTTP 10.2.x API's

raboof:
Some things need double-checking, particularly the benchmarking project
is not converted properly yet.

References #1101

Issue #928, opened 2020-04-21
Exception handler API


raboof:
The exception handler can currently be customized by passing in a `Throwable => Trailers` function.

It might be helpful to wrap this in a type alias for Scala users to make the default (and perhaps other utility functions) more discoverable for users. For Java, a class might make sense like in https://doc.akka.io/japi/akka-http/10.1.11/?akka/http/javadsl/server/ExceptionHandler.html .

raboof:
When we pick this up we can also remove the ActorSystem parameter.

PR #1135, opened 2020-10-03
Fixes #1101. Using akka.japi.function.Function .

### LINKED ISSUE #1101: Move to akka.japi.function.Function for the javadsl
As anticipated in #994, once we drop support for Akka HTTP 10.1.x (#914) we should also make sure the javadsl uses `akka.japi.function.Function` rather than `akka.japi.Function`.

dotbg:
References #1101


benchmark-java/src/main/java/akka/grpc/benchmarks/driver/LoadWorker.java
```diff
@@ -67,11 +66,11 @@
...
 
-    CompletionStage<ServerBinding> bound = Http.get(system).bindAndHandleAsync(
-        WorkerServiceHandlerFactory.create(impl, system),
-        ConnectWithHttps.toHostHttps("0.0.0.0", driverPort).withCustomHttpsContext(Utils.serverHttpContext()),
-        mat);
-
+    CompletionStage<ServerBinding> bound = Http.get(system)
+            .newServerAt("127.0.0.1", driverPort)
+            .enableHttps(Utils.serverHttpContext())
+            .withMaterializer(mat)
```
raboof, line 72: (this might not be needed)

benchmark-java/src/main/java/akka/grpc/benchmarks/qps/AsyncServer.java
```diff
@@ -77,22 +76,22 @@ public void run(InetSocketAddress address, boolean useTls) throws Exception {
 
     benchmarkService = new BenchmarkServiceImpl(mat);
 
+    final ServerBuilder serverBuilder = Http
+        .get(system)
+        .newServerAt(address.getHostName(), address.getPort())
+        .withMaterializer(mat);
```
raboof, line 82: (this line might not be needed)

runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
```diff
@@ -14,7 +14,7 @@ import akka.grpc._
 import akka.grpc.internal._
 import akka.grpc.GrpcProtocol.{ GrpcProtocolReader, GrpcProtocolWriter }
 import akka.http.javadsl.model.{ HttpRequest, HttpResponse }
-import akka.japi.Function
+import akka.japi.function.Function
```
raboof, line 17: Here we promised binary compatibility, so we cannot 'just' change from `akka.japi.Function` to `akka.japi.function.Function`.

To try and not bite off too much at once, we could keep `akka.japi.Function` for the eHandler (also in the code generators). Later we plan to replace that with a dedicated class anyway (see #928)

runtime/src/main/scala/akka/grpc/javadsl/RouteUtils.scala
```diff
@@ -14,7 +14,7 @@ import akka.http.javadsl.model.{ HttpRequest, HttpResponse }
 import akka.http.javadsl.server.Route
 import akka.http.javadsl.server.directives.RouteAdapter
 import akka.http.scaladsl.server.RouteResult
-import akka.japi.Function
+import akka.japi.function.Function
```
raboof, line 17: Instead of converting them to `akka.japi.function.Function`, I think we should not need these utilities at all anymore, we should be able to remove them here and replace them with helpers/directives now available in Akka HTTP

raboof (commented):
This partly overlaps with #1134 - sorry, I didn't link that one properly yet.

Overall the direction looks good, but there are a couple of spots that we need to look at more carefully.

raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -10,7 +10,8 @@ import akka.annotation.ApiMayChange
 import akka.annotation.InternalApi
 import akka.grpc.scaladsl.{ ServiceHandler => sServiceHandler }
 import akka.http.javadsl.model.{ HttpRequest, HttpResponse, StatusCodes }
-// using japi because bindAndHandleAsync expects that
+
+// using japi because bind expects that.
```
raboof, line 14: Bind should take a `akka.japi.function.Function`, right?

dotbg: Indeed. I am(a bit) lost as there are few places, I can not simply use the `akka.japi.function.Function` and `akka.japi.Function` simultaneously. Besides, I do not think, the implicit conversions could help for the java code. 
One way could be to create an additional class and do the changes there. Just like the [PatternsCS](https://doc.akka.io/japi/akka/current/akka/pattern/PatternsCS.html)  were used. 
In case, this solution is fine with you, I can do the changes.

dotbg:
> 
> 
> This partly overlaps with #1134 - sorry, I didn't link that one properly yet.
> 
> Overall the direction looks good, but there are a couple of spots that we need to look at more carefully.

Thanks. I am(a bit) lost as there are few places, I can not simply use the akka.japi.function.Function and akka.japi.Function simultaneously. Besides, I do not think, the implicit conversions could help for the java code.
One way could be to create an additional class and do the changes there. Just like the PatternsCS were used.
In case, this solution is fine with you, I can do the changes.

runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
dotbg: Well, I can have a look on that issue first.

raboof:
> I am(a bit) lost as there are few places, I can not simply use the akka.japi.function.Function and akka.japi.Function simultaneously

Yeah, it's a bit tricky in some places. I'll see if I can give some more detailed pointers (or amend the PR, if you don't mind) tomorrow.

dotbg:
> 
> > I am(a bit) lost as there are few places, I can not simply use the akka.japi.function.Function and akka.japi.Function simultaneously
> 
> Yeah, it's a bit tricky in some places. I'll see if I can give some more detailed pointers (or amend the PR, if you don't mind) tomorrow.

I do not mind amending the PR.

runtime/src/main/scala/akka/grpc/javadsl/WebHandler.scala
```diff
@@ -66,7 +66,8 @@ object WebHandler {
       mat: Materializer,
       corsSettings: CorsSettings): JFunction[HttpRequest, CompletionStage[HttpResponse]] = {
     import scala.collection.JavaConverters._
-    val servicesHandler = concat(handlers.asScala.toList: _*)
+
+    val servicesHandler = concat(handlers.asScala: _*)
```
raboof, line 70: this `.toList` was actually needed for compiling on Scala 2.13

dotbg: oh, thanks for pointing out

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -50,11 +49,27 @@ object ServiceHandler {
...
   def handler(handlers: JFunction[HttpRequest, CompletionStage[HttpResponse]]*)
       : JFunction[HttpRequest, CompletionStage[HttpResponse]] = {
-    val servicesHandler = concat(handlers: _*)
+    val servicesHandler = _concat(handlers: _*)
     (req: HttpRequest) => if (sServiceHandler.isGrpcRequest(req)) servicesHandler(req) else unsupportedMediaType
   }
 
-  private[javadsl] def concat(handlers: JFunction[HttpRequest, CompletionStage[HttpResponse]]*)
+  @varargs
+  def concat(handlers: akka.japi.function.Function[HttpRequest, CompletionStage[HttpResponse]]*)
```
raboof, line 57: I don't think we need this one anymore, we should just be able to use the `concatOrNotFound`

raboof (commented):
I think the ServiceHandler can be simplified, after that LGTM!

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
dotbg: Ok. will do

dotbg: done

raboof (approved):
LGTM!

Issue #917, opened 2020-04-11
Expose option to generate the standard types


claudio-scandura:
### Short description

I have come across this issue while generating sources for proto files that include references to standard schemas. The source generation step fails because the standard Protobuf types are not included in the compilation.
I have only experienced this with the Maven plugin and have not tried SBT, so I don't exclude that this issue applies there too.

### Details
If you try to generate sources for the following proto snippet using the Maven plugin it will wail because it won't find the ` google.protobuf.Timestamp` type.
```
syntax = "proto3";

import "google/protobuf/timestamp.proto";

message Test {
    google.protobuf.Timestamp updatedAt = 1;
}
```
This can be worked around with a bit of a hack where when setting up the plugin we use the `protocVersion` config to inject the `--include_std_types` param.

```
<plugin>
      <groupId>com.lightbend.akka.grpc</groupId>
      <artifactId>akka-grpc-maven-plugin</artifactId>
      <configuration>
          <protocVersion>--include_std_types</protocVersion>
      </configuration>
</plugin>
```
Luckily this is passed all the way down by the plugin and it works.

### Possible solutions
Ideally, any extra argument should be injectable by the user to allow customisation, so another field of type List could be added to the xml schema.
Alternatively, the specific argument `--include_std_type` could be modelled as a boolean flag in the xml schema and possibly defaulted to true as I would imagine it to be quite common to use standard types in proto definitions.

raboof:
> ###
> 
> Ideally, any extra argument should be injectable by the user to allow customisation, so another field of type List could be added to the xml schema.
> Alternatively, the specific argument `--include_std_type` could be modelled as a boolean flag in the xml schema and possibly defaulted to true as I would imagine it to be quite common to use standard types in proto definitions.

(I took the liberty of renaming this issue to describe this alternative, since that is the direction @dotbg is taking in #1136)

dotbg:
I believe, this issue can be closed now, @raboof ?

raboof:
Yes, thanks again for a great contribution!

PR #1136, opened 2020-10-03
Fixes. #917. Added include_std_type configuration option.

dotbg:
References #917

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -129,7 +129,6 @@ lazy val interopTests = Project(id = "akka-grpc-interop-tests", base = file("int
   .settings(
     // All io.grpc servers want to bind to port :8080
     parallelExecution := false,
-    ReflectiveCodeGen.protocOptions := Seq("include_std_types"),
     ReflectiveCodeGen.generatedLanguages := Seq("Scala", "Java"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
     // setting 'skip in publish' would be more elegant, but we need
--- a/build.sbt
+++ b/build.sbt
@@ -129,6 +129,7 @@ lazy val interopTests = Project(id = "akka-grpc-interop-tests", base = file("int
   .settings(
     // All io.grpc servers want to bind to port :8080
     parallelExecution := false,
+    ReflectiveCodeGen.protocOptions := Seq("include_std_types"),
     ReflectiveCodeGen.generatedLanguages := Seq("Scala", "Java"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
     // setting 'skip in publish' would be more elegant, but we need
--- a/build.sbt
+++ b/build.sbt
@@ -213,7 +213,8 @@ lazy val pluginTesterScala = Project(id = "akka-grpc-plugin-tester-scala", base 
     skip in publish := true,
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"))
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"),
+    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
   .pluginTestingSettings
 
 lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base = file("plugin-tester-java"))
@@ -224,7 +225,8 @@ lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base = 
     ReflectiveCodeGen.generatedLanguages := Seq("Java"),
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"))
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
+    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
   .pluginTestingSettings
 
 lazy val root = Project(id = "akka-grpc", base = file("."))
--- a/gradle-plugin/build.gradle
+++ b/gradle-plugin/build.gradle
@@ -48,7 +48,7 @@ repositories {
 }
 
 dependencies {
-  implementation 'com.google.protobuf:protobuf-gradle-plugin:0.8.12'
+  implementation 'com.google.protobuf:protobuf-gradle-plugin:0.8.13'
   testImplementation('org.spockframework:spock-core:1.3-groovy-2.5')
 }
 
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
@@ -5,6 +5,12 @@ import org.gradle.api.Project
 
 class AkkaGrpcPluginExtension {
 
+    static final String PROTOC_VERSION = "3.11.0"
+
+    static final String PROTOC_PLUGIN_SCALA_VERSION = "2.12"
+
+    static final String GRPC_VERSION = "1.32.1"
+
     static final String PLUGIN_CODE = 'com.lightbend.akka.grpc.gradle'
 
     // workaround for tests, where there's no jar and MANIFEST.MF can't be read
@@ -16,9 +22,6 @@ class AkkaGrpcPluginExtension {
     boolean serverPowerApis = false
     boolean usePlayActions = false
     boolean includeStdTypes = false
-    String protocVersion = "3.11.4"
-    String grpcVersion = "1.32.1"
-    String protocPluginScalaVersion = "2.12"
 
     List<String> extraGenerators = []
 
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
@@ -7,12 +7,6 @@ class AkkaGrpcPluginExtension {
 
     static final String PLUGIN_CODE = 'com.lightbend.akka.grpc.gradle'
 
-    static final String PROTOC_VERSION = "3.4.0"
-
-    static final String PROTOC_PLUGIN_SCALA_VERSION = "2.12"
-
-    static final String GRPC_VERSION = "1.30.0"
-
     // workaround for tests, where there's no jar and MANIFEST.MF can't be read
     final String pluginVersion = System.getProperty("akkaGrpcTest.pluginVersion", AkkaGrpcPlugin.class.package.implementationVersion)
 
@@ -21,6 +15,11 @@ class AkkaGrpcPluginExtension {
     boolean generatePlay = false
     boolean serverPowerApis = false
     boolean usePlayActions = false
+    boolean includeStdTypes = false
+    String protocVersion = "3.11.4"
+    String grpcVersion = "1.32.1"
+    String protocPluginScalaVersion = "2.12"
+
     List<String> extraGenerators = []
 
     private final Project project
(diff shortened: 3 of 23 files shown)


plugin-tester-java/pom.xml
```diff
@@ -13,7 +13,10 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <akka.http.cors.version>0.4.2</akka.http.cors.version>
+    <maven-dependency-plugin.version>3.1.2</maven-dependency-plugin.version>
+    <maven-exec-plugin.version>3.0.0</maven-exec-plugin.version>
+    <akka.http.cors.version>1.1.0</akka.http.cors.version>
+    <akka.grpc.project.version>1.0.2-5-8b94eb72-20201003-1817</akka.grpc.project.version>
```
raboof, line 19: I think we shouldn't encode this in the pom.xml, but pass it in on the command line like `mvn -Dakka.grpc.project.version=1.0.2-5-8b94eb72-20201003-1817`

plugin-tester-java/src/main/protobuf/helloworld.proto
```diff
@@ -38,4 +40,5 @@ message HelloRequest {
 // The response message containing the greetings
 message HelloReply {
     string message = 1;
+    google.protobuf.Timestamp timestamp = 2;
```
raboof, line 43: This test project now fails for the non-maven builds (sbt and gradle):
```
helloworld.proto:3:1: Import "google/protobuf/timestamp.proto" was not found or had errors.
helloworld.proto:43:5: "google.protobuf.Timestamp" is not defined.
```

maven-plugin/src/main/scala/akka/grpc/maven/AbstractGenerateMojo.scala
```diff
@@ -186,7 +189,7 @@ abstract class AbstractGenerateMojo @Inject() (buildContext: BuildContext) exten
 
       val runProtoc: Seq[String] => Int = args =>
         com.github.os72.protocjar.Protoc.runProtoc(protocVersion +: args.toArray)
-      val protocOptions = Seq.empty
+      val protocOptions = if (includeStdTypes) Seq("--include_std_types") else Seq.empty
```
raboof, line 192: (another option would be to allow arbitrary additional protoc options, but this is :+1:)

raboof (commented):
Thank you for looking into this!

The change mostly looks good to me, but the plugin-tester-java project now fails because it is also ran as part of the gradle/sbt builds, and this feature is not yet available there. Of course the 'best' way to fix this would be to implement this feature for sbt and gradle as well, but I can imagine that is beyond the scope of this PR. In that case we should perhaps remove comment out the changes to plugin-tester-java, and add tickets to fill in the gradle and sbt support before enabling the test?

plugin-tester-java/pom.xml
dotbg: indeed.

maven-plugin/src/main/scala/akka/grpc/maven/AbstractGenerateMojo.scala
dotbg: this will be a bit more tough as the java library expect the arguments as if they were received from the command line. 
I want to avoid this. Besides, this can allow user to specify some of the arguments twice

plugin-tester-java/src/main/protobuf/helloworld.proto
dotbg: Could you, please, rename the issue. I will go for doing the same for the gradle and sbt as well

dotbg:
@raboof, I have fixed the gradle buld as well. The sbt looks like a tough one as I have to pass through the options to the sbt-protoc plugin.

raboof:
> @raboof, I have fixed the gradle buld as well.

Cool!

> The sbt looks like a tough one as I have to pass through the options to the sbt-protoc plugin.

Actually it turns out most of that was already in place, so I took the liberty of amending your PR with that.

gradle-plugin/build.gradle
```diff
@@ -12,7 +12,7 @@ group = "com.lightbend.akka.grpc"
 def tag = "git describe --tags".execute().text.substring(1).split("-g")[0].replace("\n", "")
 def finalVersion = (tag == versionDetails().lastTag.substring(1)) ? tag : tag.reverse().reverse() + "-" + versionDetails().gitHash.substring(0, 8)
 
-version = finalVersion
+version = "1.0.2-6-a54379e8-20201005-1902" //versionDetails()//finalVersion
```
raboof, line 15: Looks like this was accidentally committed, could you revert?

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
```diff
@@ -21,6 +15,11 @@ class AkkaGrpcPluginExtension {
     boolean generatePlay = false
     boolean serverPowerApis = false
     boolean usePlayActions = false
+    boolean includeStdTypes = false
+    String protocVersion = "3.11.4"
+    String grpcVersion = "1.32.1"
+    String protocPluginScalaVersion = "2.12"
```
raboof, line 21: I wonder if we can do even better, and instead of referring to constants here, refer to the generated `akka.grpc.codegen.BuildInfo` class here.

Is this refactor needed for this PR? If not it might be better to do it in a separate one.

plugin-tester-java/build.gradle
```diff
@@ -3,24 +3,27 @@ plugins {
...
 
+def scalaVersion = org.gradle.util.VersionNumber.parse(System.getenv("TRAVIS_SCALA_VERSION") ?: "2.12")
+def scalaBinaryVersion = "${scalaVersion.major}.${scalaVersion.minor}"
+
 akkaGrpc {
   generateClient = true
   generateServer = true
   serverPowerApis = true // This isn't default
+  includeStdTypes = true
+  grpcVersion = "1.32.1"
```
raboof, line 14: Is this now mandatory to specify here? I'd rather see it get inherited from the plugin.

plugin-tester-java/build.gradle
```diff
@@ -3,24 +3,27 @@ plugins {
...
 repositories {
   mavenLocal()
 }
 
-def scalaVersion = org.gradle.util.VersionNumber.parse(System.getenv("TRAVIS_SCALA_VERSION") ?: "2.12")
-def scalaBinaryVersion = "${scalaVersion.major}.${scalaVersion.minor}"
-
 dependencies {
-  implementation group: 'ch.megard', name: "akka-http-cors_${scalaBinaryVersion}", version: '0.4.2'
+  implementation group: 'ch.megard', name: "akka-http-cors_${scalaBinaryVersion}", version: '1.0.0'
```
raboof, line 25: :+1: good to make this consistent.

plugin-tester-java/pom.xml
```diff
@@ -13,7 +13,9 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <akka.http.cors.version>0.4.2</akka.http.cors.version>
+    <maven-dependency-plugin.version>3.1.2</maven-dependency-plugin.version>
+    <maven-exec-plugin.version>3.0.0</maven-exec-plugin.version>
+    <akka.http.cors.version>1.1.0</akka.http.cors.version>
```
raboof, line 18: In other places we're at 1.0.0, perhaps good to keep this consistent everywhere?

plugin-tester-java/src/main/java/example/myapp/helloworld/GreeterServer.java
```diff
@@ -7,6 +7,9 @@
 
 import akka.actor.ActorSystem;
 import akka.http.javadsl.*;
+import akka.http.javadsl.model.HttpRequest;
+import akka.http.javadsl.model.HttpResponse;
+import akka.japi.Function;
```
raboof, line 12: Are these needed?

plugin-tester-scala/settings.gradle
```diff
@@ -4,10 +4,10 @@ pluginManagement {
...
-    id 'com.lightbend.akka.grpc.gradle' version "${System.getProperty('akka.grpc.project.version')}"
+    id 'com.lightbend.akka.grpc.gradle' version "1.0.2-6-a54379e8-20201005-1902"
   }
 }
-
+/*
 if (!System.getProperty('akka.grpc.project.version')) {
   throw new GradleException("System property `akka.grpc.project.version` is not provided. Use -Dakka.grpc.project.version=\$VERSION.")
-}
+}*/
```
raboof, line 13: This was probably committed by accident?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -58,6 +58,7 @@ object AkkaGrpcPlugin extends AutoPlugin {
       case object Java extends Language
     }
 
+    val akkaGrpcProtocOptions = settingKey[Seq[String]]("Additional options to be passed to protoc")
```
raboof, line 61: I don't think we should add this setting: users can just add things to `PB.protocOptions` directly.

raboof (commented):
Some comments - starting to shape up nicely!

gradle-plugin/build.gradle
dotbg: sure, thanks.

gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPluginExtension.groovy
dotbg: I'd want to refer to the BuildInfo, but do not know the way to get it done. Perhaps, a code-generation task would help. Unfortunately, my sbt knowledge is not sufficient for this task.

I do need the 'includeStdTypes' variable. I will revert other changes.

plugin-tester-java/build.gradle
dotbg: no, it is not. the `includeStdTypes`  is false by default as I want to keep existing behaviour

plugin-tester-java/pom.xml
dotbg: indeed, I will do 1.1.0 everywhere.

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/src/main/protobuf/helloworld.proto
```diff
@@ -21,9 +23,11 @@ service GreeterService {
 // The request message containing the user's name.
 message HelloRequest {
     string name = 1;
+    google.protobuf.Timestamp timestamp = 2;
 }
 
 // The response message containing the greetings
 message HelloReply {
     string message = 1;
+    google.protobuf.Timestamp timestamp = 2;
```
raboof, line 32: Either remove the timestamp here or (perhaps better) add the option to the `PB.protocOptions` in the `build.sbt` of this test

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
dotbg: As long as we can give an example somewhere.

raboof: Yes, definitely! For example in one of the 'scripted' tests, and perhaps also linking a snippet from the docs?

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/src/main/protobuf/helloworld.proto
dotbg: I will add the options, as this will be able to serve as the documentation later.

build.sbt
```diff
@@ -129,6 +129,7 @@ lazy val interopTests = Project(id = "akka-grpc-interop-tests", base = file("int
   .settings(
     // All io.grpc servers want to bind to port :8080
     parallelExecution := false,
+    ReflectiveCodeGen.protocOptions := Seq("include_std_types"),
```
raboof, line 132: I think this one isn't needed
```suggestion
```

raboof (commented):
One last minor comment, then LGTM!

raboof (approved):
LGTM, thanks! (will merge when CI is green)

PR #1145, opened 2020-10-14
Make it configurable the dispatcher to use for akka grpc client

Tochemey:
The goal of this PR is to make it configurable the generated Akka gRPC client code to use a custom dispatcher rather than the actor system dispatcher where the client is running. These are some of the use cases:

- This will help to spawn another dispatcher to make gRPC calls in case someone decide to block the call within an actor which may block a thread in the main Akka dispatcher.

### DIFF
--- a/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
@@ -36,14 +36,14 @@ public abstract class @{service.name}Client extends @{service.name}ClientPowerAp
       private final Materializer mat;
       private final ExecutionContext ec;
 
-      private Default@{service.name}Client(GrpcClientSettings settings, ClassicActorSystemProvider sys) {
+      private Default@{service.name}Client(GrpcClientSettings settings, ClassicActorSystemProvider sys,  ExecutionContext ec) {
         this.settings = settings;
         this.mat = SystemMaterializer.get(sys).materializer();
-        this.ec = sys.classicSystem().dispatcher();
+        this.ec = ec;
         this.clientState = new ClientState(
           settings,
           akka.event.Logging$.MODULE$.apply(sys.classicSystem(), Default@{service.name}Client.class, akka.event.LogSource$.MODULE$.<Default@{service.name}Client>fromAnyClass()),
-          sys);
+          sys, ec);
         this.options = NettyClientUtils.callOptions(settings);
 
         sys.classicSystem().getWhenTerminated().whenComplete((v, e) -> close());
--- a/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
@@ -41,14 +41,13 @@ import akka.grpc.internal.ClientState
 trait @{service.name}Client extends @{service.name} with @{service.name}ClientPowerApi with AkkaGrpcClient
 
 object @{service.name}Client {
-  def apply(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider): @{service.name}Client =
+  def apply(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider, ec: ExecutionContext): @{service.name}Client =
     new Default@{service.name}Client(settings)
 }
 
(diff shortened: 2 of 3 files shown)


lightbend-cla-validator:
Hi @Tochemey,

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -36,14 +36,14 @@ public abstract class @{service.name}Client extends @{service.name}ClientPowerAp
       private final Materializer mat;
       private final ExecutionContext ec;
 
-      private Default@{service.name}Client(GrpcClientSettings settings, ClassicActorSystemProvider sys) {
+      private Default@{service.name}Client(GrpcClientSettings settings, ClassicActorSystemProvider sys,  ExecutionContext ec) {
```
raboof, line 39: I'm not sure customizing the executioncontext is sufficient here: a lot of the gRPC machinery is streams-based, and will use the system materializer. I didn't check, but I think the system materializer also runs on the default dispatcher.

In any case I think it should be an optional parameter, so we should also provide a version of the constructors that only takes the `sys`.

raboof (commented):
> This will help to spawn another dispatcher to make gRPC calls in case someone decide to block the call within an actor which may block a thread in the main Akka dispatcher.

I'm not entirely convinced this change will be worth it.

If there is blocking on the main dispatcher you will be in trouble anyway, so if you *must* block you should move *those actors* off the main dispatcher rather than the other way around.

To make this change effective I think it needs to be a bit bigger, also override the materializer and possibly impact more of the API's. I'm not sure the complexity that would introduce would weight up to the advantage of being able to separate this part of your application.

Do you have a test setup where you can actually see this make a meaningful difference?

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
Tochemey: @raboof Yeah I was making a grpc call from a persistent actor and got some thread starvation when we were load testing it. However instead of generating the grpc client I wrote them by hand as the generator and this time pass in a different executor context (I mean the scala version) instead of the main actor system dispatcher and things were pretty smooth.

Tochemey: Also I think if we can specify a different materializer for the client maybe that can help a bit.

raboof: > I think if we can specify a different materializer for the client maybe that can help a bit. That materializer will be bound to the lifecycle the akka grpc client.

Hmm, so instead of using the system materializer or letting the client provide their own materializer, you're suggesting creating a dedicated materializer 'inside' the client, tied to its lifecycle? That might be interesting.

raboof (commented):
I'm on the fence about this PR still. At minimum would need some work to make the client and the internal methods that are marked `@InternalStableApi` binary compatible. I fear that would create quite a lot of noise in the methods. Another downside of this change is that AFAICS users will have to make an implicit ExecutionContext available.

Tochemey:
@raboof by the way due to some mistakes I fortunately deleted the repo. However I have got it back here https://github.com/Tochemey/akka-grpc but it seems that I cannot fix the orphan PR.

raboof (commented):
I think it should be an optional parameter, so we should also provide a version of the constructors that only takes the sys.

It may be that for the Scala version that would be ambiguous - if that is the case then perhaps it could be an method on the companion object with a descriptive name?

Tochemey:
@raboof I will make the necessary adjustment per your recommendation and push to my branch. Any suggestion on how to fix the broken PR😁 to relink back to my repo?

raboof:
> Any suggestion on how to fix the broken PRgrin to relink back to my repo?

I'm not sure... perhaps the easiest way is to open a new PR?

Tochemey:
@raboof this is the fixed PR https://github.com/akka/akka-grpc/pull/1212.

raboof:
OK TO TEST

raboof:
https://travis-ci.org/github/akka/akka-grpc/jobs/735635143 shows some compile errors in the generated Scala code

Tochemey:
> https://travis-ci.org/github/akka/akka-grpc/jobs/735635143 shows some compile errors in the generated Scala code

Taking a look

Tochemey:
Will reopen it when necessary

Issue #1130, opened 2020-09-29
Update to sbt-protoc 1.0.0


raboof:
sbt-protoc 1.0.0 seems close (https://github.com/thesamet/sbt-protoc/releases), so it might be nice to make sure we're on that when we release Akka gRPC 1.1.0?

ennru:
Updated with #1241

PR #1146, opened 2020-10-14
Sbt protoc 1.0

Marcus-Rosti:
Small changes, needs common protos now & droped support for flag

References #1130

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -212,8 +212,7 @@ lazy val pluginTesterScala = Project(id = "akka-grpc-plugin-tester-scala", base 
     skip in publish := true,
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"),
-    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"))
   .pluginTestingSettings
 
 lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base = file("plugin-tester-java"))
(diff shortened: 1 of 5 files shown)


raboof:
OK TO TEST

project/Dependencies.scala
```diff
@@ -92,6 +96,7 @@ object Dependencies {
     Compile.akkaHttp2Support,
     Compile.akkaDiscovery,
     Compile.akkaHttpCors % "provided",
+    Compile.commonProtos,
```
raboof, line 99: Is this also needed when you're generating Java code, so not using scalapb? If so then perhaps it should come in via the generators instead of always being part of the runtime?

raboof (commented):
Great to keep on top of this - I'm much looking forward to the release.

Let's not merge this until there is a non-RC release, but great to prepare for that.

build.sbt
```diff
@@ -212,8 +212,7 @@ lazy val pluginTesterScala = Project(id = "akka-grpc-plugin-tester-scala", base
     skip in publish := true,
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"),
-    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
```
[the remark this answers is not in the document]
Marcus-Rosti: No, I get protoc errors saying that `--include_std_types` isn't a flag I can pass. I also saw that the source sbt-protoc documentation mentioned providing them this way: https://github.com/thesamet/sbt-protoc#additional-options 

I could just be wrong or it could be an artifact of the RC.

project/Dependencies.scala
Marcus-Rosti: Oh yeah I'll try that!

raboof:
> I saw those fixes in akka-http 10.2 so I'm excited to benefit from them so I thought I'd help push this out

Actually, Akka gRPC should work with Akka HTTP 10.2 just fine - it just doesn't require it.

raboof:
Thanks for the research here! The changes in this PR are now broken out into #1241 and #1231 so I think we can close this one.

PR #1174, opened 2020-11-10
Akka HTTP-based client

raboof:
This PR depends on a snapshot of Akka HTTP, and will help drive discovering the requirements for the HTTP/2 client features in Akka HTTP.


ignasi35 (commented):
Some initial feedback

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,226 @@
...
+                          case Chunk(data, _) =>
+                            data
+                          case LastChunk(_, trailer) =>
+                            trailerPromise.success(trailer)
+                            ByteString.empty
+                        }
+                        .watchTermination()((_, done) =>
+                          done.onComplete(c => {
+                            log.info(s"XXX response $response termination ${c.isSuccess}")
+                            trailerPromise.trySuccess(immutable.Seq.empty)
```
ignasi35, line 174: Should the trailer promise fail if the stream failed?

ignasi35: I'm not sure I understand the relation between `trailerPromise` and `completionFuture` with the whole stream.

I think it's unexpected complexity to capture the `trailers` value so it's available in the `GrpcResponseMetadata` below but I wonder if I'm missing something else.

ignasi35 (approved):
Did a final pass over the main stream code in `invokeWithMetadata`. Left a question but otherwise LGTM.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,226 @@
...
+          Uri(s"https://${settings.overrideAuthority.getOrElse(host)}/" + descriptor.getFullMethodName),
+          source)
+        Source.lazyFutureSource[O, Future[GrpcResponseMetadata]](() => {
+          singleRequest(httpRequest).map { response =>
+            {
+              Codecs.detect(response) match {
+                case Success(codec) =>
+                  log.info(s"XXX response $response started")
+                  implicit val reader: GrpcProtocolReader = GrpcProtocolNative.newReader(codec)
+                  val trailerPromise = Promise[immutable.Seq[HttpHeader]]()
```
ignasi35, line 153: Since the `trailerPromise` is not always related to the trailers (it may be redeemed if the upstream is completed) maybe we should have a comment or rename it?

(this is more relevant on #1179)

interop-tests/src/test/scala/akka/grpc/interop/GrpcInteropSpec.scala
```diff
@@ -7,20 +7,24 @@ package akka.grpc.interop
 import io.grpc.testing.integration.TestServiceHandlerFactory
 
 class GrpcInteropIoWithIoSpec extends GrpcInteropTests(IoGrpcJavaServerProvider, IoGrpcJavaClientProvider)
-class GrpcInteropIoWithAkkaScalaSpec extends GrpcInteropTests(IoGrpcJavaServerProvider, AkkaHttpClientProviderScala)
-class GrpcInteropIoWithAkkaJavaSpec extends GrpcInteropTests(IoGrpcJavaServerProvider, AkkaHttpClientProviderJava)
+class GrpcInteropIoWithAkkaScalaSpec extends GrpcInteropTests(IoGrpcJavaServerProvider, AkkaNettyClientProviderScala)
+class GrpcInteropIoWithAkkaJavaSpec extends GrpcInteropTests(IoGrpcJavaServerProvider, AkkaNettyClientProviderJava)
```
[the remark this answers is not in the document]
raboof: I do plan on adding those, but I want to keep this branch green, and there are still some tests that succeed with AkkaHttpScalaServer and AkkaHttpClient but fail with IoServer and AkkaHttpClient, so I'm postponing adding them until that's sorted.

runtime/src/main/scala/akka/grpc/internal/Marshaller.scala
```diff
@@ -5,16 +5,18 @@
...
 import io.grpc.KnownLength
-import akka.annotation.InternalApi
+import akka.annotation.InternalStableApi
 import akka.grpc.ProtobufSerializer
 
 /**
  * INTERNAL API
  */
-@InternalApi
+@InternalStableApi
```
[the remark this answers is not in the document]
raboof: It is used from generated clients

runtime/src/main/scala/akka/grpc/internal/Marshaller.scala
```diff
@@ -31,4 +33,6 @@ final class Marshaller[T <: scalapb.GeneratedMessage](u: ProtobufSerializer[T])
 
   override def stream(value: T): InputStream =
     new ByteArrayInputStream(value.toByteArray) with KnownLength
+
+  override def protobufSerializer: ProtobufSerializer[T] = u
```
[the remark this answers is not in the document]
raboof: possibly, TIAS :)

sbt-plugin/src/sbt-test/gen-scala-server/00-interop/src/main/resources/application.conf
```diff
@@ -0,0 +1 @@
+akka.http.server.preview.enable-http2 = on
```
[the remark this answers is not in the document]
raboof: This scripted test was changed to use the normal `Http` API instead of the now-dropped `Http2` API, which is why it didn't need the setting before but does now.

I'll port this change to a PR on master and then rebase this branch, to keep things together that should be together :)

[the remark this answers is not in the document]
raboof: #1177

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,226 @@
...
+          source: Source[I, NotUsed],
+          fqMethodName: String,
+          headers: MetadataImpl,
+          descriptor: MethodDescriptor[I, O],
+          streamingResponse: Boolean,
+          options: CallOptions): Source[O, Future[GrpcResponseMetadata]] = {
+        implicit val serializer: ProtobufSerializer[I] = descriptor
+        val deserializer: ProtobufSerializer[O] = descriptor
+        val httpRequest = GrpcRequestHelpers(
+          Uri(s"https://${settings.overrideAuthority.getOrElse(host)}/" + descriptor.getFullMethodName),
```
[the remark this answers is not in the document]
raboof: #1186

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * INTERNAL API
+   */
+  @InternalApi
+  @silent("never used")
+  def createChannel(settings: GrpcClientSettings, log: LoggingAdapter)(
+      implicit sys: ClassicActorSystemProvider): InternalChannel = {
+    implicit val ec = sys.classicSystem.dispatcher
+    implicit val writer = GrpcProtocolNative.newWriter(Codecs.supportedCodecs.head)
```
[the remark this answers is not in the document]
raboof: Actually this was wrong entirely, added #1187

runtime/src/main/scala/akka/grpc/internal/Marshaller.scala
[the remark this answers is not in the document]
raboof: See #1181

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
raboof: > I'm not sure I understand the relation between `trailerPromise` and `completionFuture` with the whole stream.

`completionFuture` makes sure the stream fails with the correct `io.grpc.StatusRuntimeException` when the trailer contains a non-successful status code.

> I think it's unexpected complexity to capture the `trailers` value so it's available in the `GrpcResponseMetadata` below but I wonder if I'm missing something else.

I agree it's unexpected complexity in the implementation, but I think it produces simplicity in the API.

In the 'regular' API, a failed streaming response is represented as the stream failing with a `io.grpc.StatusRuntimeException`. When using the 'power' API, you get access to the complete `akka.grpc.GrpcResponseMetadata` which contains all the detailed (leading and (asynchronously) trailing) header information.

raboof:
(after rebasing to use Akka HTTP 10.2.2, the `GrpcInteropIoWithAkkaHttpScalaSpec` `empty_stream` integration test fails with the suspicious failure message "OK". will look into that tomorrow)

raboof:
With the release of Akka HTTP 10.2.2 this could already be merged to master (as an experimental, undocumented, off-by-default feature). While we'd likely start another branch against Akka HTTP snapshots soon, this would reduce the diff and the chance of conflicts.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,299 @@
...
+      }
+
+    val (queue, doneFuture) =
+      Source
+        .queue[HttpRequest](4242, OverflowStrategy.fail)
+        .via(http2client)
+        .toMat(Sink.foreach { res =>
+          res.attribute(ResponsePromise.Key).get.promise.trySuccess(res)
+        })(Keep.both)
+        .run()
```
jrudolph, line 110: So, right now we create a single client and when that dies, the channel is dead?

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,299 @@
...
+      case None =>
+        new StatusRuntimeException(mapHttpStatus(response).withDescription("No grpc-status found"))
+      case Some(statusCode) =>
+        val description = allHeaders.find(_.name == "grpc-message").map(_.value)
+        new StatusRuntimeException(Status.fromCodeValue(statusCode.toInt).withDescription(description.orNull))
+    }
+  }
+
+  /**
+   * See https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md
```
jrudolph, line 279: Thanks for that comment :)

runtime/src/main/scala/akka/grpc/internal/InternalChannel.scala
```diff
@@ -5,14 +5,35 @@
...
+import akka.grpc.{ GrpcResponseMetadata, GrpcSingleResponse }
+import akka.stream.scaladsl.Source
+import io.grpc.{ CallOptions, MethodDescriptor }
 
 /**
  * INTERNAL API
  */
 @InternalApi
-case class InternalChannel(managedChannel: ManagedChannel, done: Future[Done])
+abstract class InternalChannel {
```
jrudolph, line 18: Nice seemed simple enough to get away from `ManagedChannel`. Do we also want to get away from `CallOptions` and `MethodDescriptor` at some point?

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -0,0 +1,56 @@
...
+@InternalStableApi
+abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
+    extends io.grpc.MethodDescriptor.Marshaller[T]
+    with WithProtobufSerializer[T] {
+  override def parse(stream: InputStream): T = {
+    val baos = new ByteArrayOutputStream(math.max(64, stream.available()))
+    val buffer = new Array[Byte](32 * 1024)
+
+    // Blocking calls underneath...
+    // we can't avoid it for the moment because we are relying on the Netty's Channel API
```
jrudolph, line 25: What will be the fix? Do we have an issue for that already? Is that already contained somehow? Probably it only affects the netty backend?

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/build.sbt
```diff
@@ -1,4 +1,5 @@
 resolvers += Resolver.sonatypeRepo("staging")
+resolvers += Resolver.bintrayRepo("akka", "snapshots")
```
jrudolph, line 2: All of these can probably be removed for now? Or do you want to keep them in there to simplify testing the next unreleased changes / improvements?

sbt-plugin/src/sbt-test/gen-java/03-no-java-package/build.sbt
```diff
@@ -1,4 +1,5 @@
 resolvers += Resolver.sonatypeRepo("staging")
+resolvers += Resolver.bintrayRepo("akka", "snapshots")
```
jrudolph, line 2: Maybe the tests can later be refactored to use a linked base `base-settings.sbt`? So, things like that can be more easily changed?

jrudolph (approved):
LGTM, a few questions and comments.

interop-tests/src/test/scala/akka/grpc/interop/GrpcInteropSpec.scala
```diff
@@ -33,14 +60,48 @@ object AkkaHttpServerProviderJava extends AkkaHttpServerProvider {
...
+      "custom_metadata",
+      "client_compressed_unary",
+      "client_compressed_streaming",
+      "server_compressed_unary")
+}
+
+object AkkaHttpClientProviderScala extends AkkaClientProviderScala("akka-http") {
+  val pendingCases =
+    Set(
+      "cancel_after_begin",
```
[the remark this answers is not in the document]
raboof: They are the same right now because the 3 client implementations now successfully support the same set of interop tests.

I'll move the set to `AkkaClientProviderScala`: when we implement further tests, I think we should strive to implement them for all backends at the same time. If for some reason we can't, then we can split the set back out then.

sbt-plugin/src/sbt-test/gen-java/03-no-java-package/build.sbt
raboof: you mean like a symlink? IIRC those are problematic on Windows...

sbt-plugin/src/sbt-test/gen-java/01-gen-basic-java/build.sbt
raboof: I think it's kind of nice to keep them for when we want to test with snapshot versions in different PR's as well

runtime/src/main/scala/akka/grpc/internal/InternalChannel.scala
raboof: > Do we also want to get away from `CallOptions` and `MethodDescriptor` at some point?

I don't mind those so much, they seem fairly reasonable.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,299 @@
...
+    // Configuring a different default could be a future feature.
+    // Configuring compression per call could be a future power API feature.
+    implicit val writer = GrpcProtocolNative.newWriter(Identity)
+
+    val target = {
+      // TODO FIXME discovery, loadbalancing etc
+      // https://github.com/akka/akka-grpc/issues/1196
+      // https://github.com/akka/akka-grpc/issues/1197
+      val resolved = settings.serviceDiscovery.lookup(settings.serviceName, 10.seconds)
+      Await.result(resolved, 10.seconds).addresses.head
```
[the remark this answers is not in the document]
raboof: Ultimately not, of course - but this should be replaced when we switch to a loadbalancing API, anyway, so that's something to solve at that point I think

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,299 @@
...
+        .via(http2client)
+        .toMat(Sink.foreach { res =>
+          res.attribute(ResponsePromise.Key).get.promise.trySuccess(res)
+        })(Keep.both)
+        .run()
+
+    def singleRequest(request: HttpRequest): Future[HttpResponse] = {
+      val p = Promise[HttpResponse]()
+      queue.offer(request.addAttribute(ResponsePromise.Key, ResponsePromise(p))).flatMap(_ => p.future)
+    }
```
raboof, line 115: All good points - but basically all of this should probably disappear in favor of something that also does loadbalancing, anyway.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -0,0 +1,299 @@
...
+          }
+
+        builder.withCustomHttpsConnectionContext(connectionContext).http2()
+      } else {
+        builder.http2WithPriorKnowledge()
+      }
+
+    val (queue, doneFuture) =
+      Source
+        .queue[HttpRequest](4242, OverflowStrategy.fail)
```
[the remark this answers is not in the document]
hepin1989: what if the queue is full?

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
raboof: Added #1229

PR #1176, opened 2020-11-10
Don't send a trailing request header

### LINKED ISSUE #1172: io.grpc server does not recognize the path in a gRPC request
When we make a request from the Akka gRPC client to the io.grpc grpc-java server, we see:

```
INBOUND HEADERS: streamId=1 headers=GrpcHttp2RequestHeaders[:path: /grpc.testing.TestService/EmptyCall, :authority: foo.test.google.fr, :method: POST, :scheme: https, content-type: application/grpc+proto, grpc-encoding: gzip, grpc-accept-encoding: gzip,identity, date: Tue, 10 Nov 2020 11:09:06 GMT] padding=0 endStream=false
OUTBOUND HEADERS: streamId=1 headers=DefaultHttp2Headers[:status: 404, content-type: text/plain; encoding=utf-8, grpc-status: 12, grpc-message: Expected path but is missing] padding=0 endStream=false
```

You'd say the path is right there

raboof:
Fixes #1172


ignasi35 (approved):
👍

Issue #935, opened 2020-04-22
[failed] akka.grpc.scaladsl.LoadBalancingIntegrationSpec


ignasi35:
I've seen unrelated PR's fail the test `akka.grpc.scaladsl.LoadBalancingIntegrationSpec`.

See https://travis-ci.org/github/akka/akka-grpc/jobs/678093242 (triggered by #933 "Update grpc-core, grpc-interop-testing, ... to 1.29.0")

See or https://github.com/ignasi35/akka-grpc/tree/require-akka-2.6 (bumps Akka to 2.6)

ignasi35:
> See ignasi35/akka-grpc@require-akka-2.6 (bumps Akka to 2.6)

Ths issue there is not a real issue. The test was not properly ported: the test requires 2 different `Materializer` instances and the code rewrite broke the requirement.

> See travis-ci.org/github/akka/akka-grpc/jobs/678093242 (triggered by #933 "Update grpc-core, grpc-interop-testing, ... to 1.29.0")

The issue was caused by a behavior change in the underlying `grpc-java`.

raboof:
https://travis-ci.org/github/akka/akka-grpc/jobs/682825264

raboof:
The non-flakey problems are now solved.

The test is still flakey: again at https://travis-ci.org/github/akka/akka-grpc/jobs/691288184

PR #1178, opened 2020-11-10
Avoid deadlock

raboof:
The 'concat' is used to make sure the stream fails with the right error
when the trailing headers contain an error status code. To avoid
deadlock we should put that after the watchTermination that successfully
terminates the stream where there are no trailing headers at all.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,7 +18,7 @@ object Dependencies {
     // https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
     val akka = "2.6.9"
     val akkaBinary = "2.6"
-    val akkaHttp = "10.2.1+92-764faf46"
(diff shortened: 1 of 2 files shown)


ignasi35:
> The 'concat' is used to make sure the stream fails with the right error
> when the trailing headers contain an error status code. To avoid
> deadlock we should put that after the watchTermination that successfully
> terminates the stream where there are no trailing headers at all.

There was something confusing on the implementation before this change but I coulnd't put my finger on it. This explains and sorts it.

ignasi35 (approved):
LGTM

raboof:
Looking at the logs, it seems this PR does fix a problem, but then surfaces akka/akka-http#3619 I'll clean up this PR and then we can merge this part.

raboof:
seems stable now (did a couple of rebuilds, only saw the #935 intermittent failure)

Issue #1180, opened 2020-11-11
Review byte[] to ByteString to byte[] conversions


ignasi35:
Building the Serialization and Marshalling abstractions hid some conversions from `byte[]` to `ByteString` or `ByteBuffer` and back.

For example, in `ProtoMarshaller` there's:

```
// u: ProtobufSerializer[T]
u.deserialize(akka.util.ByteString(baos.toByteArray))
```

and then on `ScalapbProtobufSerializer` we do:

```
  override def deserialize(bytes: ByteString): T =
    companion.parseFrom(CodedInputStream.newInstance(bytes.asByteBuffer))
```

where `CodedInputStream.newInstance(...)` already provides an overload for `byte[]` (so we can avoid the `ByteString` and the `ByteBuffer` instances).

viktorklang:
To avoid unnecessary copying, you might want to do:

```scala
// u: ProtobufSerializer[T]
u.deserialize(akka.util.ByteString.fromArrayUnsafe(baos.toByteArray))
```

ignasi35:
> To avoid unnecessary copying, you might want to do:
> 
> ```scala
> // u: ProtobufSerializer[T]
> u.deserialize(akka.util.ByteString.fromArrayUnsafe(baos.toByteArray))
> ```

Thanks @viktorklang! 

IIUC, that would save a copy of the byte array but still allocate a `ByteString` and a `ByteBuffer`.

ignasi35:
I just noticed `CodedInputStream.newInstance` also has an overload that takes an `InputStream` which is what `Marshaller#parse` receives.

So, in order to turn an `InputStream` into a `T` the current code in `Marshaller` combined with `ScalapbProtobufSerializer` does the following:

 - read the `InputStream` into memory as a `byte[]`
 - create a copy of the `byte[]` when creating a `ByteString`
 - create a `ByteBuffer` from the `ByteString`
 - create a `CodedInputStream` from that

Instead, we could try what performance impact (if any) the following had:

 - create a `CodedInputStream` from `InputStream`

And see if this approach, as opposed to the current abstraction, pays off.

It's a trade-off. 🤷🏼‍♂️

raboof:
> Instead, we could try what performance impact (if any) the following had

Definitely worth an experiment - though there's a chance that this method is indeed actually inefficient, see #1163

viktorklang:
👍--
Cheers,
√

lomigmegard:
[...]

PR #1181, opened 2020-11-11
Bring Marshaller and ProtoMarshaller closer

ignasi35:
Removing some duplicate code...


runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -28,11 +28,27 @@ final class Marshaller[T <: scalapb.GeneratedMessage](u: ProtobufSerializer[T])
...
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalStableApi
+final class Marshaller[T <: scalapb.GeneratedMessage](protobufSerializer: ProtobufSerializer[T])
+    extends BaseMarshaller[T](protobufSerializer) {
   override def stream(value: T): InputStream =
     new ByteArrayInputStream(value.toByteArray) with KnownLength
```
ignasi35, line 43: I was tempted to use:

```
    new ByteArrayInputStream(protobufSerializer.serialize(value).toArray) with KnownLength
```

and move this method to the base class but I'd be causing unnecessary allocations (see #1180)

runtime/src/main/scala/akka/grpc/internal/ProtoMarshaller.scala
```diff
@@ -1,38 +0,0 @@
...
-import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, InputStream }
-
-import io.grpc.KnownLength
-import akka.annotation.InternalStableApi
-import akka.grpc.ProtobufSerializer
-
-/**
- * INTERNAL API
- */
-@InternalStableApi
```
raboof, line 16: We can't remove this API since it's used from generated clients, and we want people to be able to upgrade the runtime library even when downstream libraries used a previous version to generate their clients. https://doc.akka.io/docs/akka-grpc/current/binary-compatibility.html

ignasi35: I'm not removing it, I'm only reimplementing it. First I moved this code into another file, and then I extracted the common code. MiMa is not complaining. :-)

ignasi35: correction: mima is complaining but the issue is unrelated to this discussion 😅

ignasi35:
After fixing some MiMa and giving it some thought, I'm not sure I like where this is going.

runtime/src/main/scala/akka/grpc/internal/ProtoMarshaller.scala
raboof: > I moved this code into another file

oops, sorry, missed that

ignasi35:
There's some failure in `GrpcInteropAkkaHttpScalaServerWithAkkaHttpScalaClientSpec` probably fixed on some other in-flight PR?

I'd keep this as a low-priority we can revisit sometime in the future.

raboof:
> There's some failure in `GrpcInteropAkkaHttpScalaServerWithAkkaHttpScalaClientSpec` probably fixed on some other in-flight PR?

working on that in #1178

Issue #1165, opened 2020-11-05
convert non-success HTTP status codes to gRPC status codes


raboof:
In the akka-http-client branch, `GrpcInteropAkkaHttpScalaServerWithAkkaHttpScalaClientSpec`, the `unimplemented_method` and `unimplemented_service` are currently ignored because they fail.

I suspect this is because the server returns an HTTP error code, and we're not converting that status code into a gRPC status code yet (https://grpc.github.io/grpc/core/md_doc_http-grpc-status-mapping.html).

PR #1182, opened 2020-11-11
Map HTTP status codes into gRPC failures.

ignasi35:
References #1165 

This goes on top of https://github.com/akka/akka-grpc/pull/1178

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
@@ -14,7 +14,7 @@ import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.grpc.GrpcProtocol.GrpcProtocolReader
 import akka.grpc.{ GrpcClientSettings, GrpcResponseMetadata, GrpcSingleResponse, ProtobufSerializer }
-import akka.http.scaladsl.model.HttpEntity.{ Chunk, ChunkStreamPart, Chunked, Default, LastChunk }
+import akka.http.scaladsl.model.HttpEntity.{ Chunk, Chunked, Default, LastChunk }
 import akka.http.scaladsl.{ ClientTransport, ConnectionContext, Http }
 import akka.http.scaladsl.model.{ AttributeKey, HttpHeader, HttpRequest, HttpResponse, RequestResponseAssociation, Uri }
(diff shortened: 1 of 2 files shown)


raboof:
https://github.com/akka/akka-grpc/pull/1182/commits/a3b4fa6cdaa17d5dc0c98e76571c9631d50390e0 looks pretty good. Does it need to be on top of #1178, or could it be PR'ed to `akka-http-client` on its own? #1178 needs some more work...

ignasi35:
> [a3b4fa6](https://github.com/akka/akka-grpc/commit/a3b4fa6cdaa17d5dc0c98e76571c9631d50390e0) looks pretty good. Does it need to be on top of #1178, or could it be PR'ed to `akka-http-client` on its own? #1178 needs some more work...

I've rebased to move it away from #1178. It looks like it does need to be on top of #1178 😒

raboof:
> > [a3b4fa6](https://github.com/akka/akka-grpc/commit/a3b4fa6cdaa17d5dc0c98e76571c9631d50390e0) looks pretty good. Does it need to be on top of #1178, or could it be PR'ed to `akka-http-client` on its own? #1178 needs some more work...
> 
> I've rebased to move it away from #1178. It looks like it does need to be on top of #1178 unamused

I don't think so. The flaky test failures on this branch are indeed the ones that #1178 are intended to fix, but they are already present on `akka-http-client`, so this PR can be merged independently of #1178.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -14,7 +14,7 @@ import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.grpc.GrpcProtocol.GrpcProtocolReader
 import akka.grpc.{ GrpcClientSettings, GrpcResponseMetadata, GrpcSingleResponse, ProtobufSerializer }
-import akka.http.scaladsl.model.HttpEntity.{ Chunk, Chunked, LastChunk }
+import akka.http.scaladsl.model.HttpEntity.{ Chunk, ChunkStreamPart, Chunked, Default, LastChunk }
```
raboof, line 17:
```suggestion
import akka.http.scaladsl.model.HttpEntity.{ Chunk, Chunked, Default, LastChunk }
```

PR #1189, opened 2020-11-12
Refactor: move providers to a dedicated file

ignasi35:
This PR changes no code. I'm only shuffling classes around so things are a bit more organized. I'm putting all server providers into a single file and same for client providers. We had some server providers or client providers in `main` and some in `tests`. Then, some of them were top level files and some were in a shared file.

I've moved almost everything to `main` prioritizing the ordering over the visibility. Some of stuff I touched must be accessible from scripted's `00-interop` and the separation between main and test on the `interop` project is not very important. 

Finally, I create `Servers` and `Clients`:

```scala
object Clients {
  val IoGrpc = IoGrpcJavaClientProvider
  object AkkaNetty {
    val Java = AkkaNettyClientProviderJava
    val Scala = AkkaNettyClientProviderScala
  }
  object AkkaHttp {
    // FIXME: let's have Scala stable and we'll do Java later.
    // val Java = AkkaNettyClientProviderJava
    val Scala = AkkaHttpClientProviderScala
  }
}
```

To more easily create the testing matrix. E.g. `class GrpcInteropIoWithIoSpec extends GrpcInteropTests(Servers.IoGrpc, Clients.IoGrpc)`


ignasi35:
This is growing beyond what I was planning. I'll close for the time being and revisit it some time in the future.

PR #1191, opened 2020-11-12
Reviews interop tests naming

ignasi35:
Standardize the spec names.

Introduce `Servers` and `Clients`.

PR #1192, opened 2020-11-12
Test Akka HTTP-based client against io.grpc server

raboof:
This is now possible since #1182 brought both implementations up to par


raboof:
#1191 also has this change, so not needed anymore

PR #1194, opened 2020-11-12
Handle strict responses

ignasi35:
In #1182 we introduced support for HttpReponses mapped as `Default` but some interop tests replay a `404` without payload so that's mapped into `Strict`. In either case, it's not a `Chunked` and we want to map the status.

I'm not sure a match-all solution is the correct one here though.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
@@ -197,7 +197,7 @@ object AkkaHttpClientUtils {
                                 .toJava
                           }))
                     case _ =>
-                      response.entity.discardBytes
+                      response.entity.discardBytes()
                       throw mapToStatusException(response, Seq.empty)
                   }
                 case Failure(e) =>
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
@@ -14,7 +14,7 @@ import akka.annotation.InternalApi
 import akka.event.LoggingAdapter
 import akka.grpc.GrpcProtocol.GrpcProtocolReader
 import akka.grpc.{ GrpcClientSettings, GrpcResponseMetadata, GrpcSingleResponse, ProtobufSerializer }
-import akka.http.scaladsl.model.HttpEntity.{ Chunk, Chunked, Default, LastChunk }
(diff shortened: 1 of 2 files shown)


raboof (approved):
The change looks reasonable to me.

(now that I look at this I think we want to make sure to actually consume the response entity as well, but that can happen in a different PR)

ignasi35:
For the record, here are more details. there are two different behaviors when the requests produce a 404 because of an unimplemented service:

1. Akka-gRPC Scala responds
```
HttpResponse(
404 Not Found,
Vector(
 Date: Thu, 12 Nov 2020 16:17:44 GMT, 
  Server: akka-http/10.2.1+103-664568a0
),
HttpEntity.Default(
  text/plain; charset=UTF-8,
  "The requested resource could not be found."
),
HttpProtocol(HTTP/2.0)
)
```

which maps to `HttpEntity.Default(text/plain; charset=UTF-8,42 bytes total)`

2. Akka-gRPC Java responds

```
HttpResponse(
404 Not Found,
Vector(
 Date: Thu, 12 Nov 2020 16:16:50 GMT, 
 Server: akka-http/10.2.1+103-664568a0
),
HttpEntity.Strict(
  none/none,
  0 bytes total
),
HttpProtocol(HTTP/2.0)
)
```

so the payload of the response is empty. that is mapped to `HttpEntity.Strict(none/none,0 bytes total)`.


3. finally, IoGrpc responds `HttpResponse(200 OK, grpc-status: 12 ...` which already maps to `GrpcStatus.UNIMPLEMENTED`.

ignasi35:
> (now that I look at this I think we want to make sure to actually consume the response entity as well, but that can happen in a different PR)

I've added it in this PR.

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -197,8 +196,9 @@ object AkkaHttpClientUtils {
                                 .map[akka.grpc.javadsl.Metadata](h => new JavaMetadataImpl(new HeaderMetadataImpl(h)))
                                 .toJava
                           }))
-                    case Default(_, _, _) => throw mapToStatusException(response, Seq.empty)
-                    case e                => throw new IllegalStateException(s"Expected chunked or default response but got $e")
+                    case _ =>
+                      response.entity.discardBytes
```
raboof, line 200: Can't we use `()` here? as we're calling this for its side-effect

```suggestion
                      response.entity.discardBytes()
```

raboof:
I think we should definitely support all 3 variations, so this PR is valuable.

It is odd that Akka-gRPC Scala and Java behave differently, though - might be worth looking into what's going on there.

PR #1198, opened 2020-11-13
Use discovered target

raboof:


runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -51,22 +52,33 @@ object AkkaHttpClientUtils {
...
+
         }))
     }
 
     val builder = Http()
-      .connectionTo(settings.overrideAuthority.getOrElse(host))
-      .toPort(settings.defaultPort)
+      .connectionTo(settings.overrideAuthority.getOrElse(target.host))
+      .toPort(target.port.getOrElse(settings.defaultPort))
       .withClientConnectionSettings(clientConnectionSettings)
```
ignasi35, line 82: If `settings.overrideAuthority` and `target.port` are defined, this builder would use `Http().connectionTo(settings.overrideAuthority).toPort(target.port)`.

Couldn't that trip the assertion on line 68? (`assert(port == settings.defaultPort)`)

I don't think there's a way to trip the other assertion with the current code.

This class is growing a bit. Maybe we could use this PR to extract the creation of the `builder` into a dedicated class/object that's easier to test/inject/extend?

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -51,22 +52,33 @@ object AkkaHttpClientUtils {
...
     implicit val writer = GrpcProtocolNative.newWriter(Identity)
 
-    // TODO FIXME discovery, loadbalancing etc
-    val host = settings.serviceName
+    val target = {
+      // TODO FIXME discovery, loadbalancing etc
+      // https://github.com/akka/akka-grpc/issues/1196
+      // https://github.com/akka/akka-grpc/issues/1197
+      val resolved = settings.serviceDiscovery.lookup(settings.serviceName, 10.seconds)
+      Await.result(resolved, 10.seconds).addresses.head
```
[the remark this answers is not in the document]
raboof: (I agree, but I'm not sure it's worth it yet as we'll likely replace this code when we have a connection management / loadbalancing API in HTTP/2)

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
raboof: > Couldn't that trip the assertion on line 68?

Good eye, that needed an update as well indeed

> This class is growing a bit. Maybe we could use this PR to extract the creation of the `builder` into a dedicated class/object that's easier to test/inject/extend?

I agree, we should probably move the `InternalChannel` anonymous subclass into its own file eventually, but since the connection management / loadbalancing might still cause code to move between the channel and the initialization logic I think we should do it when/after we implement that.

raboof:
PLS BUILD

ignasi35 (approved):

PR #1199, opened 2020-11-13
Add test and fix for scaladsl Power API

raboof:


runtime/src/main/scala/akka/grpc/internal/MetadataImpl.scala
```diff
@@ -159,12 +159,12 @@ class HeaderMetadataImpl(headers: immutable.Seq[HttpHeader] = immutable.Seq.empt
 
   override def getText(key: String): Option[String] =
     headers.reverseIterator.collectFirst {
-      case header if header.name == key => header.value
+      case header if header.name.equalsIgnoreCase(key) => header.value
```
jrudolph, line 162: In akka-http, we often convert the `key` first to lower case and then compare to the stored lower-case name in the header (using `header.is()`), but maybe it doesn't matter for performance too much here?

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerGreeterServiceImpl.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  override def streamHellos(in: Source[HelloRequest, NotUsed], metadata: Metadata): Source[HelloReply, NotUsed] = {
+    println(s"sayHello to stream...")
+    in.map(request => HelloReply(s"Hello, ${authTaggedName(request, metadata)}"))
+  }
+
+  // Bare-bones just for GRPC metadata demonstration purposes
+  private def isAuthenticated(metadata: Metadata): Boolean =
+    metadata.getText("authorization").nonEmpty
```
jrudolph, line 42: This is already lower case, so would it fail the test without the fix?

jrudolph (approved):
LGTM with an optimization suggestion

Do the tests also fail without the fix?

raboof:
> Do the tests also fail without the fix?

Yup!

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerGreeterServiceImpl.scala
raboof: IIRC yes, because the `.name` of the parsed modeled header was `Authorization` with a capital A

jrudolph: I see, because it's a modeled header.

runtime/src/main/scala/akka/grpc/internal/MetadataImpl.scala
```diff
@@ -157,16 +155,20 @@ class HeaderMetadataImpl(headers: immutable.Seq[HttpHeader] = immutable.Seq.empt
   private lazy val map: Map[String, List[MetadataEntry]] =
     MetadataImpl.toMap(asList)
 
-  override def getText(key: String): Option[String] =
+  override def getText(key: String): Option[String] = {
+    val lcKey = key.toLowerCase(Locale.ROOT)
```
jrudolph, line 159: :+1:

PR #1200, opened 2020-11-13
Support custom headers

raboof:


runtime/src/main/scala/akka/grpc/internal/MetadataImpl.scala
```diff
@@ -80,6 +81,13 @@ import akka.grpc.javadsl
     }
     mutableMetadata
   }
+
+  def asRawHeaders: immutable.Seq[HttpHeader] =
+    entries.map {
+      case (name, StringEntry(value)) => RawHeader(name, value)
+      // FIXME check if this is the right way to encode a binary header value
+      case (name, BytesEntry(value)) => RawHeader(name, value.utf8String)
```
ignasi35, line 89: From https://github.com/akka/akka-grpc/blob/d39b7f51290012e350250c67b5fb610df017e1bc/runtime/src/main/scala/akka/grpc/internal/GrpcEntityHelpers.scala#L60

```suggestion
      case (name, BytesEntry(value)) => RawHeader(name, RawHeader(key, MetadataImpl.encodeBinaryHeader(value)))
```
??

ignasi35 (approved):
LGTM (only the FIXME re `RawHeader`'s pending).

PR #1201, opened 2020-11-14
Support plaintext interop testing on AkkaNettyJava client

### LINKED ISSUE #1193: Improve support for plaintext usage in interop tests
https://github.com/akka/akka-grpc/pull/1175 introduced partial support to run some interop tests over plaintext.

Not all combinations of server and client are ready, though. For example, at the time of this writing, running all the interop suite over plaintext causes the following four tests to fail:

```
[error]         akka.grpc.interop.GrpcInteropAkkaJavaWithIoSpec
[error]         akka.grpc.interop.GrpcInteropIoWithAkkaNettyJavaSpec
[error]         akka.grpc.interop.GrpcInteropAkkaJavaWithAkkaNettyScalaSpec
[error]         akka.grpc.interop.GrpcInteropAkkaScalaWithAkkaNettyJavaSpec
```

Note that, `GrpcInteropAkkaJavaWithAkkaNettyJavaSpec` is not on the list because both server and client on that combination ignore the `use_tls=false` setting so the test run succeeds (but it doesn't run over plaintext).

ignasi35:
Fixes #1193


raboof (approved):
sweet ;)

PR #1207, opened 2020-11-18
Test future for streaming headers redeems eagerly

ignasi35:


interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -32,16 +43,30 @@ class PowerApiSpec(backend: String)
...
+      .bind(GreeterServicePowerApiHandler(new PowerGreeterServiceImpl(numberOfRepliesOverStreams)))
+      .futureValue
+
+  var client: GreeterServiceClient = _
+
+  after {
+    if (client != null && !client.closed.isCompleted) {
+      client.close().futureValue
+    }
+  }
```
ignasi35, line 65: The implementation using the netty client will add some noise on the logs if the client is not closed nicely.

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +109,41 @@ class PowerApiSpec(backend: String)
...
+        responseSource
+          .map { x =>
+            firstItem.countDown()
+            fiveItems.countDown()
+            x
+          }
+          .to(Sink.ignore)
+          .run
+
+      headers.isCompleted should be(false)
```
ignasi35, line 135: This first assertion is racy but after multiple runs I never saw it fail.

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +109,41 @@ class PowerApiSpec(backend: String)
...
+
+      firstItem.await(500, TimeUnit.MILLISECONDS)
+
+      headers.isCompleted should be(true)
```
raboof, line 139: `headers.futureValue` would be slightly less racy perhaps?

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +109,41 @@ class PowerApiSpec(backend: String)
...
+    "(on streamed calls) redeem the headers future as soon as they're available (and trailers future when trailers arrive)" in {
+      // invoking streamed calls using the power API materializes a Future[GrpcResponseMetadata]
+      // that should redeem as soon as the HEADERS is consumed. Then, the GrpcResponseMetadata instance
+      // contains another Future that will redeem when receiving the trailers.
+      client = GreeterServiceClient(
+        GrpcClientSettings.connectToServiceAt("localhost", server.localAddress.getPort).withTls(false))
+
+      val request = HelloRequest("Alice")
+      val responseSource: Source[HelloReply, Future[GrpcResponseMetadata]] =
+        client.itKeepsReplying().invokeWithMetadata(request)
```
raboof, line 122: `streamHellos` might have been even easier to be able to 'control' the timing of the responses (and response completion) from the test, but this is :+1: as well

raboof (approved):
Nice that apparently it worked without changes, good to have that assurance!

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +115,45 @@ class PowerApiSpec(backend: String)
...
+          .run
+
+      headers.isCompleted should be(false)
+
+      // As soon as the first item has arrived, the headers future
+      // should be redeemed but the fiveItems latch should still be open
+      // and the trailers future should not have been redeemed yet.
+      firstItem.await(500, TimeUnit.MILLISECONDS)
+      val trailers = headers.futureValue.trailers
+      fiveItems.getCount.toInt should be < 5
```
raboof, line 148: I guess this is a race condition as well, as the items can arrive all at once:

```
should (on streamed calls) redeem the headers future as soon as they're available (and trailers future when trailers arrive) *** FAILED *** (180 milliseconds)
[info]   5 was not less than 5 (PowerApiSpec.scala:148)
```

I'm not sure I see a 'quick fix' other than refactoring based on `streamHellos` (and sending the 'final' hello's only after validating the (leading) headers arrived)

ignasi35: > the items can arrive all at once

Ah, of course. I assumed throttling the items on the server would have been enough.

Making the server reply forever until a new message is sent from the client seems the best move forward. 👍🏼

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
ignasi35: I completely missed this comment yesterday. I'll keep it mind.

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +117,49 @@ class PowerApiSpec(backend: String)
...
+        .run
+
+      // blocks progress until redeeming `headers`
+      val trailers = headers.futureValue.trailers
+
+      // Don't send the finalization message until the headers future was redeemed (see above)
+      // and we've received a good amount of responses.
+      latch.await(5, TimeUnit.SECONDS)
+      trailers.isCompleted should be(false)
+      p.trySuccess(HelloRequest("ByeBye"))
```
ignasi35, line 154: As suggested by @raboof, the server doesn't finish until a marking message (in this case `HelloRequest("ByeBye")`) is received, and that finalization message is not sent until `headers` is redeemed.

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +117,49 @@ class PowerApiSpec(backend: String)
...
+        }
+        .to(Sink.ignore)
+        .run
+
+      // blocks progress until redeeming `headers`
+      val trailers = headers.futureValue.trailers
+
+      // Don't send the finalization message until the headers future was redeemed (see above)
+      // and we've received a good amount of responses.
+      latch.await(5, TimeUnit.SECONDS)
```
ignasi35, line 152: I've added a latch to, at least, require 200 responses to be processed before sending the finalization message.

raboof: Do we need this latch at all? If we can observe `headers.futureValue` before sending `p.trySuccess(HelloRequest("ByeBye")` that might sufficiently prove that the header is available before the entity body has completed?

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerGreeterServiceImpl.scala
```diff
@@ -32,11 +33,27 @@ class PowerGreeterServiceImpl(implicit system: ActorSystem) extends GreeterServi
...
+        // the client should at least see that one.
+        Iterator.single(req)
+      } else
+        Iterator.continually(req)
+    }
+
+    expandedSource.map(request => HelloReply(s"Hello, ${authTaggedName(request, metadata)}"))
   }
 
+  case object Tick
```
raboof, line 55: Tick?

raboof (approved):
Perhaps can be simplified a bit further now? But looks good

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +117,49 @@ class PowerApiSpec(backend: String)
...
+
+    "(on streamed calls) redeem the headers future as soon as they're available (and trailers future when trailers arrive)" in {
+
+      // invoking streamed calls using the power API materializes a Future[GrpcResponseMetadata]
+      // that should redeem as soon as the HEADERS is consumed. Then, the GrpcResponseMetadata instance
+      // contains another Future that will redeem when receiving the trailers.
+      client = GreeterServiceClient(
+        GrpcClientSettings.connectToServiceAt("localhost", server.localAddress.getPort).withTls(false))
+
+      val p = Promise[HelloRequest]
```
raboof, line 129:
```suggestion
      val p = Promise[HelloRequest]()
```

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -84,5 +117,49 @@ class PowerApiSpec(backend: String)
...
+      val latch = new CountDownLatch(minNumberOfResponses)
+      val i: AtomicInteger = new AtomicInteger(0)
+      val headers: Future[GrpcResponseMetadata] = responseSource
+        .map { x =>
+          i.incrementAndGet()
+          latch.countDown()
+          x
+        }
+        .to(Sink.ignore)
+        .run
```
raboof, line 145:
```suggestion
        .run()
```

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
ignasi35: you're right. I wanted to let more traffic traverse the connection but it's not necessary.

raboof (approved):
Nice!

PR #1209, opened 2020-11-19
codegen: do not extend from service for powerapis

jrudolph:
Having those `extends` service just to throw an exception seems like a potential trap. I haven't followed previous discussions about that, so maybe there was a good reason that things were done like that? I cannot see though how that would ever be useful the way it is right now.

There might be compatibility reasons not to change it right now.


ignasi35 (approved):
Yep. Looks like a left-over to me. (Unless there's some hidden cast somewhere)

raboof (approved):
Hmm, this could allow creating one API that 'transparently' takes any service implementation (power or not).

Let's indeed remove this now, and re-introduce it if we ever encounter a strong need for such an API.

raboof:
I think I remember the reason: when navigating a codebase, it is kind of nice to be able to go to the service interface, ask your IDE to find all implementations, and get both the 'simple' and power-api implementations.

I don't think that's reason enough to bring it back, though.

PR #1214, opened 2020-12-01
Trigger Link Validator

ennru:


ennru:
I required a few more fixes in the build...

But works now
https://github.com/akka/akka-grpc/pull/1214/checks?check_run_id=1479246377#step:7:40

PR #1215, opened 2020-12-01
run link-validator without git

raboof:
The git command-line tool is needed for sbt-dynver to find the latest
stable version to use with mima.


ennru (approved):
LGTM.

PR #1219, opened 2020-12-02
Fix scalafmt-sbt

raboof:


ignasi35:
This is blocked by #1217 which is blocked by this.

I'll take this commit there and push #1217 forward-

PR #1222, opened 2020-12-10
Handle strict responses

raboof:


runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -169,19 +175,37 @@ object AkkaHttpClientUtils {
...
+      response.map { response =>
+        {
+          if (response.status != StatusCodes.OK) {
+            response.entity.discardBytes()
+            val failure = mapToStatusException(response, immutable.Seq.empty)
+            Source.failed(failure).mapMaterializedValue(_ => Future.failed(failure))
+          } else {
+            Codecs.detect(response) match {
+              case Success(codec) =>
+                log.info(s"XXX response $response started")
```
viktorklang, line 200: @raboof leftover debugging code?

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -169,19 +175,37 @@ object AkkaHttpClientUtils {
...
+                implicit val reader: GrpcProtocolReader = GrpcProtocolNative.newReader(codec)
+                val trailerPromise = Promise[immutable.Seq[HttpHeader]]()
+                // Completed with success or failure based on grpc-status and grpc-message trailing headers
+                val completionFuture: Future[Unit] =
+                  trailerPromise.future.flatMap(trailers => parseResponseStatus(response, trailers))
+                completionFuture.foreach(_ => log.info(s"XXX response $response completion"))
```
viktorklang, line 206: @raboof leftover debugging code?

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -197,43 +221,47 @@ object AkkaHttpClientUtils {
...
+                responseData
+                  // This never adds any data to the stream, but makes sure it fails with the correct error code if applicable
+                  .concat(
+                    Source
+                      .maybe[ByteString]
+                      .mapMaterializedValue(promise => promise.completeWith(completionFuture.map(_ => None))))
+                  // Make sure we continue reading to get the trailing header even if we're no longer interested in the rest of the body
+                  .via(new CancellationBarrierGraphStage)
+                  .via(reader.dataFrameDecoder)
+                  .map(data => deserializer.deserialize(data))
```
viktorklang, line 240: To go with the same style as the line above?

```suggestion
                  .map(deserializer.deserialize)
```

jrudolph (approved):
LGTM, seems straight-forward enough. Was that needed because of one of the optimizations we did, where we can now generate strict entities if possible? Were those responses generated by our server or but by another one?

raboof:
> Was that needed because of one of the optimizations we did, where we can now generate strict entities if possible?

Indeed!

> Were those responses generated by our server or but by another one?

This was triggered when testing against the io.grpc server implementation.

PR #1223, opened 2020-12-10
Akka HTTP-based client with 'persistent' connections

raboof:
The first 2 tests succeed, the tests that require failover to
another server (as expected) still fail.

Requires https://github.com/akka/akka-http/pull/3685

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
@@ -11,7 +11,7 @@ import akka.grpc.internal.ClientConnectionException
 import akka.grpc.scaladsl.tools.MutableServiceDiscovery
 import akka.http.scaladsl.Http
 import akka.stream.SystemMaterializer
-import com.typesafe.config.{ Config, ConfigFactory }
+import com.typesafe.config.ConfigFactory
 import example.myapp.helloworld.grpc.helloworld._
 import io.grpc.Status.Code
 import io.grpc.StatusRuntimeException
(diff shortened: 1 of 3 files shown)


interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
```diff
@@ -51,6 +60,27 @@ class NonBalancingIntegrationSpec extends AnyWordSpec with Matchers with BeforeA
...
+      for (i <- 1 to 50) {
+        client.sayHello(HelloRequest(s"Hello $i")).futureValue
+      }
+      server1.terminate(5.seconds).futureValue
+      // And restart
+      Http().newServerAt("127.0.0.1", server1.localAddress.getPort).bind(GreeterServiceHandler(service1)).futureValue
+
+      for (i <- 1 to 50) {
+        client.sayHello(HelloRequest(s"Hello $i")).futureValue
+      }
```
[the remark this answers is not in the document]
raboof: I wanted to simulate 'an actively used server', throwing more requests at it perhaps has a better chance of hitting corner cases ;)

ignasi35 (approved):
looking good

runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
```diff
@@ -53,33 +53,29 @@ object AkkaHttpClientUtils {
...
-      case Some(authority) =>
-        ClientConnectionSettings(sys).withTransport(ClientTransport.withCustomResolver((host, port) => {
+    val random = new Random()
+    val clientConnectionSettings =
+      ClientConnectionSettings(sys).withTransport(ClientTransport.withCustomResolver((host, _) => {
+        settings.overrideAuthority.foreach { authority =>
           assert(host == authority)
-          assert(port == target.port.getOrElse(settings.defaultPort))
+        }
+        settings.serviceDiscovery.lookup(settings.serviceName, 10.seconds).map { resolved =>
```
jrudolph, line 66: Is that the part that fixes the fail-over?

raboof: Basically yes: by doing the discovery inside the resolver, it is performed again for each materialization of the connection. A bit coarse perhaps, but does the trick for now here.

PR #1231, opened 2020-12-30
prepare for ScalaPB 0.10.10

raboof:


raboof:
PLS BUILD

codegen/src/main/scala/akka/grpc/gen/scaladsl/Service.scala
```diff
@@ -5,9 +5,10 @@
 package akka.grpc.gen.scaladsl
 
 import scala.collection.immutable
-
 import scala.collection.JavaConverters._
 import com.google.protobuf.Descriptors._
+//import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
```
ennru, line 10: Remove.

project/plugins.sbt
```diff
@@ -31,4 +31,4 @@ libraryDependencies += "org.eclipse.jgit" % "org.eclipse.jgit" % "5.10.0.2020120
 // scripted testing
 libraryDependencies += "org.scala-sbt" %% "scripted-plugin" % sbtVersion.value
 
-libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.10.9"
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.10.10"
```
ennru, line 34: Is this required to be in sync with sbt-protoc?
If so add a note about it.

maven-plugin/src/main/scala/akka/grpc/maven/AbstractGenerateMojo.scala
```diff
@@ -203,11 +203,13 @@ abstract class AbstractGenerateMojo @Inject() (buildContext: BuildContext) exten
...
-      protocbridge.ProtocBridge.run(
-        protocCommand,
+      protocbridge.ProtocBridge.execute(
+        ProtocRunner.fromFunction((args, _) => protocCommand(args)),
         targets,
         Seq(incPath) ++ protocOptions ++ schemas.map(_.getCanonicalPath),
-        pluginFrontend = protocbridge.frontend.PluginFrontend.newInstance)
+        artifact =>
+          throw new RuntimeException(
+            s"The version of sbt-protoc you are using is incompatible with '${artifact}' code generator. Please update sbt-protoc to a version >= 0.99.33"))
```
ennru, line 212: We'll do the sbt-protoc upgrade, will this still apply?

ennru (approved):
LGTM.

project/plugins.sbt
raboof: not strictly, I think

maven-plugin/src/main/scala/akka/grpc/maven/AbstractGenerateMojo.scala
raboof: Yes, recent versions of sbt-protoc should not touch this parameter.

(this logic was inside sbt-protoc before, but that constructor became private, so I applied it here as well)

PR #1233, opened 2021-01-08
Telemetry hook

raboof:
WIP: only scaladsl so far

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/TelemetrySpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/TelemetrySpec.scala
@@ -1,0 +1,50 @@
+package akka.grpc.scaladsl
+
+import akka.actor.ActorSystem
+import akka.grpc.internal.{ GrpcProtocolNative, GrpcRequestHelpers, Identity, TelemetryExtension, TelemetrySpi }
+import akka.http.scaladsl.model.HttpRequest
+import akka.stream.scaladsl.Source
+import akka.testkit.TestKit
+import com.typesafe.config.ConfigFactory
+import example.myapp.helloworld.grpc.helloworld.{ GreeterService, GreeterServiceHandler, HelloRequest }
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+class TelemetrySpec
+    extends TestKit(
+      ActorSystem(
+        "TelemetrySpec",
+        ConfigFactory
+          .parseString(s"""akka.grpc.telemetry-class = "akka.grpc.scaladsl.CollectingTelemetrySpi" """)
+          .withFallback(ConfigFactory.load())))
+    with AnyWordSpecLike
+    with Matchers
+    with ScalaFutures {
+  "The client-side telemetry hook" should {
+    "pick up matched requests" in {
+      val handler = GreeterServiceHandler(new CountingGreeterServiceImpl)
+      implicit val ser = GreeterService.Serializers.HelloRequestSerializer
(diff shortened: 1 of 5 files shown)


runtime/src/main/scala/akka/grpc/internal/TelemetrySpi.scala
```diff
@@ -0,0 +1,59 @@
...
+          NoOpTelemetry
+      }
+    }
+  }
+}
+
+@InternalStableApi
+trait TelemetrySpi {
+  @nowarn
+  def onRequest(prefix: String, method: String, request: HttpRequest): Unit = {}
```
johanandren, line 55: May need to be able to modify request and return a new one to pass context. You'll see later.

johanandren (commented):
Looks good so far.

runtime/src/main/scala/akka/grpc/internal/TelemetrySpi.scala
raboof: :+1: I suspect not for *this* feature, but indeed would be a nice generalization

runtime/src/main/scala/akka/grpc/internal/TelemetrySpi.scala
```diff
@@ -52,7 +52,7 @@ private[internal] object TelemetrySpi {
 @InternalStableApi
 trait TelemetrySpi {
   @nowarn
-  def onRequest(prefix: String, method: String, request: HttpRequest): Unit = {}
+  def onRequest[T <: HttpRequest](prefix: String, method: String, request: T): T = request
```
johanandren, line 55: Why was the type bound needed, something on the "other side"?

interop-tests/src/test/scala/akka/grpc/scaladsl/TelemetrySpec.scala
```diff
@@ -37,14 +40,16 @@ class TelemetrySpec
...
       prefix should be(GreeterService.name)
       method should be("SayHello")
-      request.entity.contentType should be(GrpcProtocolNative.contentType)
+      request.entity.getContentType should be(GrpcProtocolNative.contentType)
     }
   }
 }
 
 class CollectingTelemetrySpi extends TelemetrySpi {
   var requests: List[(String, String, HttpRequest)] = Nil
```
johanandren, line 49: @volatile perhaps if it should be safe to observe from another thread

raboof: this is just the spec, but yeah, good to add

runtime/src/main/scala/akka/grpc/internal/TelemetrySpi.scala
raboof: yeah, so we can keep using the scala HttpRequest as a scala HttpRequest after going through the SPI

runtime/src/main/scala/akka/grpc/internal/TelemetrySpi.scala
```diff
@@ -0,0 +1,63 @@
...
+  ExtensionIdProvider
+}
+import akka.annotation.{ InternalApi, InternalStableApi }
+import akka.http.javadsl.model.HttpRequest
+
+import scala.annotation.nowarn
+
+class TelemetryExtensionImpl(val spi: TelemetrySpi) extends Extension
+
+object TelemetryExtension extends ExtensionId[TelemetryExtensionImpl] with ExtensionIdProvider {
```
ignasi35, line 22:
```
/** INTERNAL API */
@InternalStableApi
private[grpc] ...
```

(both class and object)

ignasi35 (approved):

johanandren (approved):

Issue #239, opened 2018-06-20
Code generation parameter: map method names


raboof:
When a protocol uses method names that clash with method names already reserved by the JVM, we should allow configuring the codegen to avoid those names in the generated code (but still use them in the protocol logic).

Look at how protoc/scalapb handle this for model fields for inspiration.

johanandren:
Or maybe fail fast on those names. To clarify for example `notify`, `notifyAll` are kind of easy to forget are taken and could also easily overlap with a problem domain, `toString`, `hashCode`, `wait`, `finalize`, `getClass`, `equals` maybe less so but still nice if it failed fast with a clear error about not using those.

jroper:
I encountered this today:

```
Caused by: java.lang.IllegalArgumentException: requirement failed: The method name `import` is a reserved word in Java, please change it in your proto
```

The problem is, if it's a grpc method, you can't change the name if you don't control the service you're talking to. The gRPC method name is and must be `Import`, otherwise you can't invoke the service. The official protobuf compiler for Java will append an underscore in these instances.

PR #1235, opened 2021-01-11
Fail fast on language keywords in method names

andreaTP:
References #239

Tested with scripted tests since I haven't found a more suitable place.
This can be a first proposal that provides the user a better error message (it's going to fail the build anyhow).
We can, later on, decide to backtick those keywords in Scala.


raboof:
OK TO TEST

raboof (approved):
I agree, nice improvement already!

raboof:
> We can, later on, decide to backtick those keywords in Scala

Oh, cool, this is already in progress in #1236

andreaTP:
Minor issue:
Travis keeps failing because my fork master branch is pretty old and sbt get a `previousStableVersion` that doesn't exists.
Ref.
https://github.com/akka/akka-grpc/blob/master/build.sbt#L54

andreaTP:
Is Jenkins stuck?

raboof:
OK TO TEST

codegen/src/main/scala/akka/grpc/gen/scaladsl/Method.scala
```diff
@@ -18,6 +18,10 @@ case class Method(
     comment: Option[String] = None)(implicit ops: DescriptorImplicits) {
   import Method._
 
+  require(
+    !ReservedWords.contains(name),
```
sebastian-alfers, line 22: All scala related fail-fast logic should be removed right?

andreaTP: yes!

andreaTP: I'll rebase after #1236 get merged.

andreaTP:
rebased on latest master.

raboof (approved):
Thanks for the rebase!

PR #1236, opened 2021-01-11
Generate method names with backquotes for reserved keywords in Scala

sebastian-alfers:
Fixes the generation of method names that clash with reserved keywords. E.g.:

```
syntax = "proto3";

service HelloWorldService {
  rpc Match(...) returns (...) {};
}
```

would generate something like 
```
def `match`(...) 
```

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/Method.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/Method.scala
@@ -48,6 +48,12 @@ case class Method(
       case (true, true)   => BidiStreaming
     }
   }
+
+  val nameSafe = {
+    if (name == "match") {
+      s"""`$name`""".stripMargin
+    } else name
+  }
 }
 
 object Method {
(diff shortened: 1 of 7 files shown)


sebastian-alfers:
Goal of this PR is also to start a discussion if this is something we actually want to have or not.

patriknw (approved):
LGTM

andreaTP:
@sebastian-alfers do you wanna integrate the fail-fast part for Java implemented here:
https://github.com/akka/akka-grpc/pull/1235
?

sebastian-alfers:
@andreaTP I would more than happy to include your work on this PR.

However I was thinking if we want keep both contributions in separate PRs as they are independent features. WDYT?

I am happy with both :-)

andreaTP:
@sebastian-alfers as you prefer, I'm not going to be upset if you just copy-paste my part in your PR, otherwise I will follow up rebasing on top of yours 🙂

sebastian-alfers:
Do avoid a decision-deadlock here I think we can go with separate PRs :-)

raboof (approved):

raboof:
Nice, thanks!

Issue #1238, opened 2021-01-12
protoc doesn't support Apple silicon/M1


ennru:
### Short description

Protobuf's `protoc` is not published pre-compiled for Apple Silicon/M1 https://github.com/protocolbuffers/protobuf/issues/8062

```
protoc-jar: protoc version: 3.11.4, detected platform: osx-aarch_64 (mac os x/aarch64)
...
[error] stack trace is suppressed; run last Compile / protocGenerate for the full output
[error] (Compile / protocGenerate) error occurred while compiling protobuf files: Unsupported platform: protoc-3.11.4-osx-aarch_64.exe
```

ScalaPB instructs users to compile it locally https://github.com/scalapb/ScalaPB/issues/1024

ennru:
With the latest sbt-protoc users may choose to use a local install of protoc. The Akka gRPC docs point to the necessary settings now.

johanandren:
aarch64 support is included with newer protoc artifacts (3.18.1 for example)

raboof:
> aarch64 support is included with newer protoc artifacts (3.18.1 for example)

(which will be the default in Akka gRPC starting from version 2.1.2 onwards)

PR #1241, opened 2021-01-18
ScalaPB 1.0 and document use of local protoc for OSX ARM

ennru:
ScalaPB 1.0 (in RC right now) allows setting the `protoc` executable so that a locally built version can be used.

References #1238

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -214,8 +214,7 @@ lazy val pluginTesterScala = Project(id = "akka-grpc-plugin-tester-scala", base 
     skip in publish := true,
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"),
-    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"))
   .pluginTestingSettings
 
 lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base = file("plugin-tester-java"))
@@ -226,8 +225,7 @@ lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base = 
     ReflectiveCodeGen.generatedLanguages := Seq("Java"),
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
-    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
+    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"))
   .pluginTestingSettings
 
 lazy val root = Project(id = "akka-grpc", base = file("."))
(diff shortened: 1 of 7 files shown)


raboof:
I'm really looking forward to sbt-protoc 1.0.0, it will also fix running protoc without further workarounds on NixOS \o/

build.sbt
```diff
@@ -226,8 +225,7 @@ lazy val pluginTesterJava = Project(id = "akka-grpc-plugin-tester-java", base =
     ReflectiveCodeGen.generatedLanguages := Seq("Java"),
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := scala212,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("server_power_apis"),
-    Compile / ReflectiveCodeGen.protocOptions += "--include_std_types")
```
raboof, line 230: why did you remove this here? did it become the default behavior?

ennru: I tried to find something in the docs, but didn't succeed.
`protoc` doesn't accept the flag anymore.

ennru:
This is similar to #1146

project/Dependencies.scala
```diff
@@ -73,6 +73,10 @@ object Dependencies {
     val logback = "ch.qos.logback" % "logback-classic" % "1.2.3" % "runtime" // Eclipse 1.0
   }
 
+  object Protobuf {
+    val googleCommonProtos = "com.thesamet.scalapb.common-protos" %% "proto-google-common-protos-scalapb_0.10" % "1.18.1-0" % "protobuf"
```
ennru, line 77: Is there a need/way to deduct the versions (0.10, 1.18.1-0) from the protoc plugin?

project/Common.scala
```diff
@@ -74,5 +75,9 @@ object Common extends AutoPlugin {
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-oDF"),
     crossScalaVersions := Seq(scala212, scala213),
     mimaReportSignatureProblems := true,
-    scalafmtOnCompile := true)
+    scalafmtOnCompile := true,
+    // https://github.com/akka/akka-grpc/issues/1238
+    PB.protocExecutable := (if (protocbridge.SystemDetector.detectedClassifier() == "osx-aarch_64")
+                              file("/usr/local/bin/protoc")
```
ennru, line 81: This should be configurable. Hopefully ScalaPB can offer a good solution for it.

project/Dependencies.scala
raboof: not AFAICS

project/Common.scala
raboof: it's configurable by overriding `PB.protocExecutable` - perhaps that is sufficient?

ennru: Yes, that should be enough. A note/pointer in the Akka gRPC docs would be good.

docs/src/main/paradox/buildtools/sbt.md
```diff
@@ -106,7 +117,7 @@ If you want to use TLS-based negotiation on JDK 8 versions prior to
 [1.8.0_251](https://www.oracle.com/java/technologies/javase/8u251-relnotes.html),
 the server requires a special Java agent for ALPN.
  
-See the [Akka HTTP docs about HTTP/2](https://doc.akka.io/docs/akka-http/10.1/server-side/http2.html#application-layer-protocol-negotiation-alpn-))
+See the @extref:[Akka HTTP docs about HTTP/2](akka-http:server-side/http2.html#application-layer-protocol-negotiation-alpn-)
```
raboof, line 120: :+1:

raboof (approved):

PR #1247, opened 2021-01-22
Clarify release docs: don't forget --tags

raboof:


ennru (approved):

raboof:
(fails because trying to mima-check against the not-yet-propagated release 1.1.0-M1 release, so waiting a bit)

Issue #1237, opened 2021-01-11
grpc MethodDescriptor-s as public val-s


justcoon:
Grpc method descriptors  (io.grpc.MethodDescriptor) are now only as private variables, in code, generated for grpc client (https://github.com/akka/akka-grpc/blob/master/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt)

Is it possible to generate them as public val-s/constants in same way as generated Serializers?

Gatling grpc plugin (https://github.com/phiSgr/gatling-grpc) using those method descriptors. In addition, for this use case, scalpb grpc generation needs to be executed.

Thanks for answer.

raboof:
Thanks for the ping! The reason we've kept these variables private is because we're not entirely sure we want to commit ourselves to keep providing them in the future.

As a workaround I think you could add the `grpc` option to the generator and that would generate classes that provide access to the io.grpc.MethodDescriptor instances (https://github.com/akka/akka-grpc/pull/1108)

We've seen this question before, though, so perhaps we should expose the descriptors as public fields marked `@ApiMayChange` so they can be used, but we don't have to commit to keeping them long-term just yet. Would you be interested in creating a PR to that effect?

justcoon:
@raboof thanks for answer

I tried (sbt-akka-grpc 1.0.2)

`akkaGrpcCodeGeneratorSettings += "grpc,server_power_apis"`

but it  did not generated ServiceGrpc

it is like grpc setting do nothing

I am interested in creating PR with this feature, do you have some proposal how it should look like?

my proposal:

```scala
package com.user.api

import akka.grpc.internal.Marshaller

import io.grpc.MethodDescriptor

object UserApiService extends akka.grpc.ServiceDescription {
  val name = "com.user.api.proto.UserApiService"

  val descriptor: com.google.protobuf.Descriptors.FileDescriptor =
    com.user.api.proto.UserApiProto.javaDescriptor;

  object Serializers {

    import akka.grpc.scaladsl.ScalapbProtobufSerializer

    val RegisterUserReqSerializer = new ScalapbProtobufSerializer(com.user.api.proto.RegisterUserReq.messageCompanion)
    val RegisterUserResSerializer = new ScalapbProtobufSerializer(com.user.api.proto.RegisterUserRes.messageCompanion)
  }

  object Methods {
    val RegisterUserDescriptor: MethodDescriptor[com.user.api.proto.RegisterUserReq, com.user.api.proto.RegisterUserRes] =
[...]

PR #1250, opened 2021-01-30
public grpc method descriptors in generated code #1273

justcoon:
grpc MethodDescriptors as public val-s in generated code

References #1237


lightbend-cla-validator:

raboof:
OK TO TEST

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -93,32 +90,6 @@ final class Default@{service.name}Client(settings: GrpcClientSettings)(implicit
...
-@mapMethodType(methodType: akka.grpc.gen.MethodType) = {
-  @if(methodType == akka.grpc.gen.Unary) { MethodDescriptor.MethodType.UNARY }
-  @if(methodType == akka.grpc.gen.ClientStreaming) {MethodDescriptor.MethodType.CLIENT_STREAMING }
-  @if(methodType == akka.grpc.gen.ServerStreaming) { MethodDescriptor.MethodType.SERVER_STREAMING }
-  @if(methodType == akka.grpc.gen.BidiStreaming) {MethodDescriptor.MethodType.BIDI_STREAMING }
-}
-
-private object Default@{service.name}Client {
-
-  def apply(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider): @{service.name}Client =
```
raboof, line 105: It looks like removing this object caused MiMa to flag a potential binary incompatibility. I think this might probably be all-right since this object is private anyway, but perhaps double-check this apply method would not be used from user code. If so then we can add a mima exclusion rule to specify this is all-right.

justcoon: i removed it, because it was failing on compilation, 
after i moved method descriptors, there was just apply function which was not used

ignasi35:
LGTM once the MiMa issue is fixed.

justcoon:
@raboof: do you have some idea/proposal, how to solve MiMa issue?

ignasi35:
> @raboof: do you have some idea/proposal, how to solve MiMa issue?

Hi @justcoon, you have several options:

1. restore the `apply()` method
2. filter that method from the mima verification using the `ProblemFilter`detailed on the [error message](https://travis-ci.com/github/akka/akka-grpc/jobs/478730662#L328-L333) following the example of [mima filters in other modules](https://github.com/akka/akka-grpc/blob/master/runtime/src/main/mima-filters/1.0.x.backwards.excludes/akka-http2-client.excludes)

But I think @raboof's recommendation to first validate the `apply()` method is not used from user code would be best.

raboof (approved):
Thanks!

raboof:
this means a binary incompatibility between the client and the api trait, but I think it's reasonable to require those to be generated with the same version of Akka gRPC.

PR #1253, opened 2021-01-31
support ScalaPB package/file-level customization

bjaglin:
See https://scalapb.github.io/docs/customizations

On master, code won't compile if package/file-level options are used to change `flat_package` as only ScalaPB honors the request.

### DIFF
--- a/sbt-plugin/src/sbt-test/gen-scala-server/09-scalapb-customizations/test
+++ b/sbt-plugin/src/sbt-test/gen-scala-server/09-scalapb-customizations/test
@@ -1,0 +1,5 @@
+> protocGenerate
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/HelloRequest.scala
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/HelloworldProto.scala
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/GreeterService.scala
+> compile
--- a/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/CodeGenerator.scala
@@ -4,6 +4,8 @@
 
 package akka.grpc.gen
 
+import com.github.ghik.silencer.silent
+import com.google.protobuf.ExtensionRegistry
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorResponse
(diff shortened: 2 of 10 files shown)


raboof:
OK TO TEST

raboof:
> On master, code won't compile if package/file-level options are used to change `flat_package` as only ScalaPB honors the request.

Indeed the Akka gRPC code generation currently only allows changing `flat_package` as a 'build-wide' wide option. I think we would definitely want to update the Akka gRPC code generation to take into account package/file-level customizations, but this feature is already useful without it.

sbt-plugin/src/sbt-test/gen-scala-server/09-scalapb-customizations/test
```diff
@@ -0,0 +1,5 @@
+> protocGenerate
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/HelloRequest.scala
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/HelloworldProto.scala
+$ exists target/scala-2.12/akka-grpc/main/example/myapp/helloworld/grpc/GreeterService.scala
+> compile
```
bjaglin, line 5: just force-pushed as I had forgotten this...

ignasi35 (approved):
Nice that we're getting this. 


(discussion for a separate PR/issue): I wonder if should eventually deprecate and remove the use of `flat_package` out of the box. We would then fallback to scalapb's behavior and transparently use the options available there.

bjaglin:
>  I wonder if should eventually deprecate and remove the use of `flat_package` out of the box. We would then fallback to scalapb's behavior and transparently use the options available there.

I am not sure it will be required. ScalaPB has promising upcoming changes that will make consumption of JAR-packaged Scala stubs compiled with different options than the ones used for the local project seamless.

See https://github.com/scalapb/scalapb-validate/issues/70 & https://github.com/scalapb/ScalaPB/pull/873 for related discussions.

ignasi35:
The motivation of my comment was to eventually remove from akka-gRPC any access to scalaPB settings since those scalaPB settings are already available via options in the `.proto` file.

So users no longer need to specify `akkaGrpcCodeGeneratorSettings := akkaGrpcCodeGeneratorSettings.value.filterNot(_ == "flat_package")` and, instead, use plain scalapb syntax on the `.proto` file.

Again, I'd rather discuss these further improvements on a separate issue/PR since I'm sure there'll be use cases that require more attention.

raboof (approved):
Great! Would love to take into account the `flat_package` option better in the Akka gRPC codegen side but that's for another PR, this is already a great improvement.

Issue #1257, opened 2021-02-05
Leak in ClientState


raboof:
`ClientState` registers a `whenTerminated` callback to close the client when the actor system terminates. However, this `whenTerminated` callback is never removed, meaning it will hold on to the `ClientState`, causing a (small) memory leak each time a client is instantiated.

raboof:
Eclipse MAT showing a nice hint of where the leak was (after reproducing it in a testcase and creating a dump with jmap):

![mat](https://user-images.githubusercontent.com/131856/107056582-c58bd200-67d2-11eb-92c4-665f0b52e6b7.png)

PR #1258, opened 2021-02-05
Fix for (slow) memory leak in ClientState

raboof:
`ClientState` registered a `whenTerminated` callback to close the client when
the actor system terminates. However, this `whenTerminated` callback is never
removed, meaning it will hold on to the `ClientState`, causing a (small)
memory leak each time a client is instantiated.

This PR is intended to fix the problem, but doesn't work due to what
looks like an extension initialization deadlock. That reminds of
https://github.com/akka/akka/pull/28046 but it also seems present when I
switch to master, which is on Akka 2.6

Refs https://github.com/akka/akka-grpc/issues/1257

### DIFF
--- a/runtime/src/main/scala/akka/grpc/scaladsl/Grpc.scala
+++ b/runtime/src/main/scala/akka/grpc/scaladsl/Grpc.scala
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package akka.grpc.scaladsl
+
+import scala.concurrent.Future
+import scala.jdk.CollectionConverters._
+
+import akka.Done
+import akka.actor.{ CoordinatedShutdown, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider }
+import akka.annotation.InternalApi
+import akka.grpc.internal.ClientState
+
+import java.util.concurrent.ConcurrentHashMap
+
+/** INTERNAL API */
+@InternalApi
+private[grpc] final class GrpcImpl(system: ExtendedActorSystem) extends Extension {
+  private val clients = new ConcurrentHashMap[ClientState, Unit]
+
+  CoordinatedShutdown(system).addTask("before-actor-system-terminate", "close-grpc-clients") { () =>
+    implicit val ec = system.dispatcher
+    Future
+      .sequence(
+        clients
+          .keySet()
+          .asScala
+          .map(client =>
+            client.close().recover {
+              case e =>
+                system.log.error(e, s"Failed to gracefully close $client, proceeding with shutdown anyway")
+                Done
+            }))
+      .map(_ => Done)
+  }
+
+  /** INTERNAL API */
+  @InternalApi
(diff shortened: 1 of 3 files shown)


runtime/src/main/scala/akka/grpc/scaladsl/Grpc.scala
```diff
@@ -0,0 +1,41 @@
...
+import java.util.concurrent.ConcurrentHashMap
+
+/** INTERNAL API */
+@InternalApi
+class GrpcImpl(system: ExtendedActorSystem) extends Extension {
+  private val clients = new ConcurrentHashMap[ClientState, Unit]
+
+  CoordinatedShutdown(system).addTask("before-actor-system-terminate", "close-grpc-clients") { () =>
+    implicit val ec = system.dispatcher
+    Future.sequence(clients.keySet().asScala.map(_.close())).map(_ => Done)
```
johanandren, line 20: I think with sequence if one future op fails it won't execute the next so maybe wrap the close with a try just for good measure?

runtime/src/main/scala/akka/grpc/scaladsl/Grpc.scala
```diff
@@ -0,0 +1,41 @@
...
+import akka.Done
+import akka.actor.{ CoordinatedShutdown, ExtendedActorSystem, Extension, ExtensionId, ExtensionIdProvider }
+import akka.annotation.InternalApi
+import akka.grpc.internal.ClientState
+
+import java.util.concurrent.ConcurrentHashMap
+
+/** INTERNAL API */
+@InternalApi
+class GrpcImpl(system: ExtendedActorSystem) extends Extension {
```
johanandren, line 15: If you liked it, make it final

runtime/src/main/scala/akka/grpc/scaladsl/Grpc.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /** INTERNAL API */
+  @InternalApi
+  def deregisterClient(client: ClientState): Unit =
+    clients.remove(client)
+}
+
+/** INTERNAL API */
+@InternalApi
+object Grpc extends ExtensionId[GrpcImpl] with ExtensionIdProvider {
```
johanandren, line 36: can be `private[grpc]` perhaps?

johanandren (approved):
Lookin good, left some nitpicking as a weekend gift.

johanandren (approved):

ignasi35 (approved):
LGTM

ignasi35:
Hi @raboof, the branch for this PR starts on the commit tagged `1.0.2`. Did you want to release a `1.0.3` with this fix?

raboof:
> Hi @raboof, the branch for this PR starts on the commit tagged `1.0.2`. Did you want to release a `1.0.3` with this fix?

No, I just started there because that was the version we encountered the leak with. This way we *could* merge it into a 1.0.x branch, but I don't see a strong reason to have such a branch.

ignasi35:
workflow issue is unrelated to this PR (See #1259 )

PR #1259, opened 2021-02-08
Do a full git repo clone for dynver's pleasure

ignasi35:


raboof (approved):
maybe add a comment as it's not so obvious that '0' means 'infinite' here?

.github/workflows/link-validator.yml
```diff
@@ -14,7 +14,8 @@ jobs:
         with:
           # See https://github.com/actions/checkout/issues/299#issuecomment-677674415
           ref: ${{ github.event.pull_request.head.sha }}
-
+          fetch-depth: 0
```
ignasi35, line 17:
```suggestion
          # fetch everything https://github.com/actions/checkout#fetch-all-history-for-all-tags-and-branches
          fetch-depth: 0
```

raboof (approved):

Issue #1232, opened 2021-01-07
remove timestamp from walkthrough example


raboof:
We recently made it easier to pull in the standard dependencies, and added a timestamp to the 'helloworld' proto to test that.

Unfortunately, that was the .proto we also use for the walkthrough in the documentation (https://doc.akka.io/docs/akka-grpc/snapshot/client/walkthrough.html). I think for the walkthrough we should keep things as simple as possible, so make sure we use a proto without external types.

ignasi35:
Did a first hack in #1260 (which hides the unnecessary complexity from the rendered HTML).

The long-term solution should be:

 - create a new proto file
 - remove the use of external dependencies from `./plugin-tester-scala/src/main/protobuf/helloworld.proto` (loosing the scala version of the "use external dependencies test")
 - ... (?)

PR #1260, opened 2021-02-08
Hide import of external types from walkthrough

ignasi35:
References #1232

Just hiding it from teh rendered HTML (to not have this change snowball into code changes)


raboof (approved):
Hmm, this makes things a bit harder to follow though...

Seems like an improvement so :+1: , though let's keep #1232 open until we have cleaned this up? (I guess we can remove the 1.1.0 milestone with this change though)

ignasi35:
@raboof I've left a comment on the issue so we can revisit this in the future.

PR #1261, opened 2021-02-08
Run release drafter as a GH action

ignasi35:
It looks like release drafter stopped running some time ago (around Nov 2020).

This PR brings it back as a GH actions (instead of an external GH App).


ennru (approved):
LGTM.

PR #1262, opened 2021-02-08
Fix syntax on release-drafter.yml

ignasi35:
The instructions at https://github.com/release-drafter/release-drafter#usage are wrong and the workflow doesn't run.

I think this is correct syntax, now.


raboof (approved):
let's try

ennru (approved):
Love it!

ignasi35:
I'm not waiting on travis and jenkins.

PR #1248, opened 2021-01-22
Advertize PGV & scalapb-validate

bjaglin:

PR #1264, opened 2021-02-15
Fix scalapb-validate setup instructions by bumping sbt-protoc

bjaglin:
Follows https://github.com/akka/akka-grpc/pull/1248 which turned out to give non-working instructions for scala/sbt (see [scripted failure on first commit](https://travis-ci.com/github/akka/akka-grpc/jobs/482973309)). The instructions (untouched) only work with
- https://github.com/thesamet/sbt-protoc/commit/ff5f43aec99a98acde8640094a69b03d1464f200
- https://github.com/scalapb/common-protos/commit/7c9595a9a875d53b949b592d7ece2748162d839d

See https://github.com/scalapb/scalapb-validate/issues/70


bjaglin:
Investigating `[error]   not found: /home/travis/.ivy2/localcom.lightbend.akka.grpc/akka-grpc-codegen_2.12/1.1.0-5-8caec5bf/ivys/ivy.xml` in https://travis-ci.com/github/akka/akka-grpc/jobs/482998771 (notice the lack of path separator in `/localcom...`)

raboof:
OK TO TEST

bjaglin:
> Investigating `[error] not found: /home/travis/.ivy2/localcom.lightbend.akka.grpc/akka-grpc-codegen_2.12/1.1.0-5-8caec5bf/ivys/ivy.xml` in https://travis-ci.com/github/akka/akka-grpc/jobs/482998771 (notice the lack of path separator in `/localcom...`)

The lack of path separator was a false track. The regression was caused by a subtle behavior change exposed by https://github.com/thesamet/sbt-protoc/commit/80824b036f87a1f3641a53b37aab09d2d3542009#diff-a241d0444c16c14576b308aa58a0f4f2ed70a3a6021454cd8a71882764513144R582-R583, causing an eager evaluation of the resolver, which in the case of benchmark-java hadn't been published yet. I have updated the task dependency to reflect that, see commit message.

bjaglin:
cc @gontard

benchmark-java/build.sbt
```diff
@@ -29,7 +29,7 @@ val root = project.in(file("."))
       "org.scalatest" %% "scalatest" % "3.1.2" % "test",
       "org.scalatestplus" %% "junit-4-12" % "3.1.2.0" % "test"
     ),
-    Compile / PB.generate := ((Compile / PB.generate) dependsOn (
+    PB.artifactResolver := (PB.artifactResolver dependsOn (
```
raboof, line 32: well-found!

sbt-plugin/src/sbt-test/gen-scala-server/10-scalapb-validate/test
```diff
@@ -0,0 +1 @@
+> compile
```
raboof, line 1: is there more we can test beyond 'it compiles'?

raboof (approved):

sbt-plugin/src/sbt-test/gen-scala-server/10-scalapb-validate/test
bjaglin: good point, I'll add a basic `Main`

bjaglin: First commit amended. I wanted to go even further and verify/show that [`validate_at_construction`](https://github.com/scalapb/scalapb-validate/blob/6de254da20b8e82d3334366705067feea6ece6c9/core/src/main/protobuf/scalapb/validate.proto#L26) could be enabled so that clients sending invalid payloads would automatically get an error (maybe with a cutom exception handler) without any specific custom code on the server implementation, but I ran into https://github.com/scalapb/scalapb-validate/issues/70#issuecomment-779699193 so that will be for the next round.

raboof: > maybe with a cutom exception handler to map `scalapb.validate.ValidationFailure` to `INVALID_ARGUMENT`)

that might even be reasonable to make part of the default handler

raboof (approved):

sbt-plugin/src/sbt-test/gen-scala-server/10-scalapb-validate/test
bjaglin: > that might even be reasonable to make part of the default handler

I got the type wrong, it's [`scalapb.validate.ValidationException`](https://github.com/scalapb/scalapb-validate/blob/a39f93814523b7c74d97d6fba20ccfbfa386621d/core/src/main/scala/scalapb/validate/Validator.scala#L100-L101), which extends `com.google.protobuf.InvalidProtocolBufferException`. Would it make sense to map `InvalidProtocolBufferException` to `INVALID_ARGUMENT` ?

raboof: that makes sense to me

raboof (approved):

sbt-plugin/src/sbt-test/gen-scala-server/10-scalapb-validate/test
raboof: that sounds entirely reasonable!

PR #1270, opened 2021-02-22
Relativize Maven plugin files

andreaTP:
Having relative paths for input and output directories, currently, doesn't play well with Maven multi-module system.

This PR improves the way the Maven Plugin handles paths and their resolution.


pom.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
raboof, line 1: Was it necessary to create a parent pom, or is that just to test/show that multi-module projects now work?

I think it is slightly confusing to have the parent pom here, since it's only for the example/test projects and not the rest of akka-grpc, and the projects are intended to be fairly standalone anyway. I think I'd prefer the separate projects - those also still work, I guess?

andreaTP: Yes, it's mostly for testing, I can remove it! 👍

andreaTP: removed

andreaTP:
Forgot to mention that currently, the plugin doesn't honorate the Maven property `directory` of the parent POM, if it get's tweaked, the plugin will silently not generate the expected code.
As demonstrated here:
https://github.com/GEBIT/lagom-samples/commit/1c0cb3bfc26479d83ff223a4057feed0bce3758e

plugin-tester-scala/pom.xml
```diff
@@ -7,7 +7,7 @@
   <modelVersion>4.0.0</modelVersion>
   <name>Akka gRPC Scala example project</name>
   <groupId>com.lightbend.akka.grpc</groupId>
-  <artifactId>grpc-plugin-tester-java</artifactId>
+  <artifactId>grpc-plugin-tester-scala</artifactId>
```
raboof, line 10: :+1:

raboof (approved):

maven-plugin/src/main/maven/plugin.xml
```diff
@@ -92,8 +92,8 @@
         <generateServer implementation="boolean" default-value="true">${akka-grpc.generate-server}</generateServer>
         <generatorSettings implementation="java.util.List" default-value=""/>
         <extraGenerators implementation="java.util.List" default-value=""/>
-        <protoPaths default-value="src/main/proto,src/main/protobuf">${akka-grpc.protoPaths}</protoPaths>
-        <outputDirectory default-value="target/generated-sources">${akka-grpc.outputDirectory}</outputDirectory>
+        <protoPaths default-value="${project.build.sourceDirectory}/../proto,${project.build.sourceDirectory}/../protobuf">${akka-grpc.protoPaths}</protoPaths>
+        <outputDirectory default-value="${project.build.outputDirectory}/../generated-sources">${akka-grpc.outputDirectory}</outputDirectory>
```
johanandren, line 96: Shouldn't it be `${project.build.directory}/generated-sources` rather?

maven-plugin/src/main/maven/plugin.xml
```diff
@@ -92,8 +92,8 @@
         <generateServer implementation="boolean" default-value="true">${akka-grpc.generate-server}</generateServer>
         <generatorSettings implementation="java.util.List" default-value=""/>
         <extraGenerators implementation="java.util.List" default-value=""/>
-        <protoPaths default-value="src/main/proto,src/main/protobuf">${akka-grpc.protoPaths}</protoPaths>
-        <outputDirectory default-value="target/generated-sources">${akka-grpc.outputDirectory}</outputDirectory>
+        <protoPaths default-value="${project.build.sourceDirectory}/../proto,${project.build.sourceDirectory}/../protobuf">${akka-grpc.protoPaths}</protoPaths>
```
johanandren, line 95: I may be wrong but looking at a few other plugins they seem to rather do what would be `${basedir}/src/main/proto`, would that also break with multi-modules?

johanandren (approved):
LGTM, but I wonder if those backward-relative-paths are really needed.

maven-plugin/src/main/maven/plugin.xml
andreaTP: yes, is better, done thanks 👍

maven-plugin/src/main/maven/plugin.xml
andreaTP: you are right, aligned the definition to follow the reference pom:
https://maven.apache.org/pom.html

johanandren (approved):

andreaTP:
slightly changed the approach for `protoc`:
https://github.com/akka/akka-grpc/pull/1270/commits/e4cb17c5bac1dc95da1e906549f79203f73ecc19
it was buggy for non-multimodule

raboof (approved):

PR #1277, opened 2021-02-25
Cleanup Akka HTTP dependencies and disable auto-bumping

ignasi35:


raboof (approved):

bjaglin:
I landed here after upgrading my service using Play 2.8.7 (with an evicted Akka HTTP 10.1.13) from Akka gRPC 1.1.1 to 2.0.0, as Akka HTTP (10.2.3) started to complain about an old `akka-http2-support` (10.1.13) at the startup check. I wonder if it is not better to leave `akka-http2-support` as a dependency to avoid that?

raboof:
@bjaglin hmm, yes, I think that makes sense.

PR #1281, opened 2021-02-28
Reduce number of bytes array copies to ByteString.

lomigmegard:
Non-functional change. Use `ByteString.fromArrayUnsafe` where safe to save one bytes array copy.

References #1180


raboof (approved):
Indeed since these arrays are immutable, `fromArrayUnsafe` is OK in these cases.

raboof:
OK TO TEST

johanandren (approved):

Issue #1266, opened 2021-02-17
easier grpc server start


patriknw:
What can we do to reduce the amount of code and make it easier to start a gRPC server for a single Service?

This is how the example in the platform guide looks like now: https://github.com/akka/akka-platform-guide/blob/main/docs-source/docs/modules/microservices-tutorial/examples/shopping-order-service-java/src/main/java/shopping/order/ShoppingOrderServer.java

One idea is to generate the handler with service reflection enabled:
```
Function<HttpRequest, CompletionStage<HttpResponse>> service =
        ServiceHandler.concatOrNotFound(
            ShoppingOrderServiceHandlerFactory.create(grpcService, system),
            // ServerReflection enabled to support grpcurl without import-path and proto parameters
            ServerReflection.create(
                Collections.singletonList(ShoppingOrderService.description), system));
```

Could be a new method in `ShoppingOrderServiceHandlerFactory.createWithServerReflection`.

Taking it further we could have a `newServer` directly in ShoppingOrderService, and then ShoppingOrderServiceHandlerFactory wouldn't have to be involved. It should have customization via builder pattern like the http ServerBuilder.

Another variant would be to make http ServerBuilder more powerful/convenient and add coordinated shutdown and logging directly there.

Other ideas?

ignasi35:
Related to https://github.com/akka/akka-http/issues/3772

ignasi35:
@patriknw I've numbered the ideas in your original description to mention them more easily.


------

4. reading certificates and secrets and [building the HttpsContext](https://doc.akka.io/docs/akka-http/current/server-side/server-https-support.html#using-https) (not in the tutorial)

PR #1284, opened 2021-03-05
Convenience builder to concat implementation and reflection (#1266)

ignasi35:
References #1266

### DIFF
--- a/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
@@ -92,7 +92,7 @@ public class @{serviceName}HandlerFactory {
       return partial(implementation, prefix, SystemMaterializer.get(system).materializer(), eHandler, system);
     }
 
-@if(serviceName == "ServerReflection") {} else {
+@if(serviceName != "ServerReflection") {
 
     /**
      * Creates a `HttpRequest` to `HttpResponse` handler that can be used in for example `Http().bindAndHandleAsync`
--- a/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
@@ -92,6 +92,27 @@ public class @{serviceName}HandlerFactory {
       return partial(implementation, prefix, SystemMaterializer.get(system).materializer(), eHandler, system);
     }
 
+@if(serviceName == "ServerReflection") {} else {
+
+    /**
+     * Creates a `HttpRequest` to `HttpResponse` handler that can be used in for example `Http().bindAndHandleAsync`
+     * for the generated partial function handler. The generated handler falls back to a reflection handlers for
+     * `@{service.name}` and ends with `StatusCodes.NotFound` if the request is not matching.
+     *
+     * Use {@@link akka.grpc.javadsl.ServiceHandler#concatOrNotFound} with {@@link @{service.name}HandlerFactory#partial} when combining
+     * several services.
+     */
+    @@SuppressWarnings("unchecked")
+    public static Function<akka.http.javadsl.model.HttpRequest, CompletionStage<akka.http.javadsl.model.HttpResponse>> createWithServerReflection(@serviceName implementation, ClassicActorSystemProvider system) {
+      return akka.grpc.javadsl.ServiceHandler.concatOrNotFound(
+          @{serviceName}HandlerFactory.create(implementation, system),
+          akka.grpc.javadsl.ServerReflection.create(
+            java.util.Collections.singletonList(@{service.name}.description),
+            system));
+    }
+
+}
+
     /**
      * Creates a `HttpRequest` to `HttpResponse` handler that can be used in for example
      * `Http.get(system).bindAndHandleAsync`. It ends with `StatusCodes.NotFound` if the request is not matching.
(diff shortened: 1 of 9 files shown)


codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -92,6 +92,27 @@ public class @{serviceName}HandlerFactory {
       return partial(implementation, prefix, SystemMaterializer.get(system).materializer(), eHandler, system);
     }
 
+@if(serviceName == "ServerReflection") {} else {
```
ignasi35: Don't add the builder if the service we're processing is `ServerReflection`. #turtles

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -78,6 +78,22 @@ object @{serviceName}Handler {
...
+     * `@{service.name}` and ends with `StatusCodes.NotFound` if the request is not matching.
+     *
+     * Use `akka.grpc.scaladsl.ServiceHandler.concatOrNotFound` with `@{service.name}Handler.partial` when combining
+     * several services.
+     */
+    def withServerReflection(implementation: @serviceName)(implicit system: ClassicActorSystemProvider): model.HttpRequest => scala.concurrent.Future[model.HttpResponse] =
+        akka.grpc.scaladsl.ServiceHandler.concatOrNotFound(
+          @{serviceName}Handler.partial(implementation),
+          akka.grpc.scaladsl.ServerReflection.partial(List(@{service.name})))
+}
```
ignasi35, line 95: I'm using FQCN instead of imports to avoid adding unused imports in the case of `ServiceReflection`.

ignasi35 (commented):
comments for reviewers.

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
raboof: Does it hurt to have it available for reflection itself as well?

raboof (commented):
Neat! LGTM, but maybe also update examples/plugintesters or at least one of the scripted tests? And add a pointer to the documentation?

patriknw (approved):
LGTM

ignasi35:
(not to self)

TODO:
- [ ] test
- [ ] docs

codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
ignasi35: It doesn't. It would just be confusing. 😅

sbt-plugin/src/sbt-test/gen-java/02-server-reflection/src/main/java/example/myapp/helloworld/Main.java
```diff
@@ -36,19 +38,31 @@ public static void main(String[] args) throws Exception {
...
         return Http.get(sys)
-	  .newServerAt("127.0.0.1", 8080)
-	  .bind(
-            ServiceHandler.concatOrNotFound(
-                GreeterServiceHandlerFactory.create(impl, sys),
-                ServerReflection.create(Arrays.asList(GreeterService.description), sys)
-            )
-	  );
+          .newServerAt("127.0.0.1", 8080)
+          .bind(GreeterServiceHandlerFactory.createWithServerReflection(impl, sys));
```
ignasi35, line 65: In order to not lose the example on creating the `ServerReflection` given a list of descriptors, I added a new example (lines 41:57) with a more advanced combination of services: (1) a service build with an overload `create` method with (2) a new EchoService and (3) the ServerReflection for those two combined.

raboof (approved):
I like how this makes server reflection more discoverable.

I'm a bit concerned we're introducing many ways to do the same thing. Next thing we will be adding `partialWithServerReflection`, and then all of those but with different parameter variations, and then again with the Directive DSL...

I guess the 'one service with reflection' use case is so common that it makes sense to have a special case for it though.

ignasi35:
> Next thing we will be adding `partialWithServerReflection`, and then all of those but with different parameter variations, and then again with the Directive DSL...

I am 100% against adding a humongous list of factory methods too. This is why I kept referring to the new method as a `convenience` method and also the reason why I extended the docs with a new example of a more elaborate use case (multiple user services _and_ reflection). I would not add more than what we added here. :-)

docs/src/main/paradox/server/reflection.md
```diff
@@ -27,17 +27,31 @@ can be used to query this information
 
 ## Providing
 
-When providing gRPC Server Reflection, you will be serving multiple services
-(at least your own service and the reflection service) as described
-@ref[here](walkthrough.md) { #Serving_multiple_services }. The reflection service
-can be generated via `ServerReflection`:
+The generated service handler includes a convenience method to create an Akka HTTP 
+handler with your service with a `ServerReflection` handler fallback:
```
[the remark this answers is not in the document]
ignasi35: Good point, it doesn't make sense to present this feature in terms of how it's implemented.

At the same time, since this is a convenience method (_ehem_ a `Utils` method _/ehem_), it's not so bad to expose how it is implemented. 

I'll continue thinking of other options anyway. :-)

docs/src/main/paradox/server/reflection.md
```diff
@@ -23,21 +23,33 @@ this metadata to implement things like completion and sending arbitrary
...
+can be used to query this information.
 
 ## Providing
 
-When providing gRPC Server Reflection, you will be serving multiple services
-(at least your own service and the reflection service) as described
-@ref[here](walkthrough.md) { #Serving_multiple_services }. The reflection service
-can be generated via `ServerReflection`:
+The generated service handler includes a convenience method to create an Akka HTTP 
+handler with your service together with Server Reflection:
```
ignasi35, line 31:
```suggestion
handler for the service together with server reflection: 
```

ignasi35: I think I like `handler for the service together with server reflection: ` better.

raboof: :+1: LGTM, merge when you're happy!

Issue #1285, opened 2021-03-06
GoogleProtobufSerializer should not use reflexion


lomigmegard:
### Short description

The `GoogleProtobufSerializer` wrapper currently takes a `Class[_ <: com.google.protobuf.Message]` as argument. It then uses reflection to call the `parseFrom()` static method. This is not a good practice, and Protobuf generated messages have a `Parser` class for this use-case.

### Details

The constructor is used in generated Java sources, but there is a path for binary compatibility. I suggest changing the main constructor to take a `com.google.protobuf.Parser` instead, with a auxiliary constructor to keep backward compatibility.

The rationals for this change are good practices, future improvements that could root from having a `Parser` on-hand, and marginal performance gain.

We can discuss the impacts in the coming PR.

viktorklang:
Sounds like a good idea, @lomigmegard!

Issue #1163, opened 2020-11-04
ScalapbProtobufSerializer use the slow api for deserialize


carbonshow:
## Environment

- **scala version**
   2.13.3
- **akka grpc library**
   akka-grpc-runtime:1.0.2
- **class**
   `akka.grpc.scaladsl.ScalapbProtobufSerializer`

## Short Description

The implementation of method `deserialize` of `ScalapbProtobufSerializer` may call scalapb's slower api `parseFrom(input: InputStream)`:
``` scala
class ScalapbProtobufSerializer[T <: GeneratedMessage](companion: GeneratedMessageCompanion[T])
    extends ProtobufSerializer[T] {
  override def serialize(t: T) = ByteString(companion.toByteArray(t))
  override def deserialize(bytes: ByteString): T = companion.parseFrom(bytes.iterator.asInputStream)
}
``` 

According to real test, `companion.parseFrom(bytes.iterator.asInputStream)` is much slower than `parseFrom(bytes.toArray)` and `parseFrom(CodedInputStream.newInstance(bytes.asByteBuffer))` .

The reason is not related with `akka.util.ByteString`. In fact, scalapb `parseFrom(InputStream)` is just slow.
The benchmark is listed as follows:

### Schema Definition
``` Protobuf
message EchoRequest {
  string message = 1;
}
```

### JMH Test Code

``` scala
class AkkaByteStringParse {

  val MsgLength: Int = 128 // changed in different cases 
  val message: EchoRequest = EchoRequest(util.Random.alphanumeric.take(MsgLength).mkString(""))
  val bytes: ByteString = ByteString(message.toByteArray)

  @Benchmark
  def toArray: Array[Byte] = bytes.toArray

  @Benchmark
  def asInputStream: InputStream = bytes.iterator.asInputStream

  @Benchmark
  def parseFromArray: EchoRequest = EchoRequest.parseFrom(toArray)

  @Benchmark
  def parseFromInputStream: EchoRequest = EchoRequest.parseFrom(asInputStream)

  @Benchmark
  def parseFromByteBuffer: EchoRequest = EchoRequest.parseFrom(CodedInputStream.newInstance(bytes.asByteBuffer))
}
```

### Test Result

1 op means running 50000 times, **bytes** has the type `akka.util.ByteString` :

#### message length = 128 Bytes
[...]

PR #1286, opened 2021-03-06
Remove reflection usage in GoogleProtobufSerializer (#1285)

lomigmegard:
References #1285

Open question is how best handle binary compatibility, is the given auxiliary constructor enough? Should it be marked as deprecated?

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
@@ -12,13 +12,12 @@ import com.google.protobuf.Parser
 @ApiMayChange
 class GoogleProtobufSerializer[T <: com.google.protobuf.Message](parser: Parser[T]) extends ProtobufSerializer[T] {
 
-  // For binary compatibility in generated sources, can be dropped in version 2.x
+  @deprecated("Kept for binary compatibility, use the main constructor instead", since = "akka-grpc 1.1.2")
   def this(clazz: Class[T]) =
     this(clazz.getMethod("parser").invoke(clazz).asInstanceOf[Parser[T]])
 
   override def serialize(t: T): ByteString =
     ByteString.fromArrayUnsafe(t.toByteArray)
-  override def deserialize(bytes: ByteString): T = {
+  override def deserialize(bytes: ByteString): T =
     parser.parseFrom(bytes.toArray)
-  }
 }
--- a/runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
@@ -7,13 +7,18 @@ package akka.grpc.javadsl
 import akka.annotation.ApiMayChange
 import akka.grpc.ProtobufSerializer
 import akka.util.ByteString
+import com.google.protobuf.Parser
 
 @ApiMayChange
-class GoogleProtobufSerializer[T <: com.google.protobuf.Message](clazz: Class[T]) extends ProtobufSerializer[T] {
+class GoogleProtobufSerializer[T <: com.google.protobuf.Message](parser: Parser[T]) extends ProtobufSerializer[T] {
+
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
```diff
@@ -7,13 +7,18 @@ package akka.grpc.javadsl
...
+  // For binary compatibility in generated sources, can be dropped in version 2.x
+  def this(clazz: Class[T]) =
+    this(clazz.getMethod("parser").invoke(clazz).asInstanceOf[Parser[T]])
+
   override def serialize(t: T): ByteString =
     ByteString.fromArrayUnsafe(t.toByteArray)
   override def deserialize(bytes: ByteString): T = {
-    val parser = clazz.getMethod("parseFrom", classOf[Array[Byte]])
-    parser.invoke(clazz, bytes.toArray).asInstanceOf[T]
+    parser.parseFrom(bytes.toArray)
```
viktorklang, line 22: Wouldn't this avoid a copy?

```suggestion
    parser.parseFrom(bytes.iterator.asInputStream)
```

viktorklang: I guess one could also do:

```suggestion
    parser.parseFrom(bytes.compact.asByteBuffer)
```

But that might entail copying for fragmented ByteStrings.

lomigmegard: This PR is just aiming at removing the reflection, and then allow those changes since we will have the parser available :)

But yes both options might be more performant, it's really hard to know the impact without proper benchmarking as a single copy to an array can be faster than many small ones using a stream. My gut-feeling is that for small messages, copying  once into an array is faster, but it depends a lot on the implementation (reader-side, e.g. with buffer sizes), and on the fragmentation. We should probably align the implementations between Java/Scala APIs, see `ScalapbProtobufSerializer` that uses the `bytes.asByteBuffer` approach. Also related to #1163 for a benchmark on that point.

I might do such benchmark if I find the time, I think the hard part is to find the correct dimensions and sample data to show the tradeoffs, as shown in #1180 by @raboof and @ignasi35.

viktorklang: I'll defer to @raboof and @ignasi35  :)

raboof:
OK TO TEST

runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
```diff
@@ -7,13 +7,18 @@ package akka.grpc.javadsl
 import akka.annotation.ApiMayChange
 import akka.grpc.ProtobufSerializer
 import akka.util.ByteString
+import com.google.protobuf.Parser
 
 @ApiMayChange
-class GoogleProtobufSerializer[T <: com.google.protobuf.Message](clazz: Class[T]) extends ProtobufSerializer[T] {
+class GoogleProtobufSerializer[T <: com.google.protobuf.Message](parser: Parser[T]) extends ProtobufSerializer[T] {
+
+  // For binary compatibility in generated sources, can be dropped in version 2.x
```
raboof, line 15: Maybe mark it `@deprecated` so it's easier to find when we can break binary compatibility (which indeed is in 2.x, but also 1.3.x - https://doc.akka.io/docs/akka-grpc/current/binary-compatibility.html#deprecations )

raboof (approved):
The motivation for this PR is convincing.

The only reservation I have is that this ties us rather strongly to the `com.google.protobuf.Parser` API, but that is probably fine.

Good to stick to `toArray` for now (and perhaps consider changing that in another PR after benchmarking, indeed also considering the previous findings in that area)

runtime/src/main/scala/akka/grpc/javadsl/GoogleProtobufSerializer.scala
lomigmegard: Added

raboof:
Great, thanks!

PR #1290, opened 2021-03-10
[wip] Remove obsolete plugin settings

ignasi35:
play-related settings are noop. They should be, at least, removed from docs.

### DIFF
--- a/docs/src/main/paradox/buildtools/gradle.md
+++ b/docs/src/main/paradox/buildtools/gradle.md
@@ -26,9 +26,6 @@ Names and default values are provided.
     akkaGrpc {
         generateClient = true
         generateServer = true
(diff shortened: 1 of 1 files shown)


docs/src/main/paradox/buildtools/gradle.md
```diff
@@ -26,9 +26,6 @@ Names and default values are provided.
     akkaGrpc {
         generateClient = true
         generateServer = true
-        generatePlay = false
-        usePlayActions = false
-        serverPowerApis = false
```
ignasi35, line 31: `serverPowerApis` is still valid. I should have not removed it.

ignasi35:
I'll close this for the time being.

PR #1294, opened 2021-03-14
let scala-steward upgrade sbt-protoc

bjaglin:
sbt-protoc has reached 1.x, so there is no reason to handle bumps manually

History
- https://github.com/akka/akka-grpc/pull/594/files
- https://github.com/akka/akka-grpc/pull/702#discussion_r341852976


raboof (approved):
The original reason to have this excluded is because we wanted to keep the sbt-protoc version in sync with the one that scalapb is using - but since we haven't been doing that anyway, removing the exclude seems like an improvement to me :+1:

raboof:
OK TO TEST

johanandren (approved):

PR #1317, opened 2021-04-07
Use valid headers in unit tests

lomigmegard:
Remove annoying invalid header warnings in unit test:

```
[WARN] [04/07/2021 10:56:00.405] [GreeterServer-akka.actor.default-dispatcher-9] [akka.actor.ActorSystemImpl(GreeterServer)] Illegal header: Illegal 'authorization' header: Invalid input '<', expected basic-credential-def, oauth2-bearer-token or generic-credentials (line 1, column 1): <some auth token>
```

### DIFF
--- a/plugin-tester-java/src/test/scala/example/myapp/helloworld/JGreeterServiceSpec.scala
+++ b/plugin-tester-java/src/test/scala/example/myapp/helloworld/JGreeterServiceSpec.scala
@@ -67,7 +67,7 @@ class JGreeterServiceSpec extends Matchers with AnyWordSpecLike with BeforeAndAf
         s"use metadata in replying to single request ($ix)" in {
           val reply = clients.last
             .sayHello()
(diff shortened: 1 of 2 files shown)


raboof (approved):

raboof:
OK TO TEST

johanandren (approved):

lomigmegard:
@johanandren thanks for merging :) The `1.1.0` milestone is however probably not correct, as versions `1.1.0` and `1.1.1` are already released (and this change is probably not note-worthy).

raboof:
> @johanandren thanks for merging :) The `1.1.0` milestone is however probably not correct, as versions `1.1.0` and `1.1.1` are already released (and this change is probably not note-worthy).

Ah, thanks for the ping - we used the `1.1.0` milestone to mark issues that we didn't want to go into `1.0.x`, but now that `1.1.0` has been released we should close it.

PR #1318, opened 2021-04-07
Use provided implementation for stream parsing (#1180)

lomigmegard:
Leverage the GoogleProtobuf and ScalaPB implementations to parse directly from `InputStream`.

---

As the akka-grpc library is only a middleware between libraries, it should not make assumptions on the data that will be deserialized. Therefore, I feel like the implementation of `io.grpc.MethodDescriptor.Marshaller.parse(InputStream)` should pass the stream further down without reading it fully in memory. The upsides are a better abstraction over the implementation, actual streaming for large data, and some less copying of byte buffers around (as explained by @ignasi35 in #1180). The potential downside could be a small overhead with average-sized data that don't fit in the default buffer size (4K), but I don't think it's the role of akka-grpc to make heuristics and try to optimize for some patterns. 

References #1180


raboof:
> As the akka-grpc library is only a middleware between libraries, it should not make assumptions on the data that will be deserialized. Therefore, I feel like the implementation of `io.grpc.MethodDescriptor.Marshaller.parse(InputStream)` should pass the stream further down without reading it fully in memory. The upsides are a better abstraction over the implementation, actual streaming for large data, and some less copying of byte buffers around (as explained by @ignasi35 in #1180). 

Thanks for having a look at this! As hinted at by the "we can't avoid it for the moment" comment, I completely agree the current implementation is not great, but we felt it was unavoidable at the time - your change looks convincing, however, so perhaps indeed that limitation doesn't exist (anymore) - which would be great!

> The potential downside could be a small overhead with average-sized data that don't fit in the default buffer size (4K), but I don't think it's the role of akka-grpc to make heuristics and try to optimize for some patterns.

While I agree with the sentiment, where there are different implementations possible I think we should take care to select the 'right' one, and sometimes selecting the 'right' one will force us to make a choice... but of course if we can avoid it we should.

I'm rather swamped with other work right now, so unfortunately it might be a while before I can have a proper look at the implications of this PR, but it looks very promising and I much appreciate it!

raboof:
(the travis failure seems to be a MiMa warning that can be ignored because it's about a method being added to a class that isn't meant for user extension)

jrudolph (approved):
LGTM

I traced the down method calls inside the protobuf implementation and there are indeed multiple implementations of the internal `CodedInputStream`, one optimized for byte array based access, and a streaming one as well.

jrudolph:
> The potential downside could be a small overhead with average-sized data that don't fit in the default buffer size (4K), but I don't think it's the role of akka-grpc to make heuristics and try to optimize for some patterns.

Which buffer do you mean here?

jrudolph:
> > The potential downside could be a small overhead with average-sized data that don't fit in the default buffer size (4K), but I don't think it's the role of akka-grpc to make heuristics and try to optimize for some patterns.
> 
> Which buffer do you mean here?

Probably `CodedInputStream.DEFAULT_BUFFER_SIZE`?

jrudolph:
Also looking through the code that creates the InputStream, it seems, there might be another optimization possible later: the InputStream passed to `parse` is probably also already backed by a buffer. In that case, we might even skip another step and provide a `CodedInputStream` directly from the underlying buffer.

jrudolph:
> @jrudolph When I ran the tests in akka-grpc to check which InputStreams had KnownLength then it was only a subset of them. The Netty parts pass a non-knownlength InputStream for instance. (at least in some cases)

Interesting, from reading `io.grpc.internal.MessageDeframer`, I would imagine that this can only happen if the message was (de)compressed before. In that case, we probably don't have to worry about the extra 4k buffer anyway.

jrudolph (approved):
LGTM, it surely makes the code simpler and probably it's also faster because it avoids the extra buffer copy.

runtime/src/main/scala/akka/grpc/internal/ByteStringUtils.scala
```diff
@@ -0,0 +1,44 @@
...
+import akka.annotation.InternalApi
+import akka.util.ByteString
+import io.grpc.KnownLength
+
+import java.io.{ ByteArrayOutputStream, InputStream }
+
+@InternalApi
+private[grpc] object ByteStringUtils {
+  def fromInputStream(stream: InputStream): ByteString = {
+    val buffer =
```
jrudolph, line 12: I moved the old code to this Utils object so we can provide a default implementation in ProtobufSerializer for compat reasons.

lomigmegard:
@jrudolph Thanks for the review and compat changes. I saw that you worked on yet another optimization around this in https://github.com/akka/akka-grpc/pull/1318 that might be even better. Let me know if there is more I can do, cheers.

jrudolph:
Thanks again, @lomigmegard!

lomigmegard:
@jrudolph I think https://github.com/akka/akka-grpc/issues/1180 can be closed now, as the core issue was removed with this PR.

PR #1333, opened 2021-05-03
Fix client streaming cancellation before trailers read

pvlugter:
Noticed in the Akka Serverless tests — client-streaming calls would sometimes have empty metadata. For client-streaming requests, the `Sink.head` cancels the client-side stream, which can complete the trailer promise with a cancelled status and empty metadata before the trailer frame has been read. Add a cancellation barrier, as used elsewhere, and a simple test that confirms the trailers can be received after the reply.

Note that a not-OK status (in the trailer) after a reply doesn't affect the response future and is effectively ignored by the client. Not sure what the intended behaviour is for the reply and fail status case. The Akka HTTP client does include the status in the trailer metadata, while the Netty client doesn't — it could be useful to at least add the status to the trailers.

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
@@ -11,6 +11,7 @@ import scala.concurrent.duration._
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.grpc.GrpcClientSettings
+import akka.grpc.GrpcProtocol.TrailerFrame
 import akka.grpc.GrpcResponseMetadata
 import akka.grpc.internal.GrpcEntityHelpers
 import akka.grpc.internal.GrpcProtocolNative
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/internal/RequestBuilderImpl.scala
```diff
@@ -128,7 +128,11 @@ final class ScalaClientStreamingRequestBuilder[I, O](
     val src =
       channel.invokeWithMetadata(source, headers, descriptor, false, callOptionsWithDeadline())
     val (metadataFuture: Future[GrpcResponseMetadata], resultFuture: Future[O]) =
-      src.toMat(Sink.head)(Keep.both).run()
+      src
+        // Continue reading to get the trailing headers
+        .via(new CancellationBarrierGraphStage)
```
viktorklang, line 133: Very clean fix, great work @pvlugter!

raboof:
(failure is a Scala 2.12.11 vs 2.12.13 thing in the Gradle tests - I think we can ignore/remove those since they have been migrated to GitHub Actions and seem to work there :+1:

interop-tests/src/test/scala/akka/grpc/scaladsl/PowerApiSpec.scala
```diff
@@ -140,6 +141,37 @@ abstract class PowerApiSpec(backend: String)
...
+      response.value.message shouldBe "Hello there!"
+      response.headers.getText("foo") shouldBe Some("bar")
+
+      // only complete trailer after response received, to test reading of trailing headers
+      trailer.success(GrpcEntityHelpers.trailer(Status.OK, new HeaderMetadataImpl(List(RawHeader("baz", "qux")))))
+
+      response.trailers.futureValue.getText("baz") shouldBe Some("qux")
+
+      metadataServer.terminate(3.seconds)
+    }
```
raboof, line 174: Neatly tested as well :+1:

raboof (approved):

Issue #1323, opened 2021-04-14
Migrate from Sonatype to bintray


ignasi35:
### Short description

Bintray is sunsetting on May 1st 2021

See for example

https://github.com/akka/akka-projection/pull/530

https://github.com/akka/alpakka-kafka/pull/1353

PR #1336, opened 2021-05-04
Release directly to sonatype

raboof:
rather than going via bintray

References #1323


.travis.yml
```diff
@@ -159,15 +159,10 @@ jobs:
 
     - stage: whitesource
       script: git branch -f "$TRAVIS_BRANCH" && git checkout "$TRAVIS_BRANCH" && cp .jvmopts-travis .jvmopts && sbt whitesourceCheckPolicies whitesourceUpdate
-    
+
     - stage: publish
-      script: cp .jvmopts-travis .jvmopts && sbt ++2.12.11 publish && git status && sbt ++2.13.3 akka-grpc-runtime/publish ++2.13.3! akka-grpc-codegen/publish
-      name: sbt publish
-    - script: openssl aes-256-cbc -K $encrypted_9c9c33071881_key -iv $encrypted_9c9c33071881_iv -in gradle.properties.enc -out ./gradle-plugin/gradle.properties -d && cd gradle-plugin && ./gradlew publishPlugins
+      script: openssl aes-256-cbc -K $encrypted_9c9c33071881_key -iv $encrypted_9c9c33071881_iv -in gradle.properties.enc -out ./gradle-plugin/gradle.properties -d && cd gradle-plugin && ./gradlew publishPlugins
```
ignasi35, line 164: This looks like a left-over. 😅

ignasi35 (changes requested):
I think the gradle plugin will be published twice.

.travis.yml
raboof: no, I didn't convert the gradle publishing yet (because it needs that travis-encrypted `gradle.properties` file)

ignasi35: I see, this is still DRAFT. I'll hold.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+      - name: Set up JDK 11
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.11.0-9
```
johanandren, line 23: We don't support Java 8 or it is anyways safe to build release using 11?

gradle-plugin/build.gradle
```diff
@@ -10,7 +10,7 @@ group = "com.lightbend.akka.grpc"
 // https://github.com/palantir/gradle-git-version/issues/97
 
 def tag = "git describe --tags".execute().text.substring(1).split("-g")[0].replace("\n", "")
-def finalVersion = (tag == versionDetails().lastTag.substring(1)) ? tag : tag.reverse().reverse() + "-" + versionDetails().gitHash.substring(0, 8)
+def finalVersion = (tag == versionDetails().lastTag.substring(1)) ? tag : tag.reverse().reverse() + "-" + versionDetails().gitHash.substring(0, 8) + "-SNAPSHOT"
```
johanandren, line 13: Hmmm, always uses a "-SNAPSHOT" suffix in `finalVersion`?

johanandren (approved):
LGTM with questions

gradle-plugin/build.gradle
raboof: If `tag == versionDetails().lastTag.substring(1)` it uses the tag as-is, otherwise it adds the git hash and `-SNAPSHOT` suffix. (not sure what the `.reverse().reverse()` is about btw...)

johanandren: Ah, didn't match up the parenthesises correctly 👍

.github/workflows/publish.yml
raboof: ow, good catch, looks like we don't specify a target in the build so we should definitely go with jdk8

raboof:
https://oss.sonatype.org/content/repositories/snapshots/com/lightbend/akka/grpc/akka-grpc-runtime_2.13/1.1.1-57-18bcff93-SNAPSHOT

looking promising, but looks like there's something wrong with the SCP credentials for updating the documentation.

Issue #1349, opened 2021-05-18
Marshalling performance


jtjeferreira:
### Short description

When investigating akka-grpc performance in https://discuss.lightbend.com/t/akka-grpc-performance-in-benchmarks/8236/, I noticed that marshalling performance is affected by the use of akka-stream concat

### Details

Here is the flamegraph

![image](https://user-images.githubusercontent.com/943051/118638142-6e809a80-b7ce-11eb-8740-684c58dd3ce8.png)

jrudolph:
Yes, that's one of the bigger costs of the streaming infrastructure. In that particular case, `withDetachedInputs` in the constructor of `concat` is the main problem. IMO it's not the right thing to add it automatically everywhere but I guess we cannot fix that anymore for compatibility reasons.

I'm not convinced though that adding these kinds of micro-optimizations are the way to go. In https://github.com/akka/akka-http/issues/3815 I outlined ideas about how to avoid creating streams when possible which would also avoid this streams usage altogether.

I used this as a temporary workaround:

```diff
--- a/runtime/src/main/scala/akka/grpc/internal/GrpcEntityHelpers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/GrpcEntityHelpers.scala
@@ -13,9 +13,12 @@ import akka.grpc.scaladsl.{ headers, BytesEntry, Metadata, MetadataEntry, String
 import akka.http.scaladsl.model.HttpEntity.ChunkStreamPart
 import akka.http.scaladsl.model.HttpHeader
 import akka.http.scaladsl.model.headers.RawHeader
-import akka.stream.scaladsl.Source
+import akka.stream.{ FlowShape, Graph, SourceShape }
+import akka.stream.scaladsl.{ Concat, GraphDSL, Source }
 import io.grpc.Status
 
+import scala.annotation.unchecked.uncheckedVariance
+
 /** INTERNAL API */
 @InternalApi
 object GrpcEntityHelpers {
@@ -39,10 +42,18 @@ object GrpcEntityHelpers {
   def apply[T](e: T)(implicit m: ProtobufSerializer[T], writer: GrpcProtocolWriter): Source[ChunkStreamPart, NotUsed] =
     chunks(Source.single(e), Source.empty)
 
+  def concatCheap[U, Mat2](that: Graph[SourceShape[U], Mat2]): Graph[FlowShape[U @uncheckedVariance, U], Mat2] =
+    GraphDSL.create(that) { implicit b => r =>
[...]

PR #1350, opened 2021-05-18
Improve marshalling performance

jtjeferreira:
References #1349

Before
```
[info] GrpcMarshallingBenchmark.marshall        thrpt    3  126666.347 ± 15574.678  ops/s
[info] GrpcMarshallingBenchmark.marshallStream  thrpt    3  114767.059 ±  2759.197  ops/s
```

after `concatCheap`

```
[info] GrpcMarshallingBenchmark.marshall        thrpt    3  256018.599 ± 60197.380  ops/s
[info] GrpcMarshallingBenchmark.marshallStream  thrpt    3  221998.871 ± 83522.412  ops/s
```

As @jrudolph said maybe these microoptimizations are not worth it, so feel free to close the PR

### DIFF
--- a/benchmarks/src/main/scala/akka/grpc/HandlerProcessingBenchmark.scala
+++ b/benchmarks/src/main/scala/akka/grpc/HandlerProcessingBenchmark.scala
@@ -30,7 +30,7 @@ class HandlerProcessingBenchmark extends CommonBenchmark {
   val in = Source.repeat(ServerReflectionRequest()).take(10000)
   val request: HttpRequest = {
     implicit val serializer = ServerReflection.Serializers.ServerReflectionRequestSerializer
-    GrpcRequestHelpers(Uri("https://unused.example/" + ServerReflection.name + "/ServerReflectionInfo"), in)
+    GrpcRequestHelpers(Uri("https://unused.example/" + ServerReflection.name + "/ServerReflectionInfo"), Nil, in)
   }
 
   val handler: HttpRequest => Future[HttpResponse] = ServerReflectionHandler(new ServerReflection {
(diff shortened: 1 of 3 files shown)


benchmarks/src/main/scala/akka/grpc/HandlerProcessingBenchmark.scala
```diff
@@ -30,7 +30,7 @@ class HandlerProcessingBenchmark extends CommonBenchmark {
   val in = Source.repeat(ServerReflectionRequest()).take(10000)
   val request: HttpRequest = {
     implicit val serializer = ServerReflection.Serializers.ServerReflectionRequestSerializer
-    GrpcRequestHelpers(Uri("https://unused.example/" + ServerReflection.name + "/ServerReflectionInfo"), in)
+    GrpcRequestHelpers(Uri("https://unused.example/" + ServerReflection.name + "/ServerReflectionInfo"), Nil, in)
```
jtjeferreira, line 33: looks like the `benchmarks` project is not being compiled by travis/github actions

raboof:
OK TO TEST

benchmarks/src/main/scala/akka/grpc/HandlerProcessingBenchmark.scala
raboof: oops! seems like the build the `benchmark-java` subproject but not the `benchmarks`, indeed

raboof (approved):
> maybe these microoptimizations are not worth it, so feel free to close the PR

In general we should be a bit careful not to get too far 'in the woods' on small optimizations as it might detract from more 'big-picture' opportunities, but in this case it seems fairly targeted and not impact maintainability too much, so I'm OK with having this one in.

PR #674, opened 2019-08-23
Store the gradle plugin logfile inside the build directory

hithran:
## Purpose

The location of the logfile is changed from the users temp directory to the gradle build directory.
This in order to prevent an issues when using the plugin on Windows while the project folder is not on `C:\`

## References

References #669

## Changes

- Change the logfile location from the users temp directory to the gradle build directory
- Log file is no longer created when the plugin is configured

## Background Context

As discussed in #669, there is a split on `:` when passing the logfile location, which complicates the approach of passing a Windows path with a drive letter. Writing the logfile to the gradle build directory seems the next best thing when trying to get it all to work.

Issue #669, opened 2019-08-17
Gradle plugin fails on task generateProto when using Windows 10


hithran:
### Versions used 

akka-grpc-gradle-plugin:0.7.1

### Expected Behavior

Have it generate the akka grpc files

### Actual Behavior

Throws a FileNotFoundException

### Relevant logs

```
> Configure project :interface
[task ':interface:assemble', task ':interface:assembleDist', task ':interface:build', task ':interface:buildDependents', task ':interface:buildEnvironment', task ':interface:buildNeeded', task ':interface:check', task ':interface:classes', task ':interface:clean', task ':interface:compileJava', task ':interface:compileScala', task ':interface:compileTestJava', task ':interface:compileTestScala', task ':interface:components', task ':interface:dependencies', task ':interface:dependencyInsight', task ':interface:dependentComponents', task ':interface:distTar', task ':interface:distZip', task ':interface:help', task ':interface:installDist', task ':interface:jar', task ':interface:javadoc', task ':interface:model', task ':interface:processResources', task ':interface:processTestResources', task ':interface:projects', task ':interface:properties', task ':interface:run', task ':interface:scaladoc', task ':interface:startScripts', task ':interface:tasks', task ':interface:test', task ':interface:testClasses']

> Task :interface:extractIncludeProto UP-TO-DATE
> Task :interface:extractProto UP-TO-DATE
> Task :interface:generateProto FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':interface:generateProto'.
> protoc: stdout: . stderr: D:\interface\src\main\proto: warning: directory does not exist.
  D:\interface\app\protobuf: warning: directory does not exist.
  D:\interface\app\proto: warning: directory does not exist.
  Exception in thread "main" java.io.FileNotFoundException: \Users\<User>\AppData\Local\Temp\akka-grpc-gradle6625443006303105175.log (Het systeem kan het opgegeven pad niet vinden)
  	at java.base/java.io.FileOutputStream.open0(Native Method)
  	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:298)
  	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:237)
  	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:187)
  	at java.base/java.io.PrintWriter.<init>(PrintWriter.java:208)
  	at java.base/java.io.PrintWriter.<init>(PrintWriter.java:248)
[...]

PR #1363, opened 2021-05-30
Encode logfile option with URLEncoder

### LINKED ISSUE #786: Log file can not be found for multiproject Gradle projects

jasonxh:
References #786

## Issue
Currently, applying the gradle plugin to a subproject will fail gradle build from root project dir, due to a mismatch in the expectation of where the `logfile` is located. The gradle plugin calculates a relative path to the subproject, while the protoc codegen plugin evaluates it as relative to the current working dir (which is root dir).

## Background
There had been several back-and-forths regarding how to pass the `logfile` option in the past, ranging from one-off character substitution, to relative paths, to absolute paths (see #674 and #787). The root of the problem is that protoc codegen plugin options are passed in as a single string, encoded in a special format like this:

```
--akkaGrpc_out=<key1>=<value1>,<key2>=<value2>:<out_dir>
```

 It's easy to see how several characters play a special role in this encoding, and thus should not appear in keys or values: `[=,:]`.

This problem was originally surfaced as a Windows-only problem, where absolute file paths include `C:\` (see #669). However, it doesn't actually matter if we are passing `logfile` as absolute or relative, since those special characters mentioned above can creep in either way. Granted, no sane person would name their files like that, but it's possible.

## Proposed fix
Encode `logfile` with [URLEncoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLEncoder.html) when composing options, and decode it with [URLDecoder](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/URLDecoder.html) inside the protoc codegen plugin. This is a well-known encoding scheme that allows only `[a-zA-Z0-9.-*_+%]` in the encoded output. Note that none of the special characters mentioned above are allowed here.

I also renamed the option to `logfile_enc` to make it an explicit breaking change, in case the codegen plugin is being used in other context. But I don't feel strongly about this, so feel free to suggest otherwise.

## Testing
Verified to fix #786 on my Mac. Would be great if someone can help test this on a Windows machine.

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/Main.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/Main.scala
@@ -5,6 +5,7 @@
 package akka.grpc.gen
 
 import java.io.ByteArrayOutputStream
+import java.net.URLDecoder
 
 import com.google.protobuf.compiler.PluginProtos.CodeGeneratorRequest
 import akka.grpc.gen.javadsl.{ JavaClientCodeGenerator, JavaInterfaceCodeGenerator, JavaServerCodeGenerator }
@@ -45,7 +46,8 @@ object Main extends App {
   private val extraGenerators: List[String] =
     parameters.getOrElse("extra_generators", "").split(";").toList.filter(_ != "")
 
-  private val logger = parameters.get("logfile").map(new FileLogger(_)).getOrElse(SilencedLogger)
+  private val logger: Logger =
+    parameters.get("logfile_enc").map(URLDecoder.decode(_, "utf-8")).map(new FileLogger(_)).getOrElse(SilencedLogger)
 
   val out = {
     val codeGenerators =
--- a/codegen/src/main/scala/akka/grpc/gen/Main.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/Main.scala
@@ -46,8 +46,14 @@ object Main extends App {
   private val extraGenerators: List[String] =
     parameters.getOrElse("extra_generators", "").split(";").toList.filter(_ != "")
 
+  // Prefer logfile_enc with fallback to logfile
   private val logger: Logger =
-    parameters.get("logfile_enc").map(URLDecoder.decode(_, "utf-8")).map(new FileLogger(_)).getOrElse(SilencedLogger)
+    parameters
+      .get("logfile_enc")
+      .map(URLDecoder.decode(_, "utf-8"))
+      .orElse(parameters.get("logfile"))
+      .map(new FileLogger(_))
+      .getOrElse(SilencedLogger)
 
   val out = {
     val codeGenerators =
--- a/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
+++ b/gradle-plugin/src/main/groovy/akka/grpc/gradle/AkkaGrpcPlugin.groovy
@@ -11,6 +11,7 @@ import org.gradle.internal.component.local.model.DefaultProjectComponentSelector
 import org.gradle.util.GradleVersion
 import org.gradle.util.VersionNumber
 
+import java.net.URLEncoder
 import java.nio.file.Files
 import java.nio.file.Path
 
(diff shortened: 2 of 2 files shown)


lightbend-cla-validator:
Hi @jasonxh,

raboof:
OK TO TEST

raboof:
@hithran @orendain @eshepelyuk thoughts?

eshepelyuk:
> @hithran @orendain @eshepelyuk thoughts?
looks ok to me.

codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -45,7 +46,8 @@ object Main extends App {
   private val extraGenerators: List[String] =
     parameters.getOrElse("extra_generators", "").split(";").toList.filter(_ != "")
 
-  private val logger = parameters.get("logfile").map(new FileLogger(_)).getOrElse(SilencedLogger)
+  private val logger: Logger =
+    parameters.get("logfile_enc").map(URLDecoder.decode(_, "utf-8")).map(new FileLogger(_)).getOrElse(SilencedLogger)
```
raboof, line 50: can we fall back to the `logfile` parameter when `logfile_enc` is not present?

jasonxh: Added fallback logic

raboof (approved):
Nice, thanks!

hithran:
> @hithran @orendain @eshepelyuk thoughts?

@jasonxh Great solution! Thanks!

PR #1365, opened 2021-05-31
Strict marshalling

jrudolph:
On top of https://github.com/akka/akka-http/pull/3822 gives already about 40 - 50 % perf improvement in grpc_bench.

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/interop/AkkaHttpServerProviderScala.scala
+++ b/interop-tests/src/test/scala/akka/grpc/interop/AkkaHttpServerProviderScala.scala
@@ -7,7 +7,8 @@ package akka.grpc.interop
 import akka.NotUsed
 import akka.actor.ActorSystem
 import akka.grpc.internal.{ GrpcEntityHelpers, GrpcProtocolNative, GrpcResponseHelpers, Identity }
-import akka.http.scaladsl.model.{ HttpEntity, HttpHeader }
+import akka.http.scaladsl.model.headers.RawHeader
+import akka.http.scaladsl.model.{ AttributeKeys, HttpEntity, HttpHeader, Trailer }
 import akka.http.scaladsl.server.{ Directive0, Directives, Route }
 import akka.stream.Materializer
 import akka.stream.scaladsl.Source
@@ -95,17 +96,22 @@ object AkkaHttpServerProviderScala extends AkkaHttpServerProvider with Directive
   // TODO move to runtime library or even akka-http
   def mapTrailingResponseHeaders(f: immutable.Seq[HttpHeader] => immutable.Seq[HttpHeader]): Directive0 =
     mapResponse(response =>
-      response.withEntity(response.entity match {
-        case HttpEntity.Chunked(contentType, data) => {
-          HttpEntity.Chunked(
-            contentType,
-            data.map {
-              case chunk: HttpEntity.Chunk => chunk
-              case last: HttpEntity.LastChunk =>
-                HttpEntity.LastChunk(last.extension, f(last.trailer))
-            })
-        }
+      response.entity match {
+        case HttpEntity.Chunked(contentType, data) =>
+          response.withEntity(
+            HttpEntity.Chunked(
+              contentType,
+              data.map {
+                case chunk: HttpEntity.Chunk => chunk
+                case last: HttpEntity.LastChunk =>
+                  HttpEntity.LastChunk(last.extension, f(last.trailer))
+              }))
         case _ =>
-          throw new IllegalArgumentException("Trailing response headers are only supported on Chunked responses")
-      }))
+          val origTrailers = response
+            .attribute(AttributeKeys.trailer)
+            .map(_.headers)
+            .getOrElse(Vector.empty)
+            .map(e => RawHeader(e._1, e._2))
+          response.addAttribute(AttributeKeys.trailer, Trailer(f(origTrailers)))
+      })
 }
--- a/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
@@ -88,6 +88,8 @@ object GrpcProtocol {
       messageEncoding: Codec,
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
+      /** A shortcut to encode a data frame directly into a ByteString */
+      encodeDataToFrameBytes: ByteString => ByteString,
       /** A Flow over a stream of Frame using this frame encoding */
       frameEncoder: Flow[Frame, ChunkStreamPart, NotUsed])
 
--- a/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
@@ -9,7 +9,7 @@ import akka.actor.ActorSystem
 import akka.actor.ClassicActorSystemProvider
 import akka.annotation.InternalApi
 import akka.grpc.{ ProtobufSerializer, Trailers }
-import akka.grpc.GrpcProtocol.{ DataFrame, GrpcProtocolWriter, TrailerFrame }
+import akka.grpc.GrpcProtocol.{ GrpcProtocolWriter, TrailerFrame }
 import akka.grpc.scaladsl.{ headers, GrpcExceptionHandler }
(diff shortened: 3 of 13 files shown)


runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -39,6 +39,22 @@ object GrpcResponseHelpers {
       system: ClassicActorSystemProvider): HttpResponse =
     GrpcResponseHelpers(e, Source.single(GrpcEntityHelpers.trailer(Status.OK)), eHandler)
 
+  def responseForSingleElement[T](e: T, eHandler: ActorSystem => PartialFunction[Throwable, Trailers])(
+      implicit m: ProtobufSerializer[T],
+      writer: GrpcProtocolWriter,
+      system: ClassicActorSystemProvider): HttpResponse = {
+    def x = eHandler
+    def f = system
+    if (false) { x; f }
```
viktorklang, line 48: what?

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -39,6 +39,22 @@ object GrpcResponseHelpers {
...
+
+    val data = DataFrame(m.serialize(e))
```
viktorklang, line 50: Since `data` is an intermediary representation before encodeFrame, I'd probably pass it directly into encodeFrame.

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -39,6 +39,22 @@ object GrpcResponseHelpers {
...
+    val strictEntity = HttpEntity.Strict(writer.contentType, writer.encodeFrame(data).data)
+    val trailer = Trailer(GrpcEntityHelpers.trailer(Status.OK).trailers)
```
viktorklang, line 52: This could be a cached value, right?

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -39,6 +39,22 @@ object GrpcResponseHelpers {
...
+    HttpResponse(
+      headers = immutable.Seq(headers.`Message-Encoding`(writer.messageEncoding.name)),
+      entity = strictEntity).addAttribute(AttributeKeys.trailer, trailer)
```
viktorklang, line 55: Is it possible to add attributes in the HttpResponse constructor to avoid having to potentially allocate a new HttpResponse immediately?

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -39,6 +39,22 @@ object GrpcResponseHelpers {
...
+      system: ClassicActorSystemProvider): HttpResponse = {
+    def x = eHandler
+    def f = system
+    if (false) { x; f }
+
+    val data = DataFrame(m.serialize(e))
+    val strictEntity = HttpEntity.Strict(writer.contentType, writer.encodeFrame(data).data)
+    val trailer = Trailer(GrpcEntityHelpers.trailer(Status.OK).trailers)
+    HttpResponse(
+      headers = immutable.Seq(headers.`Message-Encoding`(writer.messageEncoding.name)),
```
patriknw, line 54: headers.`Message-Encoding`(writer.messageEncoding.name) :: Nil

patriknw (commented):
nice

johanandren (commented):
👍

jrudolph:
Also added the reverse optimization for being able to unmarshal strict entities.

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
jrudolph: The quickest way I found to ignore unused values, will remove it :)

runtime/src/main/scala/akka/grpc/javadsl/GrpcMarshalling.scala
```diff
@@ -55,6 +61,12 @@ object GrpcMarshalling {
         .via(new CancellationBarrierGraphStage)
         .mapMaterializedValue(japiFunction(_ => NotUsed)))
   }
+  def unmarshalStream[T](
+      entity: HttpEntity,
+      u: ProtobufSerializer[T],
+      @silent("never used") mat: Materializer,
```
viktorklang, line 67: It is used tho?

jrudolph:
> Also added the reverse optimization for being able to unmarshal strict entities.

Removed that commit and made a separate change for that in #1372

jrudolph:
Trying to fix the test I found that the current HTTP snapshot breaks the Akka HTTP client: if we generate a strict response with a content-length on the server, then we now create a `Default` entity on the client which cannot carry the trailers. I think we should probably disable creating `Default` entities for HTTP/2 for now (the only upside is avoiding a bit of wrapping of Chunks).

jrudolph:
This is done now but needs an updated to a version of akka-http that includes https://github.com/akka/akka-http/pull/3841.

jrudolph:
Updated to latest released Akka HTTP snapshot. Tests should now succeed.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -88,6 +88,8 @@ object GrpcProtocol {
       messageEncoding: Codec,
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
+      /** A shortcut to encode a data frame directly into a ByteString */
+      encodeDataToFrameBytes: ByteString => ByteString,
```
jrudolph, line 92: It's a bit unclear for me which parts of the runtime are really internal, and which are used from the generated code and subject to the binary compatibility guarantees.

raboof: yeah, that's tricky - we try to use `@InternalStableApi` for those places we know are used from the generated code, but it's hard to guarantee that is exhaustive. Adding this field looks fine to me, though!

runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -27,17 +38,50 @@ import scala.concurrent.{ ExecutionContext, Future }
...
       system: ClassicActorSystemProvider): HttpResponse =
-    GrpcResponseHelpers(e, Source.single(GrpcEntityHelpers.trailer(Status.OK)), eHandler)
+    GrpcResponseHelpers(e, Source.single(TrailerOk), eHandler)
+
+  def responseForSingleElement[T](e: T, eHandler: ActorSystem => PartialFunction[Throwable, Trailers])(
+      implicit m: ProtobufSerializer[T],
+      writer: GrpcProtocolWriter,
+      system: ClassicActorSystemProvider): HttpResponse =
+    try {
+      val strictEntity = HttpEntity.Strict(writer.contentType, writer.encodeDataToFrameBytes((m.serialize(e))))
```
jrudolph, line 61: This is the main change to create a Strict entity here.

interop-tests/src/test/scala/akka/grpc/interop/AkkaHttpServerProviderScala.scala
```diff
@@ -95,17 +96,22 @@ object AkkaHttpServerProviderScala extends AkkaHttpServerProvider with Directive
...
+                  HttpEntity.LastChunk(last.extension, f(last.trailer))
+              }))
         case _ =>
-          throw new IllegalArgumentException("Trailing response headers are only supported on Chunked responses")
-      }))
+          val origTrailers = response
+            .attribute(AttributeKeys.trailer)
+            .map(_.headers)
+            .getOrElse(Vector.empty)
+            .map(e => RawHeader(e._1, e._2))
```
patriknw, line 114: could these map+get+map be collapsed to one? maybe not in a hot path?

patriknw (approved):
LGTM

interop-tests/src/test/scala/akka/grpc/interop/AkkaHttpServerProviderScala.scala
jrudolph: That's test code right now.

Issue #1353, opened 2021-05-19
Update to Scala 2.12.14


raboof:
We're currently on Scala 2.12.11 because of scala/bug#12336 . Let's update when 2.12.14 comes out (which seems soon)

PR #1366, opened 2021-05-31
Avoids copying in BaseMarshalling when first read is total

viktorklang:
Reviewers, please have a look at this suggested PR which avoids allocating the ByteArrayOutputStream if the first read is smaller than the buffer (indicating that we hit EOF).

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
@@ -18,19 +18,28 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
     extends io.grpc.MethodDescriptor.Marshaller[T]
     with WithProtobufSerializer[T] {
   override def parse(stream: InputStream): T = {
-    val baos = new ByteArrayOutputStream(math.max(64, stream.available()))
     val buffer = new Array[Byte](32 * 1024)
-
     // Blocking calls underneath...
     // we can't avoid it for the moment because we are relying on the Netty's Channel API
-    var bytesRead = stream.read(buffer)
-    while (bytesRead >= 0) {
-      baos.write(buffer, 0, bytesRead)
-      bytesRead = stream.read(buffer)
+    val bytes = stream.read(buffer, 0, buffer.length) match {
+      case -1 =>
+        akka.util.ByteString.empty // EOF immediately
+      case n if n < buffer.length =>
+        akka.util.ByteString
+          .fromArrayUnsafe(buffer, 0, n) // Potentially wasteful, can we bet that the ByteString will be thrown away?
+      case full =>
+        val baos = new ByteArrayOutputStream(buffer.length * 2) // To avoid immediate resize
+        baos.write(buffer, 0, full)
+        var bytesRead = stream.read(buffer)
+        while (bytesRead >= 0) {
+          baos.write(buffer, 0, bytesRead)
+          bytesRead = stream.read(buffer)
+        }
+        akka.util.ByteString.fromArrayUnsafe(baos.toByteArray)
     }
-    protobufSerializer.deserialize(akka.util.ByteString.fromArrayUnsafe(baos.toByteArray))
-  }
 
+    protobufSerializer.deserialize(bytes)
+  }
 }
 
 /**
--- a/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
@@ -30,8 +30,9 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
       case -1 =>
         akka.util.ByteString.empty // EOF immediately
       case n if n < buffer.length =>
-        akka.util.ByteString
-          .fromArrayUnsafe(buffer, 0, n) // Potentially wasteful, can we bet that the ByteString will be thrown away?
+        // WARNING: buffer is retained in full below,
+        // which could be problematic if ProtobufSerializer.deserialize keeps a reference to the ByteString
+        akka.util.ByteString.fromArrayUnsafe(buffer, 0, n)
       case full =>
         val baos = new ByteArrayOutputStream(buffer.length * 2) // To avoid immediate resize
         var bytesRead = full
--- a/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
@@ -18,7 +18,12 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
     extends io.grpc.MethodDescriptor.Marshaller[T]
     with WithProtobufSerializer[T] {
   override def parse(stream: InputStream): T = {
-    val buffer = new Array[Byte](32 * 1024)
+    val buffer =
+      new Array[Byte](stream match {
+        case k: KnownLength => math.max(0, k.available()) // No need to oversize this if we already know the size
+        case _              => 32 * 1024
+      })
+
     // Blocking calls underneath...
     // we can't avoid it for the moment because we are relying on the Netty's Channel API
     val bytes = stream.read(buffer, 0, buffer.length) match {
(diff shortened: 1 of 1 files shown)


runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -18,19 +18,33 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
...
+        case _              => 32 * 1024
+      })
 
     // Blocking calls underneath...
     // we can't avoid it for the moment because we are relying on the Netty's Channel API
-    var bytesRead = stream.read(buffer)
-    while (bytesRead >= 0) {
-      baos.write(buffer, 0, bytesRead)
-      bytesRead = stream.read(buffer)
+    val bytes = stream.read(buffer, 0, buffer.length) match {
```
jrudolph, line 29: Is this overload better than just `stream.read(buffer)`?

jrudolph (approved):
LGTM for now. Would be nice to find a more principled solution.

Did you run some benchmarks to figure out whether it changes anything?

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
viktorklang: According to the javadoc they are semantically equivalent, but to someone who just glaces over the code might be tricked into thinking that it just reads as much as possible without blocking—so I opted to make it clear that I want buffer.length bytes or EOF.

viktorklang:
@jrudolph 

>Did you run some benchmarks to figure out whether it changes anything?

Yes and no, I was unsure what benchmarks would best exercise that logic, do you have a recommendation?

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -18,19 +18,33 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
...
+      case -1 =>
+        akka.util.ByteString.empty // EOF immediately
+      case n if n < buffer.length =>
+        akka.util.ByteString
+          .fromArrayUnsafe(buffer, 0, n) // Potentially wasteful, can we bet that the ByteString will be thrown away?
```
viktorklang, line 34: @jrudolph WDYT about this?

jrudolph: I don't think it should be a problem but I don't know the code paths in grpc exactly. @raboof what do you think?

raboof: So, the trade-off is that for streams with unknown length, we can either:
* copy the bytes we need (allocation overhead)
* hold on to the complete 32k buffer (GC pressure)

The buffer is almost immediately passed to `protobufSerializer.deserialize`, so the question is whether the deserialized object is likely to keep a reference to the buffer. *currently* both for Java and Scala we use `CodedInputStream` configured to assume the buffer is mutable (so it will copy instead of wrap). That doesn't necessarily need to be the case in the future, though, of course.

So for now this looks like a safe improvement, though we might want to record why in a comment

raboof (approved):
Looks like a good improvement.

Likely conflicts with #1318, which is still interesting to investigate in the future as well - but that shouldn't need to hold this up.

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
viktorklang: @raboof @jrudolph I'll add a comment. Alas, I am not sure we have any decent benchmarks exercising this part of the code.

viktorklang: Comment added

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -18,19 +18,35 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
...
-    while (bytesRead >= 0) {
-      baos.write(buffer, 0, bytesRead)
-      bytesRead = stream.read(buffer)
+    val bytes = stream.read(buffer, 0, buffer.length) match {
+      case -1 =>
+        akka.util.ByteString.empty // EOF immediately
+      case n if n < buffer.length =>
+        // WARNING: buffer is retained in full below,
+        // which could be problematic if ProtobufSerializer.deserialize keeps a reference to the ByteString
+        akka.util.ByteString.fromArrayUnsafe(buffer, 0, n)
```
raboof, line 35: Looking at this closer - is this reliable when the stream is not of known size? Seems like it could indeed truncate streams in that case.

viktorklang: @raboof Technically, should an InputStream be allowed to return a read-length which is less than what is requested, given that it is blocking, presuming that the actual length before EOF is not shorter than the requested length?

viktorklang: @raboof That said, either it should be `n <= buffer.length` or `n == buffer.length` depending on the answer to the question above.

viktorklang: @raboof 

How about something like this instead:

```scala
 override def parse(stream: InputStream): T = {
    val buffer =
      new Array[Byte](stream match {
        case k: KnownLength => math.max(0, k.available()) // No need to oversize this if we already know the size
        case _              => 32 * 1024
      })

    // Blocking calls underneath...
    // we can't avoid it for the moment because we are relying on the Netty's Channel API
    val initialBytes = stream.read(buffer, 0, buffer.length)
    val nextByte = if (initialBytes < 0) -1 else stream.read() // Test for EOF
    val bytes =
      if (nextByte == -1) {
        if (initialBytes < 1) akka.util.ByteString.empty // EOF immediately
        else {
          // WARNING: buffer is retained in full below,
          // which could be problematic if ProtobufSerializer.deserialize keeps a reference to the ByteString
          akka.util.ByteString.fromArrayUnsafe(buffer, 0, initialBytes)
        }
      } else {
        val baos = new ByteArrayOutputStream(buffer.length * 2) // To avoid immediate resize
        baos.write(buffer, 0, initialBytes)
        baos.write(nextByte)

        var bytesRead = stream.read(buffer)
        while (bytesRead >= 0) {
          baos.write(buffer, 0, bytesRead)
          bytesRead = stream.read(buffer)
        }

        akka.util.ByteString.fromArrayUnsafe(baos.toByteArray)
      }

    protobufSerializer.deserialize(bytes)
  }
```

viktorklang: @raboof The idea here is that we tests whether a plain single-byte read after the first chunk-read indeed is an EOF, and in that case we're already done.

viktorklang: Unfortunately I think there's an issue with the akka-grpc build:

```bash
[error] (akka-grpc-runtime / update) sbt.librarymanagement.ResolveException: Error downloading com.github.ghik:silencer-lib_2.12.14:1.7.1
[error]   Not found
[error]   Not found
[error]   not found: /Users/viktorklang/.ivy2/localcom.github.ghik/silencer-lib_2.12.14/1.7.1/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/github/ghik/silencer-lib_2.12.14/1.7.1/silencer-lib_2.12.14-1.7.1.pom
[error]   not found: https://oss.sonatype.org/content/repositories/staging/com/github/ghik/silencer-lib_2.12.14/1.7.1/silencer-lib_2.12.14-1.7.1.pom
```

raboof: Assuming InputStream is allowed to return a read-length which is less than what is requested (which indeed is not _entirely_ clear from the javadoc AFAICS) that seems like a safer approach, yes.

raboof: > Unfortunately I think there's an issue with the akka-grpc build

What commit are you on? master should be using Scala 2.12.11, not 2.12.14 (#1353)

viktorklang: @raboof I'm not sure what was going on, but I had to create #1373 to be able to run my build. Look at #1374 for my proposed solution.

raboof: were you running with `sbt ++2.12.14` perhaps?

viktorklang: @raboof Nope. go into master and do `project benchmarks` and then `update` and you should be getting the error as well.

raboof: ah, you're right - I didn't spot you were going into that project. looks like that's indeed somewhat broken. Aside from updating to 2.12.14 there seem to be a couple of things to fix, #1375.

PR #1374, opened 2021-06-03
Wip fix base marshaller √

viktorklang:


raboof (approved):
Looks safer to me. Too bad about the extra single-byte read, that could make the case where the input is streamed a little slower, but likely not by much compared to having to create the ByteArrayOutputStream so sounds like a reasonable trade-off.

This looks like it should be fine without the update to 2.12.14 (which needs some additional changes)...

johanandren (approved):

PR #1377, opened 2021-06-03
Wip base marshaller parse reimagined √

viktorklang:
Personally I think this solution is way more elegant, and we avoid the need for ByteArrayOutputStream


viktorklang:
Nevermind, edge case really requires that EOF test read()

viktorklang:
Lol, I found a hack—reading 0 bytes from an EOF stream should still produce an EOF (-1) so this recursive solution is both fast, and should be correct!

PR #1378, opened 2021-06-03
Wip base marshaller parse reimagined √

viktorklang:
Lol, I found a hack—reading 0 bytes from an EOF stream should still produce an EOF (-1) so this recursive solution is both fast, and should be correct!

Thoughts?

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
@@ -4,11 +4,12 @@
 
 package akka.grpc.internal
 
-import java.io.{ ByteArrayInputStream, ByteArrayOutputStream, InputStream }
+import java.io.{ ByteArrayInputStream, InputStream }
 
 import io.grpc.KnownLength
 import akka.annotation.InternalStableApi
 import akka.grpc.ProtobufSerializer
+import scala.annotation.tailrec
 
 /**
  * INTERNAL API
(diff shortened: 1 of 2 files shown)


viktorklang:
Lol, the "hack" is a JDK bug, as the javadoc states " If len is zero, then no bytes are read and 0 is returned; otherwise, there is an attempt to read at least one byte. If no byte is available because the stream is at end of file, the value -1 is returned; otherwise, at least one byte is read and stored into b. " <- https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#read(byte[],%20int,%20int)

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
```diff
@@ -24,28 +25,22 @@ abstract class BaseMarshaller[T](val protobufSerializer: ProtobufSerializer[T])
...
-        akka.util.ByteString.fromArrayUnsafe(baos.toByteArray)
+    @tailrec def readBytes(buffer: Array[Byte], at: Int, result: akka.util.ByteString): akka.util.ByteString = {
+      val capacity = buffer.length - at
+      val bytesRead = stream.read(buffer, at, capacity)
+      if (bytesRead < 0) {
+          if (at > 0) result ++ akka.util.ByteString.fromArrayUnsafe(buffer, 0, at) // Potentially wasteful since at could be small
+          else result
+      } else {
+        // Reading 0 bytes from an EOF stream should still EOF (-1) but didn't so has more data
+        if (capacity == 0 && bytesRead == 0)
```
raboof, line 36: so if `capacity == 0`, we call `read` just to find out if we're at EOF, but this way we don't need a single-byte `write` if there is more data after all.

Since we know by the contract of `read` that `capacity != 0 && bytesRead == 0` is invalid/impossible, I guess we don't strictly need the `capacity == 0` check here, so we could shave off something here - though that makes it even harder to follow for perhaps not too much gain.

This kinda starts to deserve a unittest, though ;)

raboof (approved):

runtime/src/main/scala/akka/grpc/internal/Marshallers.scala
viktorklang: @raboof I put that check in there if any InputStream implementation doesn't exhibit this behavior, and as such avoiding to go into an endless loop or breaking out before reading everything.

johanandren (commented):
Looks good, but trixy, so some test coverage would be good.

viktorklang:
@johanandren @raboof Added a couple of generative tests which should be able to find some bugs if there are any.

viktorklang:
The following looks like a much more nice solution though: https://github.com/akka/akka-grpc/pull/1318/files#diff-1a02adbd4ed9c032022756fe7dbbac1e967816613c7f1adef4f3e2330136fa11L22

lomigmegard:
> The following looks like a much more nice solution though: https://github.com/akka/akka-grpc/pull/1318/files#diff-1a02adbd4ed9c032022756fe7dbbac1e967816613c7f1adef4f3e2330136fa11L22

Without actual benchmarks is hard to know which approach is better especially between small (less than buffer size) and large messages. Maybe an hybrid could be to read the full stream in a single array when it's of `KnownLength` (with some max), and otherwise pass the stream further down (in both ScalaPB and Google it ends up calling `CodedInputStream.newInstance(stream)`).

Your approach is definitely an improvement over the previous implementation though, but as explain in https://github.com/akka/akka-grpc/pull/1318 it feels strange to make the decision to read the full stream in a `ByteString` whereas PB messages already have a method to decode from an `InputStream`.

jrudolph:
For now superseded by #1318

PR #1379, opened 2021-06-06
Add proto3 optionals support.

thesamet:
This change makes the Scala and Java generators declare that they support proto3
optionals (see https://github.com/protocolbuffers/protobuf/blob/master/docs/implementing_proto3_presence.md)

As akka-grpc generators operate at the message-level (rather than the
field-level) it is unexpected that any other changes in akka-grpc are
necessary.

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/javadsl/JavaCodeGenerator.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/javadsl/JavaCodeGenerator.scala
@@ -29,6 +29,7 @@ abstract class JavaCodeGenerator extends CodeGenerator {
 
   override def run(request: CodeGeneratorRequest, logger: Logger): CodeGeneratorResponse = {
     val b = CodeGeneratorResponse.newBuilder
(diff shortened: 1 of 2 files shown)


raboof (approved):
I wonder if we should support users passing in options to set these options - but this looks like a safe an useful thing to enable in any case, thanks!

insdami:
Is there a way to enable it? I'm using `sbt-akka-grpc` 2.1.6 along with `Compile / PB.protocOptions := Seq("--experimental_allow_proto3_optional"),`

I get the same error regardless
```
service.proto: is a proto3 file that contains optional fields, but code generator protoc-gen-jvm_4 hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--jvm_4_out: te 11s
```

I also looked at this issue [verification](https://github.com/scalapb/ScalaPB/commit/aaa0194d3240c309826dbc661d5f402ed5873ff7) which I tried but I keep on getting the same message.

thesamet:
Hi @insdami , this is enabled by default. What other generators you have in your build (look for `PB.targets`)?  If nothing comes up can you post a minimal reproducible example of this problem as a github repo?

PR #1380, opened 2021-06-07
Optimize Codecs

patriknw:
* especially Message-Accept-Encoding.findIn

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/Codecs.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Codecs.scala
@@ -5,19 +5,28 @@
 package akka.grpc.internal
 
 import akka.http.javadsl.{ model => jm }
+import akka.http.scaladsl.{ model => sm }
 import akka.grpc.GrpcServiceException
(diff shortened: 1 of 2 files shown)


raboof (approved):
:+1:

runtime/src/main/scala/akka/grpc/internal/Codecs.scala
```diff
@@ -5,33 +5,50 @@
...
+      Identity
+    } else if (accepted.length == 1) {
+      supportedByName.get(accepted(0)) match {
+        case Some(codec) => codec
+        case None        => Identity
+      }
+    } else {
+      accepted.collectFirst { case a if supportedNames.contains(a) => supportedByName(a) }.getOrElse(Identity)
+    }
+  }
```
viktorklang, line 51:
```scala
  accepted.length match {
     case 0 => Identity
     case 1 => supportedByName.get(accepted(0)).getOrElse(Identity)
     case _ => accepted.collectFirst(supportedNames).getOrElse(Identity)
  }
```

Might be cleaner?

PR #1381, opened 2021-06-07
Always close GZIPOutputStream

patriknw:
* it always included gzip because that's in the Codecs.supportedCodecs
* meaning that the server will gzip the responses
* I think gzip should be off by default, unless the grpc spec says otherwise (gzip is very slow)

I noticed some surprising gzip stuff showing up in profiles of our benchmarks.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
@@ -37,7 +37,7 @@ object GrpcRequestHelpers {
     HttpRequest(
       uri = uri,
       method = HttpMethods.POST,
-      // FIXME how can client exclude gzip?
+      // FIXME issue #1382 gzip shouldn't be included by default in Message-Accept-Encoding.
       headers = immutable.Seq(
         `Message-Encoding`(writer.messageEncoding.name),
         `Message-Accept-Encoding`(Codecs.supportedCodecs.map(_.name).mkString(",")),
--- a/runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
@@ -37,6 +37,7 @@ object GrpcRequestHelpers {
     HttpRequest(
       uri = uri,
       method = HttpMethods.POST,
+      // FIXME how can client exclude gzip?
       headers = immutable.Seq(
         `Message-Encoding`(writer.messageEncoding.name),
         `Message-Accept-Encoding`(Codecs.supportedCodecs.map(_.name).mkString(",")),
--- a/runtime/src/main/scala/akka/grpc/internal/Gzip.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/Gzip.scala
@@ -15,9 +15,8 @@ object Gzip extends Codec {
   override def compress(uncompressed: ByteString): ByteString = {
     val baos = new ByteArrayOutputStream(uncompressed.size)
     val gzos = new GZIPOutputStream(baos)
-    gzos.write(uncompressed.toArray)
-    gzos.flush()
-    gzos.close()
+    try gzos.write(uncompressed.toArray)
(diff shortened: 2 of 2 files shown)


runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
```diff
@@ -37,6 +37,7 @@ object GrpcRequestHelpers {
     HttpRequest(
       uri = uri,
       method = HttpMethods.POST,
+      // FIXME how can client exclude gzip?
```
patriknw, line 40: how should we support this? don't include the `Message-Accept-Encoding` header by default and only include it via the `headers` parameter? What will the end user api be for specifying the header? The low level metadata api or additional CallOption?

runtime/src/main/scala/akka/grpc/internal/Gzip.scala
```diff
@@ -15,9 +15,8 @@ object Gzip extends Codec {
   override def compress(uncompressed: ByteString): ByteString = {
     val baos = new ByteArrayOutputStream(uncompressed.size)
     val gzos = new GZIPOutputStream(baos)
-    gzos.write(uncompressed.toArray)
-    gzos.flush()
-    gzos.close()
+    try gzos.write(uncompressed.toArray)
+    finally gzos.close()
```
jrudolph, line 19: :+1: Good practice and also necessary here to free buffer from native gzip implementation bits immediately (which would only be freed by finalization otherwise).

jrudolph (approved):
LGTM

runtime/src/main/scala/akka/grpc/internal/Gzip.scala
patriknw: I'll open a separate PR for this

runtime/src/main/scala/akka/grpc/internal/GrpcRequestHelpers.scala
patriknw: created issue https://github.com/akka/akka-grpc/issues/1382 for this

patriknw, line 40:
```suggestion
      // FIXME issue #1382 gzip shouldn't be included by default in Message-Accept-Encoding.
```

runtime/src/main/scala/akka/grpc/internal/Gzip.scala
patriknw: changed my mind, just changed the title here

jrudolph:
> I think gzip should be off by default, unless the grpc spec says otherwise (gzip is very slow)

Compression is optional. It can be enabled selectively per request/response (even different for each of request and response), and even each message in a streamed call can be selectively compressed or not. In any case, every message will be compressed without context from preceding messages. This makes the whole scheme less useful. You can always add compression to individual messages without support of the protocol if hat is needed. So, I somewhat agree that maybe we should default to disable compression.

runtime/src/main/scala/akka/grpc/internal/Gzip.scala
raboof: Good to close also in case of exceptions, and good catch that flushing indeed doesn't seem necessary (https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/zip/DeflaterOutputStream.html#close() ) :+1:

runtime/src/main/scala/akka/grpc/internal/Gzip.scala
```diff
@@ -26,11 +25,13 @@ object Gzip extends Codec {
...
-      baos.write(buffer, 0, read)
-      read = gzis.read(buffer)
-    }
+    try {
+      var read = gzis.read(buffer)
+      while (read != -1) {
+        baos.write(buffer, 0, read)
+        read = gzis.read(buffer)
+      }
+    } finally gzis.close()
```
raboof, line 34: :hushed: that that was missing

raboof (approved):

PR #1389, opened 2021-06-18
Percent-encode grpc-message

note:
## Problem

Having such code:

```
new GrpcServiceException(Status.INVALID_ARGUMENT.withDescriptionSafe(e.getMessage))
```

may result in issuing invalid response if `e.getMessage` contains `\n` character. Since `grpc-message` is being transported with http headers it cannot contain arbitrary characters, see more [here](https://github.com/grpc/grpc-go/issues/576). What's recommended by standard is to percent-encode the message, see the [spec](https://github.com/grpc/grpc/issues/4672).

It's been added to grpc-java in [this PR](https://github.com/grpc/grpc-java/pull/1517/files).

## Why it is relevant

I saw this issue happening in a production setting. scalapb error messages can contain `\n` and it's getting problematic if you use linkerd. Linkerd will treat such response as incorrect. As a consequence the client will not see the original `grpc-status: 3, grpc-message: meaningful message` but `grpc-status: 13, grpc-message: protocol error: unspecific protocol error detected`.

## Solution

Copy and adapt grpc-java percent-encoding implementation.

## Other

I am not sure what's the most appropriate way of giving credits to grpc-java. Both project are on Apache 2.0 license, not sure what's required from license standpoint

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -13,6 +13,8 @@ object Common extends AutoPlugin {
 
   override def requires = JvmPlugin
 
+  private val consoleDisabledOptions = Seq("-Xfatal-warnings", "-Ywarn-unused", "-Ywarn-unused-import")
+
   override def globalSettings =
     Seq(
       organization := "com.lightbend.akka.grpc",
@@ -51,6 +53,7 @@ object Common extends AutoPlugin {
       "-P:silencer:globalFilters=Use LazyList instead of Stream",
       // ignore imports in templates
       "-P:silencer:pathFilters=.*.txt"),
+    Compile / console / scalacOptions ~= (_ filterNot consoleDisabledOptions.contains),
     javacOptions ++= List("-Xlint:unchecked", "-Xlint:deprecation"),
     Compile / doc / scalacOptions := scalacOptions.value ++ Seq(
       "-doc-title",
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -53,7 +53,7 @@ object Common extends AutoPlugin {
       "-P:silencer:globalFilters=Use LazyList instead of Stream",
       // ignore imports in templates
       "-P:silencer:pathFilters=.*.txt"),
-    Compile / console / scalacOptions ~= (_ filterNot consoleDisabledOptions.contains),
+    Compile / console / scalacOptions ~= (_.filterNot(consoleDisabledOptions.contains)),
     javacOptions ++= List("-Xlint:unchecked", "-Xlint:deprecation"),
     Compile / doc / scalacOptions := scalacOptions.value ++ Seq(
       "-doc-title",
--- a/runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
+++ b/runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
@@ -1,0 +1,108 @@
+/*
+ * Copyright (C) 2018-2021 Lightbend Inc. <http://www.lightbend.com>
+ * Copyright 2016, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package akka.grpc.scaladsl.headers
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+
+/**
+ * Excerpt from https://github.com/grpc/grpc/blob/7c9e8b425166276232653725de32ea0422a39b33/doc/PROTOCOL-HTTP2.md#responses:
+ *
+ * The value portion of Status-Message is conceptually a Unicode string description of the error, physically encoded as
+ * UTF-8 followed by percent-encoding. Percent-encoding is specified in RFC 3986 §2.1, although the form used here has
+ * different restricted characters. When decoding invalid values, implementations MUST NOT error or throw away the
+ * message. At worst, the implementation can abort decoding the status message altogether such that the user would
+ * received the raw percent-encoded form. Alternatively, the implementation can decode valid portions while leaving
+ * broken %-encodings as-is or replacing them with a replacement character (e.g., '?' or the Unicode replacement
+ * character).
+ */
+private[grpc] object PercentEncoding {
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L582
+  object Encoder {
+    private val HexArr: Array[Byte] =
+      Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+    def encode(value: String): String = {
+      val valueBytes = value.getBytes(StandardCharsets.UTF_8)
+      val firstIndexToEscape = valueBytes.indexWhere(isEscapingChar(_))
+      if (firstIndexToEscape == -1)
+        new String(valueBytes)
+      else
+        encodeSlow(valueBytes, firstIndexToEscape)
+    }
+
+    private def isEscapingChar(b: Byte): Boolean = b < ' ' || b >= '~' || b == '%'
+
+    private def encodeSlow(valueBytes: Array[Byte], riArg: Int): String = {
+      var ri = riArg
+      val escapedBytes = new Array[Byte](riArg + ((valueBytes.length - riArg) * 3))
+      // copy over the good bytes
+      if (riArg != 0) System.arraycopy(valueBytes, 0, escapedBytes, 0, riArg)
+      var wi = ri
+
+      while (ri < valueBytes.length) {
+        val b = valueBytes(ri)
+        // Manually implement URL encoding, per the gRPC spec.
+        if (isEscapingChar(b)) {
+          escapedBytes.update(wi, '%')
+          escapedBytes.update(wi + 1, HexArr((b >> 4) & 0xf))
+          escapedBytes.update(wi + 2, HexArr(b & 0xf))
+          wi += 3
+        } else {
+          escapedBytes.update(wi, b)
+          wi += 1
+        }
+        ri += 1
+      }
+      new String(escapedBytes, 0, wi, StandardCharsets.US_ASCII)
+    }
+  }
+
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L626
+  object Decoder {
+    private val TransferEncoding = StandardCharsets.US_ASCII
+
+    def decode(value: String): String =
+      if (value.indexOf('%') > -1)
+        decodeSlow(value)
+      else
+        value
+
+    private def decodeSlow(value: String): String = {
+      val source = value.getBytes(TransferEncoding)
+      val buf = ByteBuffer.allocate(source.length)
+      var i = 0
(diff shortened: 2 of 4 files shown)


runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +87,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
   override def renderInResponses = true
   override val companion = `Status-Message`
+  override def value() = toAsciiString(unencodedValue)
+
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L582
+  private def toAsciiString(value: String): String = {
+    val valueBytes = value.getBytes(StandardCharsets.UTF_8)
+    val firstIndexToEscape = valueBytes.indexWhere(isEscapingChar(_))
+    if (firstIndexToEscape == -1)
+      new String(valueBytes)
```
lomigmegard, line 104: return `value` instead of re-wrapping inside a new String?

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +87,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
 
   def findIn(headers: immutable.Seq[HttpHeader]): Option[String] =
     headers.collectFirst { case h if h.is(name) => h.value() }
+
+  private val HexArr: Array[Byte] =
+    Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L626
+  private def parseAsciiString(value: String): String = {
+    val transferEncoding = Charset.forName("US-ASCII")
```
lomigmegard, line 150: Use `StandardCharsets.US_ASCII` instead?

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +87,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
+    val firstIndexToEscape = valueBytes.indexWhere(isEscapingChar(_))
+    if (firstIndexToEscape == -1)
+      new String(valueBytes)
+    else
+      toAsciiStringSlow(valueBytes, firstIndexToEscape)
+  }
+
+  private def isEscapingChar(b: Byte): Boolean = b < ' ' || b >= '~' || b == '%'
+
+  private def toAsciiStringSlow(valueBytes: Array[Byte], riArg: Int): String = {
```
lomigmegard, line 111: move `toAsciiString`, `isEscapingChar` and `toAsciiStringSlow` to companion object? They seem to be all pure.

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
note: I know it seems like re-wrapping the same content but actually it's not

The following is not always true for any `value`: `new String(value.getBytes(StandardCharsets.UTF_8)) == value`. Consider `"a \ud801"`, its underlying array of bytes is `[97, 0, 32, 0, 1, -40]`. It's a UTF-16 string, note that even 'a' is encoded with 2 bytes. `"a \ud801".getBytes(StandardCharsets.UTF_8)` yields `[97, 32, 63]`, there are 2 effects: ascii characters are encoded with 1 byte and lone surrogate has been replaced with `?` (63 in ascii)

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +86,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
+        escapedBytes.update(wi + 1, HexArr((b >> 4) & 0xf))
+        escapedBytes.update(wi + 2, HexArr(b & 0xf))
+        wi += 3
+      } else {
+        escapedBytes.update(wi, b)
+        wi += 1
+      }
+      ri += 1
+    }
+    new String(jutil.Arrays.copyOf(escapedBytes, wi))
```
lomigmegard, line 169: Should be equivalent to the following, and being explicit for the charset is better instead of relying of the system's default:
`new String(escapedBytes, 0, wi, StandardCharsets.US_ASCII)`

patriknw (commented):
thanks for contributing, I have a few questions related to performance...

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +85,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
-  override def parse(value: String): Try[`Status-Message`] = Try(new `Status-Message`(value))
+  override def parse(value: String): Try[`Status-Message`] = Try(new `Status-Message`(parseAsciiString(value)))
 
   def findIn(headers: immutable.Seq[HttpHeader]): Option[String] =
     headers.collectFirst { case h if h.is(name) => h.value() }
+
+  private val HexArr: Array[Byte] =
+    Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L626
```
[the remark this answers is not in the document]
note: Moved to `PercentEncoding.scala`

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +85,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
+  private def parseAsciiString(value: String): String = {
+    val transferEncoding = StandardCharsets.US_ASCII
+    // This can be made faster if necessary.
```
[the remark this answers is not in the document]
note: It's used only for errors

runtime/src/main/scala/akka/grpc/scaladsl/headers/headers.scala
```diff
@@ -81,19 +85,86 @@ object `Status` extends ModeledCustomHeaderCompanion[`Status`] {
...
+    val source = value.getBytes(transferEncoding)
+    val buf = ByteBuffer.allocate(source.length)
```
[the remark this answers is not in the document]
note: You're right, reworked according to the suggestion

runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
```diff
@@ -0,0 +1,108 @@
...
+  // Copied with slight adaptations from https://github.com/grpc/grpc-java/blob/79e75bace40cea7e4be72e7dcd1f41c3ad6ee857/api/src/main/java/io/grpc/Status.java#L582
+  object Encoder {
+    private val HexArr: Array[Byte] =
+      Array('0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F')
+
+    def encode(value: String): String = {
+      val valueBytes = value.getBytes(StandardCharsets.UTF_8)
+      val firstIndexToEscape = valueBytes.indexWhere(isEscapingChar(_))
+      if (firstIndexToEscape == -1)
+        new String(valueBytes)
```
viktorklang, line 43: Missing character encoding in String constructor here

runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
```diff
@@ -0,0 +1,108 @@
...
+        value
+
+    private def decodeSlow(value: String): String = {
+      val source = value.getBytes(TransferEncoding)
+      val buf = ByteBuffer.allocate(source.length)
+      var i = 0
+      while (i < source.length) {
+        if (source(i) == '%' && i + 2 < source.length) {
+          try {
+            buf.put(Integer.parseInt(new String(source, i + 1, 2, TransferEncoding), 16).toByte)
```
viktorklang, line 92: Seems extremely costly to do this instead of looking at 2 bytes and create a number?

runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
note: Corrected

runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
note: True, reworked according to the suggestion. I ran microbenchmark and the difference for non-contrived input was significant indeed - 40% of speedup

runtime/src/main/scala/akka/grpc/scaladsl/headers/PercentEncoding.scala
```diff
@@ -0,0 +1,114 @@
...
+            i += 1
+          }
+        } else {
+          buf.put(source(i))
+          i += 1
+        }
+      }
+      new String(buf.array, 0, buf.position(), StandardCharsets.UTF_8)
+    }
+  }
```
viktorklang, line 112: It would be interesting to bench the solution above against the following:

```scala
private[this] final def decodeSlow(value: String): String = {
  @tailrec def loop(buf: ByteBuffer, source: Array[Byte], at: Int): String =
    if (at < source.length) {
      val cur = source(at)
      val skip = 
        if (cur != '%' || (at + 2) > source.length) {
          buf.put(cur)
          1
        } else {
          val ch0 = Character.digit(source(at + 1), 16)
          val ch1 = Character.digit(source(at + 2), 16)
          if (ch0 > -1 && ch1 > -1) {
            buf.put(((ch0 << 4) + ch1).toByte)
            3
          } else {
            buf.put(cur)
            1
          }
        }
      loop(buf, source, at + skip)
    } else new String(buf.array, 0, buf.position, StandardCharsets.UTF_8)

  val source = value.getBytes(TransferEncoding)
  val buf = ByteBuffer.allocate(value.length)

  loop(buf, source, 0)
}
```

Another possible improvement is to pass in the `idx` found by `indexOf('%')` and do an arraycopy from 0 … idx to avoid having to put those chars 1-by-1, as the worst-case the '%' is at the end of the string. Then you can start the `loop` at `idx` and progress from there.

patriknw: It's good that we try to find a decent performance level for this, but note that it's only for errors and that shouldn't be extremely performance sensitive.

note: With the code pasted by @viktorklang I got `0,652 ± 0,003  us/op` while with the code currently in the PR I got `0,609 ± 0,009  us/op`.

I agree with @patriknw that it's good to have reasonable performance but IMO it's not critical to get the best since it's only for errors

project/Common.scala
```diff
@@ -51,6 +53,7 @@ object Common extends AutoPlugin {
       "-P:silencer:globalFilters=Use LazyList instead of Stream",
       // ignore imports in templates
       "-P:silencer:pathFilters=.*.txt"),
+    Compile / console / scalacOptions ~= (_.filterNot(consoleDisabledOptions.contains)),
```
raboof, line 56: this change doesn't really belong in this PR, but I like it anyway ;)

runtime/src/test/scala/akka/grpc/scaladsl/headers/HeadersSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+      val examples = Table(
+        "examples",
+        "example%",
+        "yet another%%example",
+        "abc\ndef",
+        "abc\ndef\n",
+        "abc\r\ndef\r\n",
+        "abc\r\n\r\ndef\r\n",
+        "Καλημέρα κόσμε",
+        "コンニチハ",
```
raboof, line 78: 今日は!

raboof (approved):
Great, thanks!

PR #1398, opened 2021-06-29
simplify protocol selection

jrudolph:


runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -121,7 +119,13 @@ object GrpcProtocol {
    * Detects which gRPC protocol variant is indicated by a mediatype.
    * @return a [[GrpcProtocol]] matching the request mediatype if known.
    */
-  def detect(mediaType: jmodel.MediaType): Option[GrpcProtocol] = protocols.find(_.mediaTypes.contains(mediaType))
+  def detect(mediaType: jmodel.MediaType): Option[GrpcProtocol] = mediaType.subType match {
+    // FIXME: do we need to check mainType?
+    case "grpc" | "grpc+proto"                   => Some(GrpcProtocolNative)
```
jrudolph, line 124: No need for relatively complex collection logic when we can just match on strings.

runtime/src/main/scala/akka/grpc/scaladsl/GrpcMarshalling.scala
```diff
@@ -45,8 +45,8 @@ object GrpcMarshalling {
 
   def negotiated[T](req: HttpRequest, f: (GrpcProtocolReader, GrpcProtocolWriter) => Future[T]): Option[Future[T]] =
     GrpcProtocol.negotiate(req).map {
-      case (maybeReader, writer) =>
-        maybeReader.map(reader => f(reader, writer)).fold(Future.failed, identity)
+      case (Success(reader), writer) => f(reader, writer)
```
jrudolph, line 48: Avoid higher order function.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -121,7 +119,13 @@ object GrpcProtocol {
    * Detects which gRPC protocol variant is indicated by a mediatype.
    * @return a [[GrpcProtocol]] matching the request mediatype if known.
    */
-  def detect(mediaType: jmodel.MediaType): Option[GrpcProtocol] = protocols.find(_.mediaTypes.contains(mediaType))
+  def detect(mediaType: jmodel.MediaType): Option[GrpcProtocol] = mediaType.subType match {
+    // FIXME: do we need to check mainType?
```
raboof, line 123: it's not very likely that we would encounter other types where the subtype exactly matches the constants we know of here, so just matching the subtype might indeed be OK

jrudolph:
```suggestion
    // mainType is not checked. We assume it's the right one.
```

patriknw (approved):
LGTM

Issue #1396, opened 2021-06-29
Using Scala 2.13 artifacts with Scala 3


raboof:
Generally it should be possible to use Scala 2.13 artifacts in Scala 3 projects.

However, when using Akka gRPC in a Scala 3 project via the sbt plugin, the dependency on the runtime library is not correctly added: when deciding which version of the runtime library to add, Akka gRPC automatically adds the one for the 'current' Scala version (https://github.com/akka/akka-grpc/blob/master/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala#L200-L210), while in this case it should use the sbt `.cross()` feature to use the 2.13 library on Scala 3.

Perhaps we should give the runtime library its own sbt key, so at least it can be overridden?

RomanLebid:
There's a workaround for this issue that worked for us by adding 
`inConfig(Compile)(scalaBinaryVersion := "2.13")`

And that works alright for generating server code, but clients fail to compile
I assume that's different because of compiler changes related to type inference
```
32 |  private val clientState = new ClientState(settings, akka.event.Logging(sys.classicSystem, this.getClass))
   |                                                                                                          ^
   |Cannot find LogSource for Class[? <: grpc.health.v1.DefaultHealthClient] please see ScalaDoc for LogSource for how to obtain or construct one..
   |I found:
   |
   |    akka.event.LogSource.fromAnyClass[Nothing]
   |
   |But method fromAnyClass in object LogSource does not match type akka.event.LogSource[Class[? <: grpc.health.v1.DefaultHealthClient]].
```

RomanLebid:
Yeah, something has definitely change around existential types

Was only able to compile clients, by not generating them, copying the source code and changing this line
https://github.com/akka/akka-grpc/blob/e6f0402e4eab6ab87bba970cb0c4674b881f13cf/codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt#L50
to using the actual class and not the class of current instance
```
private val clientState = new ClientState(settings, akka.event.Logging(sys.classicSystem, classOf[DefaultHealthClient]))
```

raboof:
Surprising, but seems like a fine workaround - would you like to PR it?

johanandren:
That's the same workaround for that we've been doing on the Akka side, so I think it is good.

RomanLebid:
> Surprising, but seems like a fine workaround - would you like to PR it?

Yes, I think I'll make a PR for this issue

[...]

PR #1403, opened 2021-07-08
Fix incompatibilities when using Scala 2.13 artifacts with Scala 3

RomanLebid:
This does not add proper Scala 3 support for this project, but only fixes compiler errors when using 2.13 artifacts with Scala 3:

* use 2.13 Scala binary version, when Scala 3 is used, to pull dependencies
* fix type inference with existential types for LogSource
* don't indent generated methods too far to the right

References #1396

### DIFF
--- a/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
+++ b/sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
@@ -10,7 +10,7 @@ import akka.grpc.gen.javadsl.{ JavaClientCodeGenerator, JavaInterfaceCodeGenerat
 import akka.grpc.gen.{ ProtocSettings, Logger => GenLogger }
 import protocbridge.Generator
 import sbt.Keys._
-import sbt.{ GlobFilter, _ }
+import sbt._
 import sbtprotoc.ProtocPlugin
 import scalapb.ScalaPbCodeGenerator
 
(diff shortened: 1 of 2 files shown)


sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -106,6 +106,13 @@ object AkkaGrpcPlugin extends AutoPlugin {
       (if (config == Compile || config == Test) Seq() // already supported by sbt-protoc by default
        else sbtprotoc.ProtocPlugin.protobufConfigSettings) ++
       Seq(
+        // workaround to allow using Scala 2.13 artifacts in Scala 3 projects
+        scalaBinaryVersion := {
```
raboof, line 110: Hmm, wouldn't this interfere with other uses of `scalaBinaryVersion`? Perhaps it would be safer to move this inside `sandboxedGenerator`/`plainGenerator` below, so the effect is more focused?

raboof (commented):
Thanks for looking into this!

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
RomanLebid: binary version is used for pulling dependencies too, not just code generation
tried moving it into `generator` methods and published locally, that doesn't work as tries to pull Scala 3 versions
```
sbt.librarymanagement.ResolveException: Error downloading com.lightbend.akka.grpc:akka-grpc-runtime_3:2.1.0-M1-4-4ca3e1aa-20210708-0938-SNAPSHOT
```

johanandren (approved):

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
RomanLebid: @raboof okay, seems to work now, moved Scala binary version for generators and their dependencies
published it locally and seems to work fine with my Scala 3 project 👍

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -115,9 +115,9 @@ object AkkaGrpcPlugin extends AutoPlugin {
           generatorsFor(
             akkaGrpcGeneratedSources.value,
             akkaGrpcGeneratedLanguages.value,
-            ScalaBinaryVersion(scalaBinaryVersion.value),
+            generatorScalaBinaryVersion.value,
             generatorLogger) ++ akkaGrpcExtraGenerators.value.map(g =>
-            GeneratorBridge.plainGenerator(g, ScalaBinaryVersion(scalaBinaryVersion.value), generatorLogger))
+            GeneratorBridge.plainGenerator(g, generatorScalaBinaryVersion.value, generatorLogger))
```
raboof, line 120: :+1: good catch!

raboof (approved):
Great!

PR #1406, opened 2021-07-12
Add scala3 scripted test

### LINKED ISSUE #1396: Using Scala 2.13 artifacts with Scala 3

raboof:
Closes #1396


octonato (approved):
LGTM

No braces!!!

PR #1411, opened 2021-07-19
Fully migrate from travis to github actions

jrudolph:


.github/workflows/build-test.yml
```diff
@@ -28,7 +28,9 @@ jobs:
       - name: Code style check and binary-compatibility check
         run: |-
           cp .jvmopts-travis .jvmopts
-          sbt scalafmtCheckAll scalafmtSbtCheck headerCheckAll grpcVersionSyncCheck +mimaReportBinaryIssues
+          sbt scalafmtCheckAll scalafmtSbtCheck headerCheckAll grpcVersionSyncCheck +mimaReportBinaryIssues whitesourceCheckPolicies
```
jrudolph, line 31: Run whitesourceCheck also on PRs and not only after publishing in `main`.

.github/workflows/publish.yml
```diff
@@ -29,8 +31,31 @@ jobs:
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
 
+      - name: Update Whitesource Listing
+        run: sbt whitesourceUpdate
+        env:
+          WHITESOURCE_PASSWORD: "${{ secrets.WHITESOURCE_PASSWORD }}"
+
   # TODO publish gradle from here as well
```
raboof, line 39: We can drop the TODO now ;)

.github/workflows/publish.yml
```diff
@@ -29,8 +31,31 @@ jobs:
...
+          # we don't know what commit the last tag was it's safer to get entire repo so previousStableVersion resolves
+          fetch-depth: 0
+
+      - name: Set up JDK 8
+        uses: olafurpg/setup-scala@v10
+        with:
+          java-version: adopt@1.8
+
+      - name: Publish Plugin to Gradle Plugin Repository
+        run: cd gradle-plugin && ./gradlew publishPlugins -Pgradle.publish.key='${{ secrets.GRADLE_PUBLISH_KEY }}' -Pgradle.publish.secret='${{ secrets.GRADLE_SECRET }}'
```
raboof, line 57: I wonder why we didn't do it like this on the travis side before. :+1:

raboof (approved):
really great. we can remove that TODO now ;)

.github/workflows/publish.yml
jrudolph, line 57: Does the gradle plugin itself depend on anything from the sbt build? Or only the builds using it? So, would we have to publish the sbt artifacts before?

jrudolph: Should probably work, as it's the same as in the existing travis build.

jrudolph: Yep, don't know I found it in the docs.

PR #1419, opened 2021-07-26
Update to akka-http 10.2.5

jrudolph:


ignasi35 (approved):
Should we also bump the Akka version?

IIUC, the main reason to bump the Akka HTTP version is performance. Wasn't there some performance improvement in Akka core with a positive impact on Akka HTTP?

PR #1433, opened 2021-08-24
[Proposal] Chain of Filters for a rpc

sebastian-alfers:
## Idea
This PR is exploring and prototyping my proposal to introduce a composable and reusable notion of `GenericFilter[In, Out]` for a given protobuf defined `rpc`. The term `GenericFilter` is just an idea and is inspired by [finagle-filters](https://twitter.github.io/finagle/guide/ServicesAndFilters.html#filters).

## Motivation
Provide a composable approach for individual components to be able to separate concerns and keep the implementation of the service to a minimal level.

Possible areas of application you may want to chain in front of your service but do actually not belong there:
- authentication
- telemetry
- timeouts
- caching
- request enhancements
- validation
- ...

## Details
In the context of this PR a `GenericFilter[In, Out]` is highly coupled the definition of a `rpc`. Given the following definition
```
rpc SayHello (HelloRequest) returns (HelloReply) {}
```

which is basically a `HelloRequest => Future[HelloReply]`. Therefore a filter would become a `GenericFilter[HelloRequest, Future[HelloReply]]`.

The base signature of a filter looks like this:

```
trait GenericFilter[In, Out] {
  def apply(request: In, next: In => Out)(implicit ex: ExecutionContext) : Out
}
```

The most basic implementation of an identity filter looks like this:
```
new GenericFilter[HelloRequest, Future[HelloReply]] {
  override def apply(request: HelloRequest, next: HelloRequest => Future[HelloReply])(
      implicit ex: ExecutionContext): Future[HelloReply] = next(request)
}
```

Each `rpc` has its own chain of filters. By default, no filter is configured:
```
trait GreeterService {
  
  def sayHello(in: example.myapp.helloworld.grpc.HelloRequest): scala.concurrent.Future[example.myapp.helloworld.grpc.HelloReply]

  val sayHelloFilter: Seq[GreeterServiceFilterApi.GenericFilter[example.myapp.helloworld.grpc.HelloRequest, scala.concurrent.Future[example.myapp.helloworld.grpc.HelloReply]]] = Nil
}
```

The final composition and execution of the chain of Filters looks like this:
```
def compose[In, Out](request: In, filter: Seq[GenericFilter[In, Out]], client: In => Out)(implicit ex: ExecutionContext): Out = {
  filter match {
    case Nil => client(request)
    case head :: Nil => head(request, client)(ex)
    case head :: tail =>      
      val next: In => Out = req => compose(req, tail, client)        
      head(request, next)(ex)
  }
}
```

The main point is that the "next" item in the chain is passed as a function literal so it gets evaluated lazy as soon as the filter decides to call the next stage.

## Filters in detail
Filter can be chained and will be evaluated in order they are defined. A filter is able to receive the unmarshalled request and may
- forward it downstream
- forward a modified version downstream (`.copy( ... )`)
- decide to not call the next "stage" (which may be the next filter or the service itself) to short circuit.

Furthermore a filter receives the response from the next "stage" (which may be the next filter or the service itself) and may
- return it as a response
- return a modified version of the response

## Usage

To opt in to the generation of the `FilterApi` works similar to the way how `server_power_apis` is enabled on the build level.

Currently there is support for `FilterApi` and `FilterPowerApi` and work for request and/or response streaming.

## Limitations

A filter is not able to translate between events that are not part of the protobuf `rpc` definition. This means for the example a Filter can receive `HelloRequest` and return `HelloReply`. This is caused by the tight coupling of the plain http-request payload, path, marshalling and service invocation of akka gRPC.

---

What is missing:
- [ ] Java api


sebastian-alfers:
The idea of this PR was to go beyond the "before parsing" stage as composing on a plain Request/Response level should be  possible: https://doc.akka.io/docs/akka-grpc/current/server/akka-http.html.

Issue #1459, opened 2021-09-22
grpc-web does not work with akka-gRPC 2.1.0 / akka-http 10.2.6


kallependers:
After updating akka-http from 10.2.4 to 10.2.6 and akka-grpc from 2.0.0 to 2.1.0 gRPC web requests (accept: application/grpc-web-text) are failing.

- no errors in the logs
- the gRPC request succeeds
- the response fails

So I guess it goes somewhere wrong when generating the proper response for the accept.

jrudolph:
Moved to akka-grpc.

Thanks for the report, @kallependers. Interesting that this wouldn't be tested.

jrudolph:
Could you post the full output of using `curl -v` to test this issue?

kallependers:
```
$ curl -v 'https://example.com/grpc/Endpoint' \
-H 'authority: example.com' \
-H 'sec-ch-ua: "Google Chrome";v="93", " Not;A Brand";v="99", "Chromium";v="93"' \
-H 'x-user-agent: grpc-web-javascript/0.1' \
-H 'sec-ch-ua-mobile: ?0' \
-H 'authorization: Bearer JWT' \
-H 'content-type: application/grpc-web-text' \
-H 'accept: application/grpc-web-text' \
-H 'x-grpc-web: 1' \
-H 'user-agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/93.0.4577.82 Safari/537.36' \
-H 'sec-ch-ua-platform: "Linux"' \
-H 'origin: https://example.com' \
-H 'sec-fetch-site: same-site' \
-H 'sec-fetch-mode: cors' \
-H 'sec-fetch-dest: empty' \
-H 'accept-language: en-US,en;q=0.9,nl;q=0.8' \
--data-raw 'AAAAAAA=' \
--compressed
*   Trying IP:443...
* Connected to example.com (IP) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*  CAfile: /etc/ssl/certs/ca-certificates.crt
*  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
[...]

PR #1462, opened 2021-10-01
Fix request/response issues with grpc-web-text (#1459)

drmontgomery:
* Modify strict decode path in AbstractGrpcPrototocol to apply the
flow adapter before reading frame.
* Modify GrpcResponseHelper.responseWithTrailers to create the
HttpResponse with HTTP 1/1 to fix exception in HttpResponseRender.
This  isn't technically correct, but it matches the behavior of
akka-grpc v2.0.0 and appears to work with both HTTP/1.1 and HTTP/2.0.

References #1459


raboof (approved):

jrudolph (approved):
LGTM, thanks a lot for figuring this out @drmontgomery!

jrudolph:
More test coverage would be nice, but I also created #1465 to work on that separately.

PR #1516, opened 2021-12-14
build: use +~ for cross-building

jrudolph:
Porting raboof's command from https://github.com/akka/akka/pull/30981/files

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -20,6 +20,7 @@ object Common extends AutoPlugin {
 
   override def globalSettings =
     Seq(
+      commands += switchVersion,
       organization := "com.lightbend.akka.grpc",
       organizationName := "Lightbend Inc.",
       organizationHomepage := Some(url("https://www.lightbend.com/")),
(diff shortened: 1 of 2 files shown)


jrudolph:
Scala 3 scripted test fails because the `_~` cannot parse with arguments.

raboof:
> Scala 3 scripted test fails because the `+~` cannot parse with arguments.

Hmm, I hoped at least `sbt "+~ 2.13 \"sbt-akka-grpc/scripted scala3/*\""` would do the trick (possibly by adding extra `"` to the `++` invocation inside), but it looks like that doesn't work either..

project/Common.scala
```diff
@@ -77,4 +78,30 @@ object Common extends AutoPlugin {
...
+        parsed.resultEmpty match {
+          case e: sbt.internal.util.complete.Parser.Failure =>
+            throw new IllegalStateException(e.errors.mkString(", "))
+          case sbt.internal.util.complete.Parser.Value(v) =>
+            v()
+        }
+      }
+      val commands = s"++$requestedVersion" +: args.tail
+      commands.foldLeft(initialState)(run)
+    }
```
He-Pin, line 105: it would be great to port this to sbt too.

raboof (approved):

project/Common.scala
raboof: I would like that, though switching has been a much-discussed topic over the years - I'm curious if people would like this approach. I opened the discussion at https://github.com/sbt/sbt/discussions/6893

jrudolph: Alternatively, we could consider moving to https://github.com/sbt/sbt-projectmatrix in the future.

raboof: > Alternatively, we could consider moving to https://github.com/sbt/sbt-projectmatrix in the future.

Yeah, that might be interesting in any case

Issue #1528, opened 2021-12-23
Improve serverReflectionInfo response time


github-samuel-clarenc:
## Short description 

The server reflection can be long for large proto files. 

## Details 

The [ServerReflectionImpl.serverReflectionInfo](https://github.com/akka/akka-grpc/blob/main/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala#L35) can be relatively long to execute, depending on the proto size. This is mainly due to the execution of [AbstractMessageLite.toByteString](https://github.com/protocolbuffers/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/AbstractMessageLite.java#L56) which is called each time. 

## Solution proposal

We could introduce a simple `Map` to cache the proto as a ByteString to avoid calling `toByteString` every time.

raboof:
Sounds reasonable to me - would you be up for creating a PR?

github-samuel-clarenc:
> Sounds reasonable to me - would you be up for creating a PR?

Sure, I'm currently working on it (and checking that we can gain something with the jmh benchmarks!)

PR #1530, opened 2021-12-23
Cache proto ByteString in server reflection (#1528)

github-samuel-clarenc:
References #1528

This PR adds a simple cache to the proto to bytes method in ServerReflection. 

The unit tests are completed in this PR but we can't really test whether the cache is actually used or not.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
@@ -5,12 +5,15 @@
 package akka.grpc.internal
 
 import com.google.protobuf.Descriptors.FileDescriptor
-
 import akka.NotUsed
 import akka.annotation.InternalApi
 import akka.stream.scaladsl._
-
 import _root_.grpc.reflection.v1alpha.reflection._
+import com.google.protobuf.ByteString
+
+import java.util.concurrent.ConcurrentHashMap
+import scala.collection.concurrent
+import collection.JavaConverters._
 
 /**
  * INTERNAL API
@@ -19,6 +22,9 @@ import _root_.grpc.reflection.v1alpha.reflection._
 final class ServerReflectionImpl private (fileDescriptors: Map[String, FileDescriptor], services: List[String])
     extends ServerReflection {
   import ServerReflectionImpl._
+
+  private val protoBytesLocalCache: concurrent.Map[String, ByteString] =
+    new ConcurrentHashMap[String, ByteString]().asScala
 
   def serverReflectionInfo(in: Source[ServerReflectionRequest, NotUsed]): Source[ServerReflectionResponse, NotUsed] = {
     in.map(req => {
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
```diff
@@ -52,6 +55,9 @@ final class ServerReflectionImpl private (fileDescriptors: Map[String, FileDescr
       ServerReflectionResponse(req.host, Some(req), response)
     })
   }
+
+  private def getProtoBytes(fileDescriptor: FileDescriptor): ByteString =
+    protoBytesLocalCache.getOrElseUpdate(fileDescriptor.getName, fileDescriptor.toProto.toByteString)
```
raboof, line 60: I think this isn't thread-safe, is it? Might be worth picking a concurrent implementation, perhaps even just good old `ConcurrentHashMap`

raboof (commented):
Looks good, thanks for also extending the tests. One comment for consideration, otherwise LGTM!

runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala
github-samuel-clarenc: Indeed it's not thread safe this way, I'll make the change !

github-samuel-clarenc: Done !

raboof (approved):
Nice one, thanks!

raboof:
Ah, it looks like it doesn't compile on Scala 2.13 though:
```
[error] /home/runner/work/akka-grpc/akka-grpc/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala:16:8: value mapAsScalaConcurrentMapConverter is not a member of object scala.jdk.CollectionConverters
[error] import scala.jdk.CollectionConverters.mapAsScalaConcurrentMapConverter
[error]        ^
[error] /home/runner/work/akka-grpc/akka-grpc/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala:27:49: value asScala is not a member of java.util.concurrent.ConcurrentHashMap[String,com.google.protobuf.ByteString]
[error]     new ConcurrentHashMap[String, ByteString]().asScala
```

github-samuel-clarenc:
> Ah, it looks like it doesn't compile on Scala 2.13 though:
> 
> ```
> [error] /home/runner/work/akka-grpc/akka-grpc/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala:16:8: value mapAsScalaConcurrentMapConverter is not a member of object scala.jdk.CollectionConverters
> [error] import scala.jdk.CollectionConverters.mapAsScalaConcurrentMapConverter
> [error]        ^
> [error] /home/runner/work/akka-grpc/akka-grpc/runtime/src/main/scala/akka/grpc/internal/ServerReflectionImpl.scala:27:49: value asScala is not a member of java.util.concurrent.ConcurrentHashMap[String,com.google.protobuf.ByteString]
> [error]     new ConcurrentHashMap[String, ByteString]().asScala
> ```

:scream: 

I changed the collection converter to `collection.JavaConverters._` which seems to work for both 2.12 and 2.13 !

raboof:
> I changed the collection converter to `collection.JavaConverters._` which seems to work for both 2.12 and 2.13 !

Jep, looks like that did the trick! Thanks!

PR #1539, opened 2022-01-03
Generated annotation

### LINKED ISSUE #1538: Introduce @AkkaGrpcGenerated annotation to generate classes
This can be useful to provide an api for tool like e.g. JaCoCo to exclude generated classes from the report.

The annotation can look like:
```
import java.lang.annotation.{Retention, RetentionPolicy}
@Retention(RetentionPolicy.RUNTIME)
trait AkkaGrpcGenerated {}
```

and could be added like this to generated classes:

```
@AkkaGrpcGenerated
class Something ...
```

sebastian-alfers:
Open questions:
- [x] Should Java's `static class` also get this annotation?
- [x] Should  Scala's `object` also get this annotation?

Fixes #1538


raboof:
>     * [ ]  Should Java's `static class` also get this annotation?
>     * [ ]  Should  Scala's `object` also get this annotation?

I guess all generated classes should get the annotation? Any reason why not?

sebastian-alfers:
> Any reason why not?

Nothing I am aware of, thanks for confirming :-)

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -21,13 +21,17 @@ import static @{service.packageName}.@{service.name}.Serializers.*;
...
+
 @GenMethodImports(service)
 
+@@AkkaGrpcGenerated
 public abstract class @{service.name}Client extends @{service.name}ClientPowerApi implements @{service.name}, AkkaGrpcClient {
   public static final @{service.name}Client create(GrpcClientSettings settings, ClassicActorSystemProvider sys) {
     return new Default@{service.name}Client(settings, sys);
   }
 
+  @@AkkaGrpcGenerated
```
raboof, line 34: I'm not sure this is still necessary since it's an inner class of a class that is already annotated with `AkkaGrpcGenerated`, but cant hurt ;)

raboof (commented):
Maybe also `ScalaServer/PowerApiTrait.scala.txt`?

Otherwise LGTM. Would be good to actually test with JaCoCo, but I'd also be OK with merging and testing against the published snapshot

sebastian-alfers:
Added the annotation for all traits and tested locally with JaCoCo - all good. 

Now the only generated classes that are still in the report are based on the `scalapb` codegen which we do not control.

raboof:
> Added the annotation for all traits and tested locally with JaCoCo - all good.

Great!

> Now the only generated classes that are still in the report are based on the `scalapb` codegen which we do not control.

In some cases we might be able to tweak via https://scalapb.github.io/docs/customizations/#adding-annotations - but perhaps that's a 'next step'

sebastian-alfers:
Maybe I misunderstood this but I think this has to be added directly into the proto files so that should be done by the end-user?!

raboof:
> Maybe I misunderstood this but I think this has to be added directly into the proto files so that should be done by the end-user?!

Yes, I think so, too. That might be fine, since it's the user who wants to run JaCoCo and optimize the reporting?

sebastian-alfers:
@raboof Anything else you would like to see here? Or is this good to go?

raboof (approved):
Thanks for the ping, looks good now indeed!

PR #1540, opened 2022-01-05
Simplify Rich error model api

sebastian-alfers:
This PR introduces a `AkkaGrpcException` to provide some convenience over the "manual" approach using the `io.grpc.StatusRuntimeException` directly. This new api is called the "native" approach.

The main change is happening in the `RequestBuilderImpl` that adds recovery and converts `StatusRuntimeException` into `AkkaGrpcException`.

Backwards compatibility should be maintained as `AkkaGrpcException` is a subtype of `StatusRuntimeException`. 

The docs now only mention the new "native" approach but the test for the previous "manual" approach are still there.

This change also pulls in the dependency `io.grpc.grpc-protobuf`.


runtime/src/main/scala/akka/grpc/AkkaGrpcException.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package akka.grpc
+
+import com.google.protobuf.any.Any
+import com.google.rpc.{ Code, Status }
+import io.grpc.StatusRuntimeException
+import io.grpc.protobuf.StatusProto
+import collection.JavaConverters._
+
+object AkkaGrpcException {
```
raboof, line 13: It's a bit weird that we now have both a `AkkaGrpcException` and a `GrpcServiceException`.

I wonder if we could unify them, but that might not be easy. For the client side, I think it is nice that `AkkaGrpcException` extends `StatusRuntimeException`, however I'm not sure if it'd make sense to have `GrpcServiceException` extend `StatusRuntimeException`. Needs further investigation.

runtime/src/main/scala/akka/grpc/AkkaGrpcException.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def apply(code: Int, message: String, details: Option[scalapb.GeneratedMessage]): AkkaGrpcException = {
```
raboof, line 15: I think having a convenience constructor that takes code, message and details is a great idea.

For `code`, I think having `com.google.rpc.Code` is slightly neater compared to `Int`

For `details`, it might be nice to accept a `Seq`.

interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorModelNativeTest.java
```diff
@@ -0,0 +1,83 @@
...
+            AkkaGrpcException statusEx = response.toCompletableFuture().handle((res, ex) -> {
+                return (AkkaGrpcException) ex;
+            }).get();
+
+            assertEquals("type.googleapis.com/google.rpc.LocalizedMessage", statusEx.status().getDetails(0).getTypeUrl());
+
+            assertEquals(Status.INVALID_ARGUMENT.getCode().value(), statusEx.code());
+            assertEquals("What is wrong?", statusEx.message());
+
+            LocalizedMessage details = statusEx.getDetails(0, com.google.rpc.error_details.LocalizedMessage.messageCompanion());
```
raboof, line 71: I'm not too sold on this API: it's still pretty verbose, and you still need to first check the `typeUrl` to see the type before you can call `getDetail` with the correct `messageCompanion` to parse it.

I wonder if we should generalize the `AnySupport` helper class we use in the akkaserverless-java-sdk (https://github.com/lightbend/akkaserverless-java-sdk/blob/main/sdk/java-sdk/src/main/scala/com/akkaserverless/javasdk/impl/AnySupport.scala), though turning that into a public API would probably require quite some API design effort. I wonder if there's any conventions around this on the grpc-java side yet?

That way, you could just use the regular `status.getDetailsList.asScala`, and turn them into domain classes with that more general-purpose utility.

As a less ambitious first step, we could provide a `GrpcServiceExtension.getDetails(status)` that just supports the types from https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto . A question is what we should do with protobuf objects we don't recognize - I guess we could just return them in their `Any` form.

docs/src/main/paradox/server/details.md
```diff
@@ -44,43 +44,14 @@ Java
 :   @@snip[ExceptionGreeterServiceImpl](/interop-tests/src/test/java/example/myapp/helloworld/grpc/ExceptionGreeterServiceImpl.java) { #streaming }
 
 ## Rich error model
-Beyond status codes you can also use the [Rich error model](https://www.grpc.io/docs/guides/error/#richer-error-model). Currently there is no native support for this concept in Akka gRPC. However you can use the following manual approach.
-
-Add the following dependency to receive required classes (that are based on the [common protobuf](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto)):
```
raboof, line 49: I think it was useful to include a reference to https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto in the docs - sure, it's linked from https://www.grpc.io/docs/guides/error/#richer-error-model as well but I've found myself searching for it in the past ;)

raboof (commented):
I like the direction the server-side is going in, that could probably be finished fairly easily with some tweaks.

I'm not so sure about the client side, I think that needs some more thinking about what the API should look like.

runtime/src/main/scala/akka/grpc/AkkaGrpcException.scala
johanandren: If we keep both I think it would be good with a name on this new one that more clearly expresses what it is.

sebastian-alfers: I am looking into your suggestions next 👍

runtime/src/main/scala/akka/grpc/AkkaGrpcException.scala
sebastian-alfers: This are some good ideas, especially the last one ;)

docs/src/main/paradox/server/details.md
sebastian-alfers: Yes we can keep it, especially since it is used in the spec

runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
```diff
@@ -25,12 +56,6 @@ class GrpcServiceException(val status: Status, val metadata: Metadata) extends R
     this(status, metadata.asScala)
   }
 
-  /**
-   * Java API: The response status.
-   */
-  def getStatus: Status =
```
sebastian-alfers, line 31: This is now inherited from `StatusRuntimeException `

runtime/src/main/scala/akka/grpc/AkkaGrpcException.scala
sebastian-alfers: I managed to unify them. The only way to make this compatible to the "manual" rich error approach is to `GrpcServiceException` extend `StatusRuntimeException` capturing the raw `io.grpc.Metadata`.

interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorModelNativeTest.java
sebastian-alfers: I am still looking into this.

sebastian-alfers:
Will look at the mima issues soon.

interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorNativeImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    @Override
+    public CompletionStage<HelloReply> sayHello(HelloRequest in) {
+
+        ArrayList<scalapb.GeneratedMessage> ar = new ArrayList<>();
+        ar.add(LocalizedMessage.of("EN", "The password!"));
+
+        GrpcServiceException exception = GrpcServiceException.apply(
+                Code.INVALID_ARGUMENT,
+                "What is wrong?",
+                JavaConverters.asScalaBuffer(ar).toSeq()
```
raboof, line 32: I'm sure we could do a nicer Java API, but perhaps we can leave that for a separate PR

raboof (approved):

raboof:
as discussed on slack: I split out the client-side changes from this PR, as I think the server-side changes are pretty much good to be merged. We can restore a PR for the client-side changes later.

WDYT? Anything I might have missed?

johanandren (approved):

runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
raboof: That's source compatible but not binary compatibly - but only users that called 'getStatus' would be impacted by that, and this class is marked `ApiMayChange`, so that indeed looks OK to me.

sebastian-alfers:
Looks, good, the `// #rich_error_model_unary` comment in the scala test is not used anymore I think, but does not matter much if we keep it. Lets merge it :)

PR #1544, opened 2022-01-07
Documentation fix for round robin

Marcus-Rosti:
The docs in one place says round-robin but the reference and code examples say round_robin


Marcus-Rosti:
Per: https://github.com/akka/akka-grpc/blob/e6f0402e4eab6ab87bba970cb0c4674b881f13cf/runtime/src/main/resources/reference.conf#L26

johanandren (approved):

Issue #1465, opened 2021-10-04
Review and improve test coverage of grpc-web


jrudolph:
As observed in https://github.com/akka/akka-grpc/issues/1459, I introduced a regression in grpc-web that wasn't covered by tests (but which a simple smoke test should have uncovered).

UnsolvedCypher:
@jrudolph I have run into this issue as well. If you pull the quickstart Dockers [here](https://grpc.io/docs/platforms/web/quickstart/) then you can run `curl 'http://localhost:8080/grpc.gateway.testing.EchoService/Echo' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:96.0) Gecko/20100101 Firefox/96.0' -H 'Accept: application/grpc-web-text' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate' -H 'custom-header-1: value1' -H 'Content-Type: application/grpc-web-text' -H 'X-User-Agent: grpc-web-javascript/0.1' -H 'X-Grpc-Web: 1' -H 'Origin: http://localhost:8081' -H 'Connection: keep-alive' -H 'Referer: http://localhost:8081/' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data-raw 'AAAAAAYKBGFvZXU='` to see what a proper response looks like.

There is a Scala example of what is needed to respond correctly here: https://github.com/improbable-eng/grpc-web/issues/194#issuecomment-970503606 . If there's any other information I can provide, please let me know!

jrudolph:
Thanks, @UnsolvedCypher. I ran your instructions and got this response:

```
λ base64 -d | hexdump -C
[...]

PR #1552, opened 2022-01-18
Fix grpc-web trailer rendering for fast path

jrudolph:
Turns out we lost trailer rendering in the performance optimization for strict entities. This is an attempt at adding it back. Would still be good to add some tests. I currently don't have time to develop this further but maybe we can prepare the fix for the next release.

Refs #1465

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
@@ -87,7 +87,7 @@ object GrpcProtocol {
       messageEncoding: Codec,
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
-      /** A shortcut to encode a data frame directly into a ByteString */
+      /** A shortcut to encode a data frame directly into a Response */
       encodeDataToResponse: (ByteString, immutable.Seq[HttpHeader], Trailer) => HttpResponse,
       /** A Flow over a stream of Frame using this frame encoding */
       frameEncoder: Flow[Frame, ChunkStreamPart, NotUsed])
--- a/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
@@ -10,11 +10,12 @@ import akka.annotation.InternalStableApi
 import akka.grpc.GrpcProtocol.{ GrpcProtocolReader, GrpcProtocolWriter }
 import akka.grpc.internal.{ Codec, Codecs, GrpcProtocolNative, GrpcProtocolWeb, GrpcProtocolWebText }
 import akka.http.javadsl.{ model => jmodel }
-import akka.http.scaladsl.model.{ ContentType, HttpHeader }
+import akka.http.scaladsl.model.{ ContentType, HttpHeader, HttpResponse, Trailer }
 import akka.http.scaladsl.model.HttpEntity.ChunkStreamPart
 import akka.stream.scaladsl.Flow
 import akka.util.ByteString
 
+import scala.collection.immutable
 import scala.util.Try
 
 /**
@@ -87,7 +88,7 @@ object GrpcProtocol {
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
       /** A shortcut to encode a data frame directly into a ByteString */
-      encodeDataToFrameBytes: ByteString => ByteString,
+      encodeDataToResponse: (ByteString, immutable.Seq[HttpHeader], Trailer) => HttpResponse,
       /** A Flow over a stream of Frame using this frame encoding */
       frameEncoder: Flow[Frame, ChunkStreamPart, NotUsed])
 
--- a/runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
@@ -7,33 +7,50 @@ package akka.grpc.internal
 import akka.grpc.GrpcProtocol._
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.HttpEntity.{ Chunk, ChunkStreamPart }
-import akka.util.ByteString
+import akka.util.{ ByteString, ByteStringBuilder }
 import io.grpc.{ Status, StatusException }
+
+import scala.collection.immutable
 
 abstract class GrpcProtocolWebBase(subType: String) extends AbstractGrpcProtocol(subType) {
   protected def postEncode(frame: ByteString): ByteString
   protected def preDecode(frame: ByteString): ByteString
 
   override protected def writer(codec: Codec): GrpcProtocolWriter =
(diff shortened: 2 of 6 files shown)


runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
```diff
@@ -30,8 +32,22 @@ abstract class GrpcProtocolWebBase(subType: String) extends AbstractGrpcProtocol
...
+    HttpResponse(
+      status = StatusCodes.OK,
+      headers = headers,
+      entity = HttpEntity(contentType, encodeDataToFrameBytes(codec, data, trailer)),
+      protocol = HttpProtocols.`HTTP/1.1`)
+
+  private def encodeDataToFrameBytes(codec: Codec, data: ByteString, trailer: Trailer): ByteString = {
+    val dataFrame = AbstractGrpcProtocol.encodeFrameData(codec.compress(data), codec.isCompressed, isTrailer = false)
+    // poor man's header rendering
+    val trailerData = ByteString(trailer.headers.map { h => s"${h._1.toLowerCase}:${h._2}" }.mkString("\r\n"))
```
viktorklang, line 46: @jrudolph Seems really expensive?

wb14123:
Any plan to merge this in? This support is really important in my project. Otherwise my client will be broken because of the missing trailer.

raboof (approved):
@jrudolph I added the necessary MiMa filter, is there anything more you were planning to look into here because you marked it as 'Draft'?

jrudolph:
I reworked the code.

I tested against akka-grpc-quickstart-scala and looked at the output which now looks reasonable when queried like this:

```
$ cur  'http://localhost:8080/GreeterService/SayHello' -X POST -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:96.0) Gecko/20100101 Firefox/96.0' -H 'Accept: application/grpc-web-text' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate' -H 'custom-header-1: value1' -H 'Content-Type: application/grpc-web-text' -H 'X-User-Agent: grpc-web-javascript/0.1' -H 'X-Grpc-Web: 1' -H 'Origin: http://localhost:8081' -H 'Connection: keep-alive' -H 'Referer: http://localhost:8081/' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' --data-raw 'AAAAAAYKBGFvZXU=' | base64 -d |hexdump -C

00000000  00 00 00 00 0d 0a 0b 48  65 6c 6c 6f 2c 20 61 6f  |.......Hello, ao|
00000010  65 75 80 00 00 00 0f 67  72 70 63 2d 73 74 61 74  |eu.....grpc-stat|
00000020  75 73 3a 30 0d 0a                                 |us:0..|
00000026
```

jrudolph:
While with 2.1.3 the output is this, i.e. missing the trailer completely:

```
00000000  00 00 00 00 0d 0a 0b 48  65 6c 6c 6f 2c 20 61 6f  |.......Hello, ao|
00000010  65 75                                             |eu|
00000012
```

runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
jrudolph: Thanks, now slightly improved.

runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
```diff
@@ -42,11 +59,15 @@ abstract class GrpcProtocolWebBase(subType: String) extends AbstractGrpcProtocol
     }
   }
 
-  @inline
-  private final def encodeTrailer(trailer: Seq[HttpHeader]): ByteString =
-    ByteString(trailer.mkString("", "\r\n", "\r\n"))
+  private final def encodeTrailerHeaders(trailerHeaders: Iterator[(String, String)]): ByteString = {
```
jrudolph, line 62: More code is now reused between strict and streamed responses.

runtime/src/main/scala/akka/grpc/internal/GrpcProtocolWeb.scala
```diff
@@ -7,33 +7,50 @@ package akka.grpc.internal
...
+      AbstractGrpcProtocol.encodeFrameData(codec.compress(trailerData), codec.isCompressed, isTrailer = true)
+    postEncode(encodeFrameToBytes(codec, DataFrame(data)) ++ trailerFrame)
+  }
+
+  private def encodeFrameToBytes(codec: Codec, frame: Frame): ByteString =
+    frame match {
       case DataFrame(data) =>
         AbstractGrpcProtocol.encodeFrameData(codec.compress(data), codec.isCompressed, isTrailer = false)
       case TrailerFrame(trailer) =>
-        AbstractGrpcProtocol.encodeFrameData(encodeTrailer(trailer), codec.isCompressed, isTrailer = true)
```
jrudolph, line 29: This line might be another bug I introduced in the optimization: we potentially report compression but don't actually do compression.

jrudolph:
Would be nice if you could have another look at the implementation, @raboof, given that we have no tests.

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -87,7 +88,7 @@ object GrpcProtocol {
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
       /** A shortcut to encode a data frame directly into a ByteString */
```
raboof, line 90:
```suggestion
      /** A shortcut to encode a data frame directly into a Response */
```

raboof (approved):
Looks good, haven't tested it myself yet

runtime/src/main/scala/akka/grpc/GrpcProtocol.scala
```diff
@@ -87,7 +87,7 @@ object GrpcProtocol {
       messageEncoding: Codec,
       /** Encodes a frame as a part in a chunk stream. */
       encodeFrame: Frame => ChunkStreamPart,
-      /** A shortcut to encode a data frame directly into a ByteString */
+      /** A shortcut to encode a data frame directly into a Response */
       encodeDataToResponse: (ByteString, immutable.Seq[HttpHeader], Trailer) => HttpResponse,
```
viktorklang, line 91: @jrudolph I think Trailer could get some optimizations as well—having a cached instance for an empty Trailer, and avoiding creating a collection of 1 element and then immediately creating a new collection via map: https://github.com/akka/akka-http/blob/main/akka-http-core/src/main/scala/akka/http/scaladsl/model/Trailer.scala#L33

jrudolph: For akka-grpc we are already using a cached instance here: https://github.com/jrudolph/akka-grpc/blob/efda1b1ff4a8fd9849be66a5cd8c66369f6027af/runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala#L32. For akka-http itself, you could also just omit the trailer when not needed.

Issue #1578, opened 2022-02-18
Double promise completion on shutdown


raboof:
Probably a matter of changing the `failure` to a `tryFailure`, but let's double-check.

```
java.lang.IllegalStateException: Promise already completed.
	at scala.concurrent.Promise.complete(Promise.scala:57)
	at scala.concurrent.Promise.complete$(Promise.scala:56)
	at scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:104)
	at scala.concurrent.Promise.failure(Promise.scala:109)
	at scala.concurrent.Promise.failure$(Promise.scala:109)
	at scala.concurrent.impl.Promise$DefaultPromise.failure(Promise.scala:104)
	at akka.grpc.internal.AkkaNettyGrpcClientGraphStage$$anon$1.postStop(AkkaNettyGrpcClientGraphStage.scala:175)
	at akka.stream.impl.fusing.GraphInterpreter.finalizeStage(GraphInterpreter.scala:600)
	at akka.stream.impl.fusing.GraphInterpreter.afterStageHasRun(GraphInterpreter.scala:580)
	at akka.stream.impl.fusing.GraphInterpreter.execute(GraphInterpreter.scala:394)
	at akka.stream.impl.fusing.GraphInterpreterShell.runBatch(ActorGraphInterpreter.scala:650)
	at akka.stream.impl.fusing.GraphInterpreterShell$AsyncInput.execute(ActorGraphInterpreter.scala:521)
	at akka.stream.impl.fusing.GraphInterpreterShell.processEvent(ActorGraphInterpreter.scala:625)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$processEvent(ActorGraphInterpreter.scala:800)
	at akka.stream.impl.fusing.ActorGraphInterpreter.akka$stream$impl$fusing$ActorGraphInterpreter$$shortCircuitBatch(ActorGraphInterpreter.scala:787)
	at akka.stream.impl.fusing.ActorGraphInterpreter$$anonfun$receive$1.applyOrElse(ActorGraphInterpreter.scala:819)
	at akka.actor.Actor.aroundReceive(Actor.scala:537)
	at akka.actor.Actor.aroundReceive$(Actor.scala:535)
	at akka.stream.impl.fusing.ActorGraphInterpreter.aroundReceive(ActorGraphInterpreter.scala:716)
[...]

PR #1580, opened 2022-02-21
fix: don't throw if the promise got completed in the meanwhile

janory:
References #1578.

Switching `failure` to `tryFailure`, because it can happen that the promise gets completed in the meanwhile, in this case we shouldn't throw an exception.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
@@ -172,7 +172,7 @@ private final class AkkaNettyGrpcClientGraphStage[I, O](
           call = null
         }
         if (!matVal.isCompleted) {
-          matVal.failure(new AbruptStageTerminationException(this))
+          matVal.tryFailure(new AbruptStageTerminationException(this))
         }
       }
 


runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
```diff
@@ -172,7 +172,7 @@ private final class AkkaNettyGrpcClientGraphStage[I, O](
           call = null
         }
         if (!matVal.isCompleted) {
-          matVal.failure(new AbruptStageTerminationException(this))
+          matVal.tryFailure(new AbruptStageTerminationException(this))
```
raboof, line 175: I guess we can remove the `matVal.isCompleted` check now, then?

raboof (approved):
looks good, one idea to take this further

runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
janory: Hmm.. 🤔 
Does it make sense to try if we know it's completed?

On the other hand removing the check would reduce the LOC, so I can do that if you want. 👍 
Should I?

raboof: A matter of taste I suppose - let's leave it like this :+1:

Issue #1461, opened 2021-09-30
Support for GRPC HTTP/JSON transcoding (server)


aklikic:
### Short description

Adding support for [Transcoding HTTP/JSON to gRPC](https://cloud.google.com/endpoints/docs/grpc/transcoding) to Akka GRPC server.

### Details

GRPC Protobuf descriptor supports HTTP/JSON transcoding definitions using [Transcoding HTTP/JSON to gRPC](https://cloud.google.com/endpoints/docs/grpc/transcoding) annotations.
By supporting this in Akka GRPC `server`, developers would benefit from the out-of-the-box support for exposing HTTP/REST server endpoints without needing to have any additional moving parts to achieve it (for example like using [Envoy](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/grpc_json_transcoder_filter)).
This would help in use cases where inter-service communication is migrating from HTTP/REST to GRPC and "legacy" consumers still depend on HTTP/REST.

ihostage:
You read my mind, Alan @aklikic 😄 
Lagom, as you know of course, already has this opportunity([lagom-java-grpc-example](https://developer.lightbend.com/guides/lagom-java-grpc-example/)). And me too, frequently think about support REST/JSON in `akka-grpc` out-of-the-box. Then it will be cool that any Akka users could do that without additional abstract layers and frameworks.

raboof:
> Akka-HTTP/Akka-gRPC also supports serving multiple services/routes from a single HTTP server:port if you maintain the REST/JSON and the gRPC descriptors manually

(jup, this is documented at https://doc.akka.io/docs/akka-grpc/current/server/akka-http.html )

ihostage:
Thank you for your comment Ignasi @ignasi35 👍 
I understood that the Lagom feature isn't a zero cost, but I thought that this cost is adequate for support one business implementation with two (REST/gRPC) transport interfaces 🙂  
But I missed 🤦‍♂️  that `akka-grpc` also has the same feature, thank you and Arnout @raboof for your comments. ✋ 

Still, I warmly support the idea of Alan @aklikic and think that o-o-t-b `Transcoding HTTP/JSON to gRPC` and the minimalization cost of this feature will be a great opportunity for Akka users.

viktorklang:
@aklikic This could be a good start for you: https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala

[...]

PR #1583, opened 2022-02-24
feat: http transcoding grpc

WayneWang12:
References #1461.

Most code of this commit are from file https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala

Just do some change to make it may work in akka-grpc.

### DIFF
--- a/codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
@@ -128,12 +128,17 @@ object @{serviceName}Handler {
           .recoverWith(GrpcExceptionHandler.from(eHandler(system.classicSystem))(system, writer))
       ).getOrElse(unsupportedMediaType)
 
-      Function.unlift((req: model.HttpRequest) => req.uri.path match {
+
+
+      val httpHandler = akka.grpc.HttpApi.serve(@{service.name}.descriptor, handle _)
+
+      val grpcHandler = Function.unlift((req: model.HttpRequest) => req.uri.path match {
         case model.Uri.Path.Slash(model.Uri.Path.Segment(`prefix`, model.Uri.Path.Slash(model.Uri.Path.Segment(method, model.Uri.Path.Empty)))) =>
           Some(handle(spi.onRequest(prefix, method, req), method))
         case _ =>
           None
       })
+      httpHandler orElse grpcHandler
     }
   }
 }
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -79,6 +79,7 @@ object Dependencies {
 
   object Protobuf {
     val protobufJava = "com.google.protobuf" % "protobuf-java" % Versions.googleProtobuf
+    val protobufJavaUtil = "com.google.protobuf" % "protobuf-java-util" % Versions.googleProtobuf
     val googleCommonProtos = "com.google.protobuf" % "protobuf-java" % Versions.googleProtobuf % "protobuf"
   }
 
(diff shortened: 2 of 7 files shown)


lightbend-cla-validator:

WayneWang12:
I've only implement scala mode. I'm not quite familiar with Java.

Also, I think we can make grpc client works for json format http request. That will make akka-grpc usable even we only use protobuf to describe service protocols.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -128,12 +128,24 @@ object @{serviceName}Handler {
...
+     @for(method <- service.methods if !method.inputStreaming) {
+      lazy val @{method.name}Handler = @if(method.outputStreaming) {
+             akka.grpc.HttpApi.getSourceHandler@{if(powerApis) "PowerApi" else ""}("@{method.name}", @{service.name}.descriptor, implementation.@{method.nameSafe})
+               } else {
+             akka.grpc.HttpApi.getHttpFutureHandler@{if(powerApis) "PowerApi" else ""}("@{method.name}", @{service.name}.descriptor, implementation.@{method.nameSafe})
+               }
+     }
+
+      val httpApiHandlers = (List.empty[PartialFunction[model.HttpRequest, scala.concurrent.Future[model.HttpResponse]]]  @for(method <- service.methods if !method.inputStreaming){ ++ @{method.name}Handler})
+        .foldLeft(PartialFunction.empty[model.HttpRequest, scala.concurrent.Future[model.HttpResponse]])((acc, p) => acc.orElse(p))
```
viktorklang, line 140: It's very likely that you'll want to use a different method of partial function composition (I remember that I had to do something more custom to make sure that it was well-behaved in the face of many endpoints when I did the Cloudstate version). Otherwise you'll have to do a match linearly against all possible cases for each request.

WayneWang12: Yes, you are right. I've changed my implementation. I can get grpc handler and use it, so what I need to do is transforming json http request to grpc request and handle it with grpc handler, then transform grpc response to json http response. Not too efficient, but it might work.

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -128,12 +128,17 @@ object @{serviceName}Handler {
           .recoverWith(GrpcExceptionHandler.from(eHandler(system.classicSystem))(system, writer))
       ).getOrElse(unsupportedMediaType)
 
-      Function.unlift((req: model.HttpRequest) => req.uri.path match {
+
+
+      val httpHandler = akka.grpc.HttpApi.serve(@{service.name}.descriptor, handle _)
```
raboof, line 133: This effectively elevates `HttpApi.serve` to stable API: if a project uses code generated by this version, and updates the Akka gRPC runtime library, we'll have to make sure these new versions of the runtime library still provide that `serve` function with the same signature and behavior. I'm not sure I'm comfortable with that yet.

I wonder if it would be possible to use this `HttpApi` from the main client code, so the user can optionally enable it, and we can mark it `ApiMayChange` to leave room for evolving the API further later?

codegen/src/main/twirl/templates/ScalaServer/Handler.scala.txt
```diff
@@ -128,12 +128,17 @@ object @{serviceName}Handler {
...
+
+      val grpcHandler = Function.unlift((req: model.HttpRequest) => req.uri.path match {
         case model.Uri.Path.Slash(model.Uri.Path.Segment(`prefix`, model.Uri.Path.Slash(model.Uri.Path.Segment(method, model.Uri.Path.Empty)))) =>
           Some(handle(spi.onRequest(prefix, method, req), method))
         case _ =>
           None
       })
+      httpHandler orElse grpcHandler
```
raboof, line 141: This also seems potentially tricky: some high-throughput use cases might only care about the gRPC protocol, and hard-coding the HTTP transcoding support here might impact that hot path.

raboof (commented):
This is already much less intrusive, which is great! I wonder if we could take that even further?

johanandren:
Closing this since continued in #1834

Issue #1588, opened 2022-03-08
Feature: Sharing connection between service clients


drmontgomery:
Most gRPC client libraries allow multiple service clients to share a single HTTP connection to a server. This is typically presented as a separate "channel" abstraction, which used to initialize the service client. (For example, see https://grpc.io/docs/languages/java/basics/#client)

In akka-grpc, on the other hand, each service client creates its own connection to the server (via creating a ClientState, which creates an InternalChannel). While this works, it does mean that an akka-grpc client that uses multiple services will use many more HTTP connections than clients using other gRPC implementations. In some contexts (e.g. load testing), this difference can be significant.

Would it be possible to add support for reusing a single HTTP connection with multiple service clients? I think exposing a channel abstraction, like other gRPC libraries, would be best, since it allows the library user to control how connections are shared between services.

drmontgomery:
Added pull request as a more concrete example of what I'm proposing.

raboof:
The feature has been merged, but we'd stll like to add a code example to the docs - perhaps let's leave the issue open for that?

PR #1589, opened 2022-03-08
Add support for service client connection reuse.

drmontgomery:
Refactored ClientState into GrpcChannel.
Modified generated service clients to use GrpcChannel instead of GrpcClientSettings, with an apply() shim to create a GrpcChannel from GrpcClientSettings.
Added GrpChannelSpec to confirm that connection is being reused.

References #1588


raboof (commented):
This seems like a great idea, and the PR looks very good already! I think it needs some care to make sure binary compatibility is ensured, though - particularly between the generated code and the runtime library.

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -28,26 +29,27 @@ import akka.grpc.AkkaGrpcGenerated;
...
+  }
+
+  public static final @{service.name}Client create(GrpcChannel channel, ClassicActorSystemProvider sys) {
+    return new Default@{service.name}Client(channel, sys);
   }
 
   @@AkkaGrpcGenerated
   protected final static class Default@{service.name}Client extends @{service.name}Client {
 
-      private final ClientState clientState;
```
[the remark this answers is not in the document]
drmontgomery: Thank you for the feedback!

As you suggested, I recreated ClientState as a wrapper around a GrpcChannel. I think it properly supports the constructor and methods that are used by older generated clients.

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
```diff
@@ -6,54 +6,37 @@ package akka.grpc.internal
...
-import scala.concurrent.Future
+import akka.grpc.{ GrpcChannel, GrpcClientSettings }
 
 /**
  * INTERNAL API
  *
- * Client utilities taking care of Channel reconnection and Channel lifecycle in general.
+ * Deprecated: This class wraps a GrpcChannel for compatibility with clients generated by previous versions.
  */
+@deprecated("Kept for binary compatibility between generated code and runtime", "akka-grpc 2.1.4")
```
drmontgomery, line 22: Is a deprecation annotation appropriate here?
Should version number be 2.1.4 (next patch release) or 2.2.0 (next minor release)?
Should version number include library name? I see both styles being used elsewhere in the project.

drmontgomery (commented):
I think the remaining MiMa errors about ClientState can be safely ignored, since they're about methods that were not used by the generated code.

I'm less certain about the change to the DefaultServerReflectionClient constructor. The documentation would steer users toward using `ServerReflectionClient.apply` over `new DefaultServerReflectionClient`, but the constructor was public. Should we add a deprecated constructor that takes GrpcClientSettings to the generated default client?

runtime/src/main/scala/akka/grpc/internal/ClientState.scala
raboof: > Is a deprecation annotation appropriate here?

Yes, I think so :+1: 

> Should version number be 2.1.4 (next patch release) or 2.2.0 (next minor release)?

I think the next patch release would be best

> Should version number include library name? I see both styles being used elsewhere in the project.

I see https://www.scala-lang.org/api/current/scala/deprecated.html promotes adding the library name, but TBH I don't think it's necessary - I'm OK with either

drmontgomery:
After thinking about MiMa warnings a bit more, I added access modifiers to GrpcChannel and ChannelUtils to limit unintended exposure of constructors and methods. I think we could make the entire ChannelUtils object private to the akka package, but I wasn't sure if that is within scope for this PR.

I also noticed that ChannelUtils.closeCS was no longer used, so I removed it.

runtime/src/main/scala/akka/grpc/GrpcChannel.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.compat.java8.FutureConverters._
+import scala.concurrent.Future
+
+import akka.Done
+import akka.actor.ClassicActorSystemProvider
+import akka.annotation.InternalApi
+import akka.grpc.internal.{ ChannelUtils, InternalChannel }
+import akka.grpc.scaladsl.Grpc
+
+final class GrpcChannel private (val settings: GrpcClientSettings, @InternalApi val internalChannel: InternalChannel)(
```
raboof, line 18: `settings` is used by the generated code, so could you mark it `@InternalStableApi`?

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -89,16 +91,18 @@ final class Default@{service.name}Client(settings: GrpcClientSettings)(implicit
     @{method.nameSafe}().invoke(in)
   }
 
-  override def close(): scala.concurrent.Future[akka.Done] = clientState.close()
-  override def closed: scala.concurrent.Future[akka.Done] = clientState.closed()
+  override def close(): scala.concurrent.Future[akka.Done] = channel.close()
```
raboof, line 94: Hmm, this seems rather tricky - so now closing one client will close the channel, potentially also making other clients unusable. Is that what we want? If so, perhaps we should document it prominently?

raboof (commented):
Shaping up nicely! Can you add the mima exclusions for changed internal API's that you've verified are not used from old generated code?

runtime/src/main/scala/akka/grpc/GrpcChannel.scala
drmontgomery: Good catch! When I wrote that, I was thinking that it could be convenient for library users if they could access the settings associated with a channel. Looking through the code, however, it looks like settings are generally only consumed and not exposed.

As you suggested, I added the InternalStableApi annotation to settings, and updated internalChannel to use it as well (instead of InternalApi).

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
drmontgomery: I agree. This is tricky and potentially surprising. Unfortunately, I think the trickiness is an unavoidable consequence of simultaneously supporting both client-owned and externally-owned channels within the same generated service client interface.

At least initially, I think reusing channels will be a power-user feature, like the client PowerApi. As such, I hope that callers who opt-in to shared channels will also understand the channel itself should be closed, rather than the clients that share it, but you're right that this should be more clearly called out in the documentation. I can add a warning to the API docs for the close methods, as well as a section about reusing channels in client/details.md.

We could also add a flag to track whether the service client created the channel, and throw an exception if close is called with an externally-owned channel. This is still a potentially surprising run-time error, but the exception would make the error immediately obvious. Do you think this would be worthwhile?

raboof: > I can add a warning to the API docs for the close methods, as well as a section about reusing channels in client/details.md.

Sounds great!

> We could also add a flag to track whether the service client created the channel, and throw an exception if close is called with an externally-owned channel. This is still a potentially surprising run-time error, but the exception would make the error immediately obvious. Do you think this would be worthwhile?

I think that would be valuable, but for now just having clear docs should be sufficient as well

codegen/src/main/twirl/templates/ScalaClient/Client.scala.txt
```diff
@@ -43,62 +44,58 @@ trait @{service.name}Client extends @{service.name} with @{service.name}ClientPo
...
-}
-
-@@AkkaGrpcGenerated
-final class Default@{service.name}Client(settings: GrpcClientSettings)(implicit sys: ClassicActorSystemProvider) extends @{service.name}Client {
-  import @{service.name}.MethodDescriptors._
+    new Default@{service.name}Client(GrpcChannel(settings), isChannelOwned = true)
+  def apply(channel: GrpcChannel)(implicit sys: ClassicActorSystemProvider): @{service.name}Client =
+    new Default@{service.name}Client(channel, isChannelOwned = false)
+
+  private class Default@{service.name}Client(channel: GrpcChannel, isChannelOwned: Boolean)(implicit sys: ClassicActorSystemProvider) extends @{service.name}Client {
```
drmontgomery, line 51: Making the default service client private to the service object feels like an improvement to me. The default client was already marked final to prevent subclassing, and user should be using the apply() method on the service client object rather than creating default clients directly. Making the class private to the service client just enforces these rules more strongly, and brings the generated scala client into alignment with the generated java client.


That said, I'm still calibrating my application of the [Boy Scout Rule](https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html). Let me know if this goes too far.

raboof:
I'm sorry to leave this hanging, I hope to return to it later.

In the mean time, it seems like the CI jobs weren't correctly triggered - could you update/push a commit to get those going again?

drmontgomery:
I added some documentation to API docs and client details page.

Ideally, I'd also like to add a code snippet in the details that shows how to  create a GrpcChannel and use it to create multiple generated clients, but I wasn't sure about the best way to accomplish this.

I initially created a file similar to runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly, but I ran into a snag when I tried to define some example protobuf services for it. I created a file to define the service (runtime/src/test/protobuf/example.proto), but the ReflectiveCodeGen plug-in didn't seem to generate source for it.

I also considered annotating the relevant section of the GrpcChannelSpec test and referencing that from the documentation. This is less appealing, since the test uses two instances of the same service, which might be confusing to someone reading the documentation, and I'd rather not muddy the test by defining and implementing a second service.

Any suggestions here?

raboof:
> I added some documentation to API docs and client details page.

Looks great!

> Ideally, I'd also like to add a code snippet in the details that shows how to create a GrpcChannel and use it to create multiple generated clients, but I wasn't sure about the best way to accomplish this.
>
> I initially created a file similar to runtime/src/test/scala/docs/akka/grpc/client/GrpcClientSettingsCompileOnly, but I ran into a snag when I tried to define some example protobuf services for it. I created a file to define the service (runtime/src/test/protobuf/example.proto), but the ReflectiveCodeGen plug-in didn't seem to generate source for it.
>
> I also considered annotating the relevant section of the GrpcChannelSpec test and referencing that from the documentation. This is less appealing, since the test uses two instances of the same service, which might be confusing to someone reading the documentation, and I'd rather not muddy the test by defining and implementing a second service.
>
> Any suggestions here?

I agree this would be nice, and that it would be best to show the feature with two different services. Not sure why `ReflectiveCodeGen` wouldn't be picking up your proto, perhaps it doesn't look at the test scope? Moving the sample code to the akka-grpc-integration-tests project might help?

I think the feature is already very attractive, though, so I intend to merge this PR as-is - we can add a code example in a later PR.

Thanks a lot!

raboof (approved):

PR #1614, opened 2022-05-13
Document the correct version of the default protoc used.

orendain:
Documents the correct version of the default `protoc` used by parameterizing the docs with an additional `protobuf.version` property.

**Sample of doc generated:**
image


raboof (commented):
I'm not entirely sure this is accurate: indeed this will list the correct version of the protobuf-java dependency, but I'm not sure if this also determines the version of the `protoc` compiler, which is invoked indirectly. Perhaps we just shouldn't mention a version here, if we can't easily keep it up-to-date?

raboof:
superseded by #1664 - thanks for pointing out the issue!

Issue #262, opened 2018-06-28
keep grpc-stub versions in sync


raboof:
We depend on grpc-java directly, but there's also `scalapb.compiler.Version.grpcJavaVersion`.

We might want to make sure we keep those consistent?

PR #1615, opened 2022-05-13
Sync gRPC versions between the Gradle plugin and the rest of the project.

orendain:
Resolves https://github.com/akka/akka-grpc/issues/262.

Originally, the goal was for the Gradle plugin to not depend on `io.groc:grpc-stub:1.32.1`, due to it being vulnerable to [CVE-2020-8908](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908).

The approach was to keep the Gradle plugin's version in sync with the rest of the project, naturally resulting in an updated version of its gRPC dependencies.

Changes:
- Update gRPC version used in the Gradle plugin from `1.32.1` to `1.46.0`.
- Add `AkkaGrpcPluginException.groovy` as an artifact to be version-checked.


raboof (approved):
great, thanks!

PR #1643, opened 2022-08-02
fix: allow messages with the same name from different packages

### LINKED ISSUE #1625: Bad naming for message classes of the same name but from different packages yielding collisions
https://github.com/akka/akka-grpc/blob/756a185d129ff8dc988b82679b759440944dbe53/codegen/src/main/twirl/templates/JavaCommon/ApiInterface.scala.txt#L30

is already defined as value 
```
 val SendOrderRequestSerializer = new ScalapbProtobufSerializer(...
```

jroper:
Fixes #1625

This allows messages with the same name but different packages to be used by the same service by prefixing the messages generated serializer names with the underscore separated package name if they came from if they come frome a different package to the package the service is declared in.


raboof (approved):
Neat!

jroper:
Could I get an akka-grpc release with this? I'm somewhat blocked on a customer issue until I can get this fix (I can work around it, but that will involve making breaking changes to a public API later when we want to move to the fix).

PR #1645, opened 2022-08-03
Use fully qualified name for Metadata in PowerApi trait

### LINKED ISSUE #1642: Metadata message type in PowerAPI can cause a name collision
### Versions used 

Akka gRPC `1.1.1`, but this is also present in `main` and on `2.1.3`/`2.1.4`.

Akka version: 
Unimportant, but `2.6`.

### Expected Behavior

1. Create a proto file with `Metadata` as a message name.
2. Compile an akka-grpc server in power mode (PowerApi)
3. Everything should compile normally.

### Actual Behavior

1. Create a proto file with `Metadata` as a message name.
2. When you compile akka-grpc server in power mode, an internal proto file collides:
https://github.com/akka/akka-grpc/blob/main/codegen/src/main/twirl/templates/ScalaServer/PowerApiTrait.scala.txt#L19

We were able to verify that this can be fixed by foregoing the import [here](https://github.com/akka/akka-grpc/blob/38bd8348dd10bce0c167e2444e86357911619583/codegen/src/main/twirl/templates/ScalaServer/PowerApiTrait.scala.txt#L10), and instead using the full path `akka.grpc.scaladsl.Metadata` on line 19.  E.g.,
```
def @{method.nameSafe}(in: @method.parameterType, metadata: akka.grpc.scaladsl.Metadata)
```

dwhjames:
fixes #1642

### DIFF
--- a/codegen/src/main/twirl/templates/ScalaServer/PowerApiTrait.scala.txt
+++ b/codegen/src/main/twirl/templates/ScalaServer/PowerApiTrait.scala.txt
@@ -7,7 +7,6 @@
 @akka.grpc.gen.Constants.DoNotEditComment
 package @service.packageName
 
-import akka.grpc.scaladsl.Metadata
 import akka.grpc.AkkaGrpcGenerated
 
 /*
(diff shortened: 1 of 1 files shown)


patriknw (approved):
LGTM

dwhjames:
@patriknw, thanks for the review. I think [this](https://github.com/akka/akka-grpc/runs/7894847037?check_suite_focus=true) is a spurious failure, or at least unrelated. Could you rerun that on my behalf?

raboof (approved):

PR #1647, opened 2022-08-03
TLS related settings should force useTls=true

dwhjames:
rather than
```scala
settings.withTls(true).withSslContext(context)
```
instead,
```scala
settings.withSslContext(context)
```
should suffice

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
@@ -212,15 +212,17 @@ final class GrpcClientSettings private (
   def withCallCredentials(value: CallCredentials): GrpcClientSettings = copy(callCredentials = Option(value))
   def withOverrideAuthority(value: String): GrpcClientSettings = copy(overrideAuthority = Option(value))
   def withSslProvider(sslProvider: SslProvider): GrpcClientSettings =
-    copy(useTls = true, sslProvider = Option(sslProvider))
+    Option(sslProvider).fold(this)(sslProvider => copy(useTls = true, sslProvider = Some(sslProvider)))
 
   /**
    * Prefer using `withContextManager`: withSslContext forces the ssl-provider 'jdk', which is known
    * not to work on JDK 1.8.0_252.
    */
-  def withSslContext(sslContext: SSLContext): GrpcClientSettings = copy(useTls = true, sslContext = Option(sslContext))
+  def withSslContext(sslContext: SSLContext): GrpcClientSettings =
+    Option(sslContext).fold(this)(sslContext => copy(useTls = true, sslContext = Option(sslContext)))
   def withTrustManager(trustManager: TrustManager): GrpcClientSettings =
-    copy(useTls = true, trustManager = Option(trustManager))
+    Option(trustManager).fold(this)(trustManager => copy(useTls = true, trustManager = Option(trustManager)))
+
   def withResolveTimeout(value: FiniteDuration): GrpcClientSettings = copy(resolveTimeout = value)
 
   /**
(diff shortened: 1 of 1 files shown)


raboof (commented):
I agree it's nice to be able to skip that 'redundant' line. On the other hand it makes the settings code slightly less consistent/predictable/simple. Also it looks like you can pass `null` to clear these fields - should that also reset the `useTls` flag?

dwhjames:
@raboof, thanks for the review!

> On the other hand it makes the settings code slightly less consistent/predictable/simple.

The builder pattern already has the limitation that the effects of a chain are sequential and later methods calls can conflict with or undo the effects of earlier method calls.

The main issue I’m seeking to address here is that `useTls` is `true` by default, but only because of the default config seeds that into the builder
https://github.com/akka/akka-grpc/blob/5c6e9c01d481342c047c9467f1c3b5e08b1f5cba/runtime/src/main/resources/reference.conf#L34

Without this change, it is possible to have a security incident where:
- `.withTls(true)` is not given (because it was assumed and previously observed to be true by default)
- some config clash in a `reference.conf` or `application.conf` overrides `akka.grpc.client."*".use-tls = false`
- despite `.withSslContext(context)`, the TLS remains _off_

Lexically, the code would appear to demand TLS, but the code would revert to no TLS.

With this change, the behavior can only be consistent with what appears in the code.

raboof:
Sure. I'm OK with the change if we find a sensible behavior when passing `null` to these three methods.

dwhjames:
> Sure. I'm OK with the change if we find a sensible behavior when passing `null` to these three methods.

I’m sorry, I missed the second half of your original comment.

> Also it looks like you can pass null to clear these fields - should that also reset the useTls flag?

---

How about

```scala
def withSslProvider(sslProvider: SslProvider): GrpcClientSettings = {
  Option(sslProvider).fold(this,  copy(useTls = true, sslProvider = Some(sslProvider)) 
}
```
?

raboof:
> How about
> 
> ```scala
> def withSslProvider(sslProvider: SslProvider): GrpcClientSettings = {
>   Option(sslProvider).fold(this,  copy(useTls = true, sslProvider = Some(sslProvider)) 
> }
> ```

OK, so this will enable `useTls` automatically when setting `sslProvider` to a value, but not disable it again when removing it again. A bit surprising but I can live with that.

dwhjames:
> OK, so this will enable `useTls` automatically when setting `sslProvider` to a value, but not disable it again when removing it again. A bit surprising but I can live with that.

It’s pretty hard to cater to the case of `null`. And the semantics here get awkward if more than one of the multiple TLS related config building methods are used.

For example,
```scala
builder
  .withTls(false)
  .withSslContext(sslContext)
  .withTrustManager(manager)
  .withSslContext(null)
```

Under the above proposal, this comes out to be `useTls = true`, `sslContext = Some(sslContext)`, `trustManager = Some(trustManager)`, and for the Netty Client (and hopefully shortly for the Akka Client https://github.com/akka/akka-grpc/pull/1652) this will evaluate to the sslContext being used in preference to the trust manager.

We’ve kinda reached the limit of what is sane to express with the builder pattern.
I’d say the `null` protection is reasonable to avoid polluting the config with null values, but using `null` to _unset_ config seems problematic.

patriknw (approved):
LGTM, I think the arguments to keep `useTls=true` are valid and important. If it's not perfect for null unset is less important and I assume much less used.

patriknw:
@dwhjames would you be able to run `scalafmtAll` here (or maybe just Test / compile would be enough)

dwhjames:
@patriknw, oops! I fixed the formatting in https://github.com/akka/akka-grpc/pull/1647/commits/38fb55350fc7e2cbfbf3d7ca733c3f9cdec43b7e

@raboof, thanks for the review. Feedback addressed in https://github.com/akka/akka-grpc/pull/1647/commits/aa19052c5e513261776d9429669c8581e94e7b9c

raboof (approved):

PR #1649, opened 2022-08-03
Clean up NettyClientUtils#createNettySslContext

dwhjames:
Inserting a JDK `javax.net.ssl.SSLContext` into Netty originated in commit
https://github.com/akka/akka-grpc/blob/aa05239c6cddcb20dfa0770e8e8e7649e3bbaaef/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala#L59-L82
in PR https://github.com/akka/akka-grpc/pull/266

It was removed in PR https://github.com/akka/akka-grpc/pull/964
to address Issue https://github.com/akka/akka-grpc/issues/946

It was returned in PR https://github.com/akka/akka-grpc/pull/979
to address Issue https://github.com/akka/akka-grpc/issues/978

---

Original comment was
```scala
    // FIXME: Create a JdkSslContext using a normal constructor. Need to work out sensible values for all args first.
    // In the meantime, use a Netty SslContextBuild to create a JdkSslContext, then use reflection to patch the
    // object's internal SSLContext. It's not pretty, but it gets something working for now.
```

---

This commit addresses the original `FIXME` comment, and avoids using
deprecated constructors on `io.grpc.netty.shaded.io.netty.handler.ssl.JdkSslContext`

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
@@ -172,23 +172,23 @@ object NettyClientUtils {
    */
   @InternalApi
   private def createNettySslContext(javaSslContext: SSLContext): SslContext = {
-    import io.grpc.netty.shaded.io.netty.handler.ssl._
-    import java.lang.reflect.Field
-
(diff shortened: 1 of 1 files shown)


raboof (approved):
good find!

Issue #1650, opened 2022-08-05
grpc-web-text throws Base64 exceptions when unmarshalling non-strict requests


drmontgomery:
Hi there!

My team has encountered a bug when sending large (> 1 megabyte) requests using the grpc-web-text protocol. When sending requests over an actual network (rather than loopback), akka-grpc intermittently throws IllegalArgumentException and IllegalStateException while unmarshalling the request.

After looking into this, I believe the problem is in the `AbstractGrpcProtocol.reader` method. The `flowAdapter` parameter allows each protocol implementation to provide an optional pre-decode function to manipulate the request bytes before they are decoded into gRPC data frames. The `reader` method incorporates this function into the frame decoding flow by mapping each element of the stream with it.

For the grpc-web-text protocol, each data frame is encoded with Base64, so `GrpcProtocolWebText` passes in a pre-decode function that calls `decodeBase64`.  Because Base64 encodes 3 8-bit bytes into 4 6-bit digits, simply mapping `decodeBase64` over the chunks of a Base64-encoded stream isn't guaranteed to work. If the chunk sizes do not align with the four byte encoded blocks, mapping calls to `decodeBase64` over the stream can produce incorrect results or throw errors.

Example (incorrect result):
```
scala> ByteString(0, 1, 2).encodeBase64.grouped(2).map(_.decodeBase64).toList
res0: List[akka.util.ByteString] = List(ByteString(0), ByteString(16))
```

Example (exception):
```
scala> ByteString(0, 1, 2).encodeBase64.grouped(3).map(_.decodeBase64).toList
java.lang.IllegalArgumentException: Input byte[] should at least have 2 bytes for base64 bytes
```

In our testing, large requests worked flawlessly when running locally. We only encountered issues when we tested against remote servers, probably because network delays were significant enough to cause the stream of request bytes to diverge from the 4-byte Base64 blocks. We were subsequently able to elicit the failure with local servers by using a modified HTTP client that sends the encoded body in blocks of five bytes with a delay between each block.

I'll submit a PR shortly with a test case that reproduces the problem and a proposed fix.

### Versions used 

Akka gRPC version: 2.1.2

We're using Akka gRPC 2.1.2, but I have confirmed the same issue exists in the main branch.

### Expected Behavior

Sending large requests using grpc-web-text should not generate errors.

### Actual Behavior

Sending large requests using grpc-web-text to remote servers causes Base64 decoding exceptions.

### Relevant logs

[...]

PR #1651, opened 2022-08-05
Fix base64 request decoding for grpc-web-text

drmontgomery:
Added DecodeBase64 flow that enforces 4-byte alignment.
Added preDecodeFlow parameter to AbstractGrpcProtocol.
Added tests.

References #1650

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/DecodeBase64.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/DecodeBase64.scala
@@ -1,0 +1,72 @@
+package akka.grpc.internal
+
+import akka.NotUsed
+import akka.annotation.InternalApi
+import akka.stream.scaladsl.Flow
+import akka.stream.stage.{ GraphStage, GraphStageLogic, InHandler, OutHandler }
+import akka.stream.{ Attributes, FlowShape, Inlet, Outlet }
+import akka.util.ByteString
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] object DecodeBase64 {
+  def apply(): Flow[ByteString, ByteString, NotUsed] =
+    Flow[ByteString].via(new DecodeBase64)
+}
+
+/**
+ * INTERNAL API
+ */
+@InternalApi private[akka] class DecodeBase64 extends GraphStage[FlowShape[ByteString, ByteString]] {
+  private val in = Inlet[ByteString]("DecodeBase64.in")
+  private val out = Outlet[ByteString]("DecodeBase64.out")
+
+  override def initialAttributes = Attributes.name("DecodeBase64")
+
+  final override val shape = FlowShape(in, out)
+
+  override def createLogic(inheritedAttributes: Attributes): GraphStageLogic =
+    new GraphStageLogic(shape) with InHandler with OutHandler {
+      private var buffer: ByteString = ByteString.empty
+
+      override def onPush(): Unit = {
+        buffer ++= grab(in)
+
+        val length = buffer.length
+        val decodeLength = length - length % 4
+
+        if (decodeLength > 0) {
+          val (decodeBytes, remaining) = buffer.splitAt(decodeLength)
+          push(out, decodeBytes.decodeBase64)
+          buffer = remaining
+        } else {
+          pull(in)
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        if (buffer.nonEmpty) {
+          if (isAvailable(out)) {
+            push(out, buffer.decodeBase64)
+            buffer = ByteString.empty
+          }
+        } else {
+          completeStage()
+        }
+      }
+
+      override def onPull(): Unit = {
+        if (isClosed(in)) {
+          if (buffer.nonEmpty) {
+            push(out, buffer.decodeBase64)
+            buffer = ByteString.empty
+          } else {
+            completeStage()
+          }
+        } else pull(in)
+      }
+
+      setHandlers(in, out, this)
+    }
+}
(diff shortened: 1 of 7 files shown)


drmontgomery:
Re binary compatibility: 
```
[error]  * abstract method preDecode(akka.util.ByteString)akka.util.ByteString in class akka.grpc.internal.GrpcProtocolWebBase does not have a correspondent in current version
[error]  * abstract method preDecodeFlow()akka.stream.scaladsl.Flow in class akka.grpc.internal.GrpcProtocolWebBase is present only in current version
[error]  * abstract method preDecodeStrict(akka.util.ByteString)akka.util.ByteString in class akka.grpc.internal.GrpcProtocolWebBase is present only in current version
[error]  * method preDecode(akka.util.ByteString)akka.util.ByteString in object akka.grpc.internal.GrpcProtocolWeb does not have a correspondent in current version
[error]  * method preDecode(akka.util.ByteString)akka.util.ByteString in object akka.grpc.internal.GrpcProtocolWebText does not have a correspondent in current version
[error]  * method reader(akka.grpc.internal.Codec,scala.Function2,scala.Function1)akka.grpc.GrpcProtocol#GrpcProtocolReader in object akka.grpc.internal.AbstractGrpcProtocol does not have a correspondent in current version
[error]  * static method preDecode(akka.util.ByteString)akka.util.ByteString in class akka.grpc.internal.GrpcProtocolWeb does not have a correspondent in current version
[error]  * static method preDecode(akka.util.ByteString)akka.util.ByteString in class akka.grpc.internal.GrpcProtocolWebText does not have a correspondent in current version
```
I think it's safe to filter all of these, since they're entirely contained within akka.grpc.internal (both definitions and potential references).

drmontgomery:
Re DecodeBase64:
The DecodeBase64 flow is effectively doing two things: enforcing the 4-byte chunk alignment and then passing the aligned chunks into ByteString.decodeBase64. It feels like the chunk alignment aspect could be useful and reusable on its own, perhaps as an addition to akka.stream.scaladsl.Framing.

Example:
```
def decodeBase64: Flow[ByteString, ByteString, NotUsed] = 
    Flow[ByteString]
        .via(Framing.aligned(4))
        .map(_.decodeBase64)
```

Is this worth pursuing, either for this PR or as a PR for akka streams?

runtime/src/main/scala/akka/grpc/internal/DecodeBase64.scala
```diff
@@ -0,0 +1,76 @@
...
+          buffer = remaining
+        } else {
+          pull(in)
+        }
+      }
+
+      override def onUpstreamFinish(): Unit = {
+        if (buffer.nonEmpty) {
+          if (isAvailable(out)) {
+            push(out, buffer.decodeBase64)
```
patriknw, line 55: Can we be sure that the right number of bytes have been pushed to the buffer, or is there a risk of some premature upstream finish? Maybe best to have the same `val decodeLength = length - length % 4` logic here?

patriknw: and same in the push from onPull

patriknw (commented):
looking good

runtime/src/main/scala/akka/grpc/internal/DecodeBase64.scala
drmontgomery: Padding in base64 is only required for correctness when concatenating encoded chunks. Most decoders will implicitly add padding for any missing bytes at the end of encoded data. I think this is an instance of the robustness principle ("be conservative in what you send, liberal in what you accept").

akka.util.ByteString.decodeBase64 automatically adds trailing padding:
```
scala> val data = akka.util.ByteString(List(0.toByte))
val data: akka.util.ByteString = ByteString(0)

scala> val encoded = data.encodeBase64
val encoded: akka.util.ByteString = ByteString(65, 65, 61, 61)

scala> val decoded = encoded.dropRight(2).decodeBase64
val decoded: akka.util.ByteString = ByteString(0)
```

I followed this same principle when implementing the DecodeBase64 flow. If the stream ends with an unaligned chunk of data, we decode it and pass it downstream before completing the stage. This should allow the akka-grpc servers to work with grpc-web client implementations that omit padding, without affecting correctness.

patriknw: I'm not sure we talk about the same problem. Here in onUpstreamFinish it tries to `decodeBase64` whatever is currently in the buffer. For example, `ByteString.fromString("a")` but only 1 byte has been pushed to the buffer and then the stream is finished. That can probably not happen for well behaved encoding upstreams but should this flow fail in such case or should it ignore the incomplete trailing part?

```
scala> ByteString.fromString("a").encodeBase64.take(1).decodeBase64.toString
java.lang.IllegalArgumentException: Input byte[] should at least have 2 bytes for base64 bytes
  at java.base/java.util.Base64$Decoder.decodedOutLength(Base64.java:709)
  at java.base/java.util.Base64$Decoder.decode(Base64.java:662)
  at akka.util.ByteString$ByteString1.decodeBase64(ByteString.scala:384)
```

raboof: > That can probably not happen for well behaved encoding upstreams but should this flow fail in such case or should it ignore the incomplete trailing part?

I don't think we should silently ignore/drop unaligned input.

For unaligned-but-parsable input (`ByteString.fromString("a").encodeBase64.take(2)`) I'm OK with either parsing it or throwing an error. The current implementation parses it, which seems fine by me.

For unaligned unparsable input (`ByteString.fromString("a").encodeBase64.take(1)`) we should fail the stream. Here we could either explicitly `fail`/`failStage` or just allow the exception to be thrown and have akka streams deal with it. I'm not sure I have a strong preference between those two, you?

patriknw: ok, sounds good to me. Then current code does the right thing already.

raboof (approved):

PR #1652, opened 2022-08-07
support SSLContext setting in Akka-Http backend

dwhjames:
mirroring the behavior of the Netty backend

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala
@@ -85,12 +85,14 @@ object AkkaHttpClientUtils {
       if (settings.useTls) {
         val connectionContext =
           ConnectionContext.httpsClient {
-            settings.trustManager match {
-              case None => SSLContext.getDefault
-              case Some(trustManager) =>
-                val sslContext: SSLContext = SSLContext.getInstance("TLS")
-                sslContext.init(Array[KeyManager](), Array[TrustManager](trustManager), new SecureRandom)
-                sslContext
+            settings.sslContext.getOrElse {
+              settings.trustManager match {
+                case None => SSLContext.getDefault
+                case Some(trustManager) =>
+                  val sslContext: SSLContext = SSLContext.getInstance("TLS")
+                  sslContext.init(Array[KeyManager](), Array[TrustManager](trustManager), new SecureRandom)
+                  sslContext
+              }
             }
           }
 


raboof (commented):
This means if `sslContext` is set the `trustManager` setting will be ignored, right?

I think that probably makes sense, but perhaps we should add a validation and throw an error when a user sets both?

dwhjames:
> This means if `sslContext` is set the `trustManager` setting will be ignored, right?
> 
> I think that probably makes sense, but perhaps we should add a validation and throw an error when a user sets both?

Sorry @raboof, I could do with paying closer attention to my notifications 😄 .

I was attempting to mirror the existing behavior

https://github.com/akka/akka-grpc/blob/5c0f851980d5e5ddf9c81c97f7f88bf7cb4ee803/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala#L59-L76

---

But I take your point that it would be a good safety feature in the config builder to throw an error on ‘ambiguous’ configuration, rather than taking some silent ordering of precedence.
However, my main concern there would be a conflict between loaded config (including defaults) from the resource classpath and then overriding in code, i.e., it might be legitimate to want to use the builder to _override_ rather than just accumulate. Which would mean that you’d need two methods in the builder: one that errors, and one that supplants all prior TLS related config.

Could we leave that for a follow on PR?

raboof:
> > This means if `sslContext` is set the `trustManager` setting will be ignored, right?
> > I think that probably makes sense, but perhaps we should add a validation and throw an error when a user sets both?
>
> I was attempting to mirror the existing behavior
> 
> https://github.com/akka/akka-grpc/blob/5c0f851980d5e5ddf9c81c97f7f88bf7cb4ee803/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala#L59-L76
> 
> But I take your point that it would be a good safety feature in the config builder to throw an error on ‘ambiguous’ configuration, rather than taking some silent ordering of precedence. (...) 
>
> Could we leave that for a follow on PR?

Yes, I guess that makes sense.

> my main concern there would be a conflict between loaded config (including defaults) from the resource classpath and then overriding in code, i.e., it might be legitimate to want to use the builder to _override_ rather than just accumulate. Which would mean that you’d need two methods in the builder: one that errors, and one that supplants all prior TLS related config.

I'm not sure I understand what you mean here. I agree it's legitimate to want to use the builder to override rather than just accumulate. When an existing `GrpcClientSettings` object already has a `trustManager` set, it might not be unreasonable to require the user to explicitly remove that `trustManager` before setting the `sslContext`?

raboof (approved):

raboof:
> Could we leave that for a follow on PR?

filed as issue #1674

PR #1654, opened 2022-08-08
release tags

patriknw:
I tagged `v2.1.5` from the releases page, but that didn't trigger any Publish. Trying this...


patriknw:
failed because it can't find released 2.1.5, merging anyway

PR #1657, opened 2022-08-22
fix link validator on tags

raboof:
The link validator would not work when the current HEAD is a tag
(i.e. https://github.com/akka/akka-grpc/runs/7946004526?check_suite_focus=true)
auto-detect the version in that case


ennru (approved):
Great.

Issue #1668, opened 2022-08-25
grpc-web not working for java api


sebastian-alfers:
As mentioned in [this](https://github.com/akka/akka-grpc/issues/1574) ticket, the java-api currently works different when it comes to the way how the request is validated based on request-type.

This causes the problem that grpc-web is not working for the java-api. [Here](https://github.com/akka/akka-grpc/blob/main/runtime/src/main/scala/akka/grpc/javadsl/WebHandler.scala#L74) a `grpc-web` is checked to contain the valid headers (either `GrpcProtocolWeb` or `GrpcProtocolWebText` currently). If that predicates matches, the request is forwarded to [this](https://github.com/akka/akka-grpc/blob/main/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala#L53) place where it is checked to be a request of type `GrpcProtocolNative`.

As a request obviously can not be both, it fails.

Issue #1574, opened 2022-02-14
Align concatOrNotFound between javadsl and scaladsl


sebastian-alfers:
Using the `scaladsl` version of `concatOrNotFound` allows to combine http-routes with grpc-routes.

However using the `javadsl` does fail http-requests at runtimee as of [this](https://github.com/akka/akka-grpc/blob/main/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala#L53) (`isGrpcRequest`) check which may be not expected.

PR #1669, opened 2022-08-25
remove isGrpcRequest check from the java api to align with scala api

sebastian-alfers:
Fixes https://github.com/akka/akka-grpc/issues/1668 and https://github.com/akka/akka-grpc/issues/1574

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
@@ -41,9 +41,7 @@ object ServiceHandler {
   /**
    * Creates a `HttpRequest` to `HttpResponse` handler for gRPC services that can be used in
    * for example `Http().bindAndHandleAsync` for the generated partial function handlers:
-   *  - The generated handler supports the `application/grpc` media type.
-   *  - If the request is for an invalid media type, then a _415: Unsupported Media Type_ response is produced.
-   *  - Otherwise if the request is not handled by one of the provided handlers, a _404: Not Found_ response is produced.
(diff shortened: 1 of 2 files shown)


johanandren (approved):
LGTM

johanandren:
(but let's wait to see what @raboof thinks before merging)

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
```diff
@@ -50,7 +49,7 @@ object ServiceHandler {
   def handler(handlers: JFunction[HttpRequest, CompletionStage[HttpResponse]]*)
       : JFunction[HttpRequest, CompletionStage[HttpResponse]] = {
     val servicesHandler = concat(handlers: _*)
-    (req: HttpRequest) => if (sServiceHandler.isGrpcRequest(req)) servicesHandler(req) else unsupportedMediaType
+    (req: HttpRequest) => servicesHandler(req)
```
raboof, line 52: The comment says "If the request is for an invalid media type, then a _415: Unsupported Media Type_ response is produced.". This probably only is true for paths matched by the handler? Might be something to double-check and perhaps make the comment more specific?

raboof (approved):
Indeed looks more consistent AFAICS. Would be nice to have more test coverage of the expected behavior in this area, to make sure this 'stays fixed', but if you have tested this for the scenario you care about I'm OK with merging as-is too.

runtime/src/main/scala/akka/grpc/javadsl/ServiceHandler.scala
sebastian-alfers: Good point, I changed the commend.

Issue #1667, opened 2022-08-24
Ability to handle rpc functions named close


benthecarman:
### description

I am trying to use a proto file that has a rpc function named close. This overlaps with the akka grpc function also called close so it cannot compile. Ideally it should rename to something else.

raboof:
Good observation!

I think we should probably change the generated method name. Eventually it would be great if we could allow users to configure various aspects of code generation on the method level, then they could resolve this naming clash themselves (#451). Until that time we could treat 'close' as a special case and rename it automatically - perhaps adding some suffix? Would you be interested in contributing this feature?

benthecarman:
> Would you be interested in contributing this feature?

I would love to but not too sure how to do it. Willing to provide a small bitcoin bounty for the feature :)

benthecarman:
I think I was able to do it here: https://github.com/akka/akka-grpc/pull/1670

raboof:
Awesome, thanks!

PR #1670, opened 2022-08-26
Update scala reserved words, handle already defined method names like `close`

benthecarman:
References: #1667

1st commit:
Updates the lists to the latest of reserved words for scala, also fixes the comment so it is pointed to a commit so the exact line isn't wrong if they change the file

2nd commit:
Adds `Method` suffix to a method's name if is a method that will already be defined by scala/akka

### DIFF
--- a/codegen/src/main/scala/akka/grpc/gen/scaladsl/Method.scala
+++ b/codegen/src/main/scala/akka/grpc/gen/scaladsl/Method.scala
@@ -80,7 +80,7 @@ object Method {
     messageType.scalaType.fullName
   }
 
-  // https://github.com/scalapb/ScalaPB/blob/master/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala#L1038
+  // https://github.com/scalapb/ScalaPB/blob/38845c0cf21173a2242a5d14ed48a7c33b981bae/compiler-plugin/src/main/scala/scalapb/compiler/DescriptorImplicits.scala#L1115
   private val ReservedWords = Set(
     "abstract",
     "case",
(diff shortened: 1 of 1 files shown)


lightbend-cla-validator:
Hi @benthecarman,

raboof (approved):
Nice, looks good!

I was going to recommend making the corresponding change on the Java codegen side as well, but it seems that's already less sophisticated than the Scala one, so we can leave that for a future PR.

benthecarman:
Is there any timeline for a release so I can start using this?

raboof:
You could already use a snapshot release from the sonatype snapshot repository (2.1.5-14-6e38aed5-SNAPSHOT or later, https://oss.sonatype.org/content/repositories/snapshots/com/lightbend/akka/grpc/akka-grpc-runtime_2.13/2.1.5-14-6e38aed5-SNAPSHOT/ ).

We don't have a specific release date planned but there's a number of interesting things merged since the last one, so we might do one in the next week(s)

benthecarman:
Is there a way to do snapshots for plugins?

raboof:
I think it should work the same as for regular dependencies: add `resolvers += Resolver.sonatypeRepo("releases")` to `project/plugins.sbt` and use the snapshot version - does that work for you?

benthecarman:
tried with all 3 of these and it didn't work
```
resolvers ++= Resolver.sonatypeOssRepos("snapshots")
resolvers ++= Resolver.sonatypeOssRepos("releases")
resolvers += Resolver.sonatypeRepo("releases")
```

raboof:
Ah, sorry for responding in a hurry, looks like it should have been `resolvers += Resolver.sonatypeRepo("snapshots")`, and placed in `build.sbt` rather than `project/plugins.sbt`

PR #1719, opened 2022-11-17
Send transitive dependencies in server reflection

### LINKED ISSUE #1718: Server reflection calls should return transitive dependencies
When the server reflection support returns file descriptors, it should not just return the file descriptor requested, but also all that file descriptors transitive dependencies. See:

```
    // This message is used to answer file_by_filename, file_containing_symbol,
    // file_containing_extension requests with transitive dependencies. As
    // the repeated label is not allowed in oneof fields, we use a
    // FileDescriptorResponse message to encapsulate the repeated fields.
    // The reflection service is allowed to avoid sending FileDescriptorProtos
    // that were previously sent in response to earlier requests in the stream.
    FileDescriptorResponse file_descriptor_response = 4;
```

Right now, we're only returning the file descriptor requested. This causes problems for clients that expect transitive dependencies to be included, and also makes server reflection very slow, because it means each descriptor must be fetched individually. We could of course implement support for tracking which we've already sent on a given stream, and not send them, that would be pretty straight forward.

jroper:
Fixes #1718

Not only has sending of transitive dependencies been implemented, but also filtering out ones already sent by the server on a given stream, as allowed by the server reflection spec.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,7 +58,8 @@ lazy val runtime = Project(id = akkaGrpcRuntimeName, base = file("runtime"))
     AutomaticModuleName.settings("akka.grpc.runtime"),
     ReflectiveCodeGen.generatedLanguages := Seq("Scala"),
     ReflectiveCodeGen.extraGenerators := Seq("ScalaMarshallersCodeGenerator"),
-    PB.protocVersion := Dependencies.Versions.googleProtobuf)
+    PB.protocVersion := Dependencies.Versions.googleProtobuf,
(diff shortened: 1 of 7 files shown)


lightbend-cla-validator:
Hi @jroper,

We see that you have signed the Lightbend Contributors License Agreement before, however, the CLA has changed since you last signed it.
Please review the new CLA and sign it before we proceed with reviewing this pull request:

johanandren (approved):
LGTM!

PR #1726, opened 2022-11-28
ci: JDK 8 for Fossa

ennru:
The target release flags for Scala and Java are missing. That mixes class file versions up when Fossa tries to run Gradle, AFAIU.
Trying this to see if it solves part of the problem before introducing flags into the sbt build.

Sample CI problem in https://github.com/akka/akka-grpc/actions/runs/3564968871/jobs/5989562205#step:5:486


ennru:
Superseded by #1727

PR #1727, opened 2022-11-29
build: set JDK 8 as target platform

ennru:
Introduce compiler flags to select JDK 8 as the target platform so that releases can be built on later JDKs.

### DIFF
--- a/project/Common.scala
+++ b/project/Common.scala
@@ -23,7 +23,7 @@ object Common extends AutoPlugin {
       organization := "com.lightbend.akka.grpc",
       organizationName := "Lightbend Inc.",
       organizationHomepage := Some(url("https://www.lightbend.com/")),
-      resolvers += Resolver.sonatypeRepo("staging"), // makes testing HTTP releases early easier
+      resolvers ++= Resolver.sonatypeOssRepos("staging"), // makes testing HTTP releases early easier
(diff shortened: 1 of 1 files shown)


project/Common.scala
```diff
@@ -23,23 +23,23 @@ object Common extends AutoPlugin {
...
       scmInfo := Some(ScmInfo(url("https://github.com/akka/akka-grpc"), "git@github.com:akka/akka-grpc")),
       developers += Developer(
         "contributors",
         "Contributors",
         "https://gitter.im/akka/dev",
         url("https://github.com/akka/akka-grpc/graphs/contributors")),
-      licenses := Seq(
-        ("BUSL-1.1", url("https://raw.githubusercontent.com/akka/akka-grpc/main/LICENSE"))
-      ), // FIXME change s/main/v2.2.1/ when released
+      licenses := Seq(("BUSL-1.1", url("https://raw.githubusercontent.com/akka/akka-grpc/v2.2.1/LICENSE"))),
```
patriknw, line 34: This is wrong. In `main` branch this should remain pointing to `main` until (before) we release 2.3.0. There is a task in the release template for that.

In 2.2.x branch we should change.

johanandren (approved):

PR #1738, opened 2022-12-20
bump(project): sbt-site 1.5.0-M3 (was 1.4.1)

ennru:
Bumping sbt-site to a renovated version to avoid a Scala XML mess.

~~Bumping up a patch of scalapb.~~


ennru:
The scalapb bump seems to break things in CI 🤔

johanandren (approved):

PR #1665, opened 2022-08-24
Rich error model client api

raboof:
Based on #1540 - see discussion there

PR #1740, opened 2022-12-21
feat: Better name for rich status metadata, better details lookup API

johanandren:
Follow up to #1665

### DIFF
--- a/interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorModelNativeTest.java
+++ b/interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorModelNativeTest.java
@@ -7,9 +7,7 @@ package example.myapp.helloworld.grpc;
 import akka.actor.ActorSystem;
 import akka.grpc.GrpcClientSettings;
 import akka.grpc.GrpcServiceException;
-import akka.grpc.internal.JavaMetadataImpl;
-import akka.grpc.internal.RichGrpcMetadataImpl;
-import akka.grpc.javadsl.RichMetadata;
+import akka.grpc.javadsl.MetadataStatus;
 import akka.http.javadsl.Http;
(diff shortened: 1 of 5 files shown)


interop-tests/src/test/java/example/myapp/helloworld/grpc/RichErrorModelNativeTest.java
```diff
@@ -67,13 +65,13 @@ public void testNativeApi() throws Exception {
             }).get();
 
             GrpcServiceException ex = GrpcServiceException.apply(statusRuntimeException);
-            RichMetadata meta = (RichMetadata) ex.getMetadata();
+            MetadataStatus meta = (MetadataStatus) ex.getMetadata();
```
johanandren, line 68: Went with status since that is the google rpc name of it

runtime/src/main/scala/akka/grpc/javadsl/Metadata.scala
```diff
@@ -50,13 +50,17 @@ trait Metadata {
...
 @ApiMayChange
 @DoNotInherit
-trait RichMetadata extends Metadata {
+trait MetadataStatus extends Metadata {
+  def getStatus(): com.google.rpc.Status
   def getCode(): Int
   def getMessage(): String
   def getDetails(): List[com.google.protobuf.any.Any]
-  def getParsedDetails[K <: scalapb.GeneratedMessage](index: Int, companion: scalapb.GeneratedMessageCompanion[K]): K
+  def getParsedDetails[K <: scalapb.GeneratedMessage](companion: scalapb.GeneratedMessageCompanion[K]): List[K]
```
johanandren, line 65: Likely contain only one of each type, but it's also possible with many, LocalizedMessage would be one per language for example.

ennru (approved):
LGTM.

Issue #1711, opened 2022-11-14
Failed: NonBalancingIntegrationSpecAkkaHttp


johanandren:
https://github.com/akka/akka-grpc/actions/runs/3461767253/jobs/5779790575#step:5:6504

```
[info] - should send requests to a single endpoint that is restarted in the middle *** FAILED *** (246 milliseconds)
[info]   62 was not equal to 100 (NonBalancingIntegrationSpec.scala:96)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.matchers.MatchersHelper$.indicateFailure(MatchersHelper.scala:407)
[info]   at org.scalatest.matchers.should.Matchers$ShouldMethodHelperClass.shouldMatcher(Matchers.scala:6778)
[info]   at org.scalatest.matchers.should.Matchers.should(Matchers.scala:6823)
[info]   at org.scalatest.matchers.should.Matchers.should$(Matchers.scala:1808)
[info]   at akka.grpc.scaladsl.NonBalancingIntegrationSpec.should(NonBalancingIntegrationSpec.scala:33)
[info]   at akka.grpc.scaladsl.NonBalancingIntegrationSpec.f$proxy2$1(NonBalancingIntegrationSpec.scala:96)
[info]   at akka.grpc.scaladsl.NonBalancingIntegrationSpec.$init$$$anonfun$1$$anonfun$2(NonBalancingIntegrationSpec.scala:68)
```

johanandren:
https://github.com/akka/akka-grpc/actions/runs/3525784747/jobs/5912931128#step:5:6323

johanandren:
https://github.com/akka/akka-grpc/actions/runs/3526228899/jobs/5913907347#step:5:6441

johanandren:
https://github.com/akka/akka-grpc/actions/runs/3583408086/jobs/6028798735#step:5:6405

johanandren:
https://github.com/akka/akka-grpc/actions/runs/3681671502/jobs/6228594973

patriknw:
https://github.com/akka/akka-grpc/actions/runs/3754393400/jobs/6378573366

octonato:
https://github.com/akka/akka-grpc/actions/runs/3784643666/jobs/6434095059#step:5:8929

octonato:
https://github.com/akka/akka-grpc/actions/runs/3804340546/jobs/6471472387#step:5:6395

johanandren:
[...]

PR #1742, opened 2023-01-05
fix: NonBalancingIntegrationSpec failing 

johanandren:
References #1711

After investigations in #1741 it seems the unbind-rebind not happening fast enough could be the culprit, we get a UNAVAILABLE response to one of the requests in the second instance batch.

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
@@ -80,20 +80,15 @@ class NonBalancingIntegrationSpec(backend: String)
 
       Future.sequence(requestsOnFirstConnection).futureValue
       server1.terminate(5.seconds).futureValue
+      // Give it some time to complete unbind (suspected to cause test flakyness)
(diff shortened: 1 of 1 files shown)


interop-tests/src/test/scala/akka/grpc/scaladsl/NonBalancingIntegrationSpec.scala
```diff
@@ -80,20 +80,15 @@ class NonBalancingIntegrationSpec(backend: String)
...
 
       val requestsOnSecondConnection = List.fill(requestsPerConnection)(client.sayHello(HelloRequest(s"Hello")))
 
-      val replies = Future.sequence(requestsOnSecondConnection)
-      Await.ready(replies, 15.seconds)
-
-      if (this.isInstanceOf[NonBalancingIntegrationSpecNetty] && service1.greetings.get != numberOfRequests) {
-        system.log.warning(
-          s"Found only ${service1.greetings.get} requests rather than $numberOfRequests, likely a flakey test")
-        pending
```
johanandren, line 94: I think the netty one was perhaps because of this as well, so re-enabling.

efgpinto (approved):
LGTM

PR #1745, opened 2023-01-23
fix: Don't log peer closing connection at error

johanandren:
Downstream fix for https://github.com/akka/akka-http/issues/4041

As we don't have information about how to reproduce, matching the logged error message is the best I could do and I think this is where the exception is logged at error.

### DIFF
--- a/runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
+++ b/runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
@@ -18,6 +18,7 @@ import io.grpc.{ Status, StatusRuntimeException }
 
 import scala.concurrent.ExecutionException
 import akka.event.Logging
+import akka.http.scaladsl.model.http2.PeerClosedStreamException
 
 @ApiMayChange
 object GrpcExceptionHandler {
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
```diff
@@ -32,7 +33,9 @@ object GrpcExceptionHandler {
 
   /** INTERNAL API */
   @InternalApi
-  private def default(system: ActorSystem): jFunction[Throwable, Trailers] =
+  private def default(system: ActorSystem): jFunction[Throwable, Trailers] = {
+    val log = Logging(system, "akka.grpc.javadsl.GrpcExceptionHandler")
+
```
patriknw, line 38: This isn't called repeatedly, which would add overhead of creating logger?

patriknw (approved):
LGTM

runtime/src/main/scala/akka/grpc/javadsl/GrpcExceptionHandler.scala
johanandren: In the Kalix proxy we create the partial function once at bootstrap (Serve.createRoute) and re-use for all calls, so a slight optimization for the peer closed and unexpected error scenarios. Not very important though.

Issue #1747, opened 2023-01-30
Include more error details in RuntimeStatusException


johanandren:
When getting an error status response, it is turned into a `io.grpc.StatusRuntimeException, with message: UNKNOWN: Unexpected error`, often with a stack that does not help (since async). 

Can we include more details about what service and method was called in there to make it easier to understand?

PR #1748, opened 2023-02-03
feat: Include information about called service in StatusRuntimeException

johanandren:
References #1747 

Is this a potential leak of internal details, should it be opt in through config? Otoh, if a service built with Akka gRPC just passes thrown exceptions with details and all to a calling client they already are likely leaking that kind of details to clients?

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/interop/AkkaGrpcScalaClientTester.scala
+++ b/interop-tests/src/test/scala/akka/grpc/interop/AkkaGrpcScalaClientTester.scala
@@ -270,6 +270,7 @@ class AkkaGrpcScalaClientTester(val settings: Settings, backend: String)(implici
     throwable shouldBe a[StatusRuntimeException]
     val e = throwable.asInstanceOf[StatusRuntimeException]
     assertEquals(expectedStatus.getCode, e.getStatus.getCode)
-    assertEquals(expectedMessage, e.getStatus.getDescription)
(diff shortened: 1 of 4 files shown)


franciscolopezsancho:
Is this giving back the clients' details to the client that made the call?

johanandren:
@franciscolopezsancho Imagine a service that accepts external client calls, and as a part of the logic for handling those calls, it calls some service over gRPC, if that fails and the service just passes that exception through to the client, that would be leaking the internal detail of what kind of service the other, possibly internal service is.

Best practice would generally be to not just pass through arbitrary exception details to clients, but log locally and give the client some "internal error" type of message. This is ofc different from for example validation errors/exceptions where the client sent something that was wrong, that information is good to send back so that the client can fix its request.

franciscolopezsancho:
I see more like this then. 
Paper German 34 2

franciscolopezsancho:
This better 
Paper German 34 3

johanandren:
Exactly, worst case an exception just passed through could contain credentials or something like that, what kinds of services it calls is less critical but still a leak of internal implementation information that does not generally belong in the hands of an external client.

franciscolopezsancho:
This makes me think that as you mentioned 'opt in' is the safest and probably most expected choice. IMHO

efgpinto (approved):
LGTM

This being the default behavior (rather than opt-in by config) makes sense IMO.

Issue #1300, opened 2021-03-19
Failure to import Google annotations proto


conradoverta:
### Versions used 

Akka version: 1.1.1

### Expected Behavior

When creating the Java files from protobuf in Maven with
```
            <plugin>
                <groupId>com.lightbend.akka.grpc</groupId>
                <artifactId>akka-grpc-maven-plugin</artifactId>
                <version>${akka.grpc.version}</version>
                <configuration>
                    <language>Java</language>
                    <includeStdTypes>true</includeStdTypes>
                    <protoPaths>
                        <protoPath>my_protos_path</protoPath>
                    </protoPaths>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>generate</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```
I'd expect the compiler to be able to pick up the standard file `google/api/annotations.proto` and correctly compile the protobufs.

### Actual Behavior

The compilation fails with
```
[ERROR] Failed to execute goal com.lightbend.akka.grpc:akka-grpc-maven-plugin:1.1.1:generate (default) on project registry: Execution default of goal com.lightbend.akka.grpc:akka-grpc-maven-plugin:1.1.1:generate failed: protoc exit code 1: google/api/annotations.proto: File not found.
```

For comparison, using
```
            <plugin>
                <groupId>org.xolstice.maven.plugins</groupId>
                <artifactId>protobuf-maven-plugin</artifactId>
                <version>0.6.1</version>
                <configuration>
                    <protocArtifact>com.google.protobuf:protoc:3.8.0:exe:${os.detected.classifier}</protocArtifact>
                    <pluginId>grpc-java</pluginId>
                    <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.16.1:exe:${os.detected.classifier}</pluginArtifact>
                    <protoSourceRoot>my_protos_path</protoSourceRoot>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>compile</goal>
                            <goal>compile-custom</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```
correctly creates the Java implementations.

I have also tried configuring `<protocVersion>-v3.8.0</protocVersion>` to match the versions but it gives the same error.

### Relevant logs

Relevant build logs shared above.

### Reproducible Test Case

[...]

Issue #686, opened 2019-09-17
Multiply protoPaths not compiling properly


Dr-Nikson:
### Versions used 

Akka version: 2.5.22
akka-grpc-runtime: 0.7.1
akka-grpc-maven-plugin: 0.7.1

### Expected Behavior

```
<protoPaths>
  <protoPath>src/main/proto</protoPath>
  <protoPath>../../protos/src/main/proto</protoPath>
</protoPaths>
```

I've configured two protoPaths and I'm expecting that plugin will look for all proto files before compilation starts. One of my protoPaths contains google prototobuf types, second protoPath contains my own protos based on the first ones.

### Actual Behavior

Compilation fails because it can't find proto files from one path while compiling another:

```google/protobuf/timestamp.proto: File not found.```

raboof:
Good observation! Would you be interested in trying to fix this and testing whether that works? It looks like in `GenerateMojo.scala` the list of paths should be passed into `generate` instead of calling `generate` for each path separately.

jewertow:
I could take care of this issue.

Tochemey:
Hello, please what is the outcome of this issue. I am facing the same issue too

jewertow:
I'm working on this issue.

Issue #152, opened 2018-04-06
Support for proto files from dependencies


raboof:

patriknw:
Is this needed from MVP? Is it difficult?

raboof:
Not sure if difficult, removing it from the MVP for now (though we might want/have to do it anyway to support the lagom workflow, that's an unknown for me at this point).

johanandren:
Did you learn that it was needed for Lagom @raboof given that you put it pack?

raboof:
I don't remember :hushed: ... @renatocaval what do you think, would it be common in Lagom to publish the .proto file in one artifact and generate the client in another, or would it be OK to copy the service definition manually for MVP?

johanandren:
Isn't Lagom using sbt though?

raboof:
Also Maven

pradasouvanlasy:
Hi guys,

while using Maven, an alternative could have been to leverage the `maven-dependency-plugin` and the `unpack` goal to extract `.proto` files from external dependencies and make them available at compile time.
However, I wanted to extract such external files somewhere under the `target` directory but it seems that the `akka-grpc-maven-plugin` does not support reading `.proto` files that are not within the `src` directory.

The obvious workaround would be to extract those external `.proto` files somewhere under `src` to duplicate them into the current module. I don't think this is ideal since it could allow people to make unwanted changes in those external `.proto` files.
WDYT? Would it make sense to you to allow `protoPaths` that are pointing somewhere else than `src`?

raboof:
> Will continue to experiment a bit to see whether I manage to reach an acceptable setup

Interesting, looking forward to what you'll find!

algolov:
Continuing @pradasouvanlasy thought, I came to the following fully working configuration:

```xml
<build>
    <plugins>
    	<!--  unpack proto files from external dependencies   -->
        <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-dependency-plugin</artifactId>
            <executions>
                <execution>
                    <id>unpack</id>
                    <phase>generate-sources</phase>
                    <goals>
                        <goal>unpack</goal>
                    </goals>
                    <configuration>
                        <artifactItems>
                            <artifactItem>
                                <groupId>com.some.company</groupId>
                                <artifactId>company-protos</artifactId>
                                <version>${company-protos.version}</version>
                                <type>jar</type>
                                <overWrite>true</overWrite>
[...]

PR #1768, opened 2023-04-12
fix: protoPaths as import params to protoc (Maven)

sebastian-alfers:
Pass all items listed in `<protoPath>` individually as `-I` to `protoc`.

Fixes:
- https://github.com/akka/akka-grpc/issues/1300
- https://github.com/akka/akka-grpc/issues/152 (examples taken from here)
- https://github.com/akka/akka-grpc/issues/686

### DIFF
--- a/plugin-tester-java/pom.xml
+++ b/plugin-tester-java/pom.xml
@@ -18,6 +18,9 @@
     <akka.http.cors.version>1.1.0</akka.http.cors.version>
     <grpc.version>1.48.1</grpc.version> <!-- checked synced by VersionSyncCheckPlugin -->
     <project.encoding>UTF-8</project.encoding>
+    <build-helper-maven-plugin>3.3.0</build-helper-maven-plugin>
+    <protobuf-java.version>3.22.2</protobuf-java.version>
+    <proto-google-common-protos.version>2.15.0</proto-google-common-protos.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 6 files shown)


plugin-tester-java/src/main/java/example/myapp/helloworld/GreeterServer.java
```diff
@@ -40,10 +46,17 @@ public static CompletionStage<ServerBinding> run(ActorSystem sys) throws Excepti
     // Instantiate implementation
     GreeterService impl = new GreeterServiceImpl(mat);
 
+    // Create the reflection handler for multiple services
+    Function<HttpRequest, CompletionStage<HttpResponse>> reflectionPartial =
+      ServerReflection.create(Arrays.asList(GreeterService.description), sys);
```
sebastian-alfers, line 51: Not needed, but helped me to test locally.

.github/workflows/build-test.yml
```diff
@@ -217,7 +217,7 @@ jobs:
       - name: Test Maven Java
         run: |-
           cd plugin-tester-java
-          mvn -Dakka.grpc.project.version=`cat ~/.version` akka-grpc:generate compile
```
sebastian-alfers, line 220: `akka-grpc:generate` should not be needed explicitly as it is already run bin `compile` (otherwise it would fail I assume).

project/Dependencies.scala
```diff
@@ -93,6 +97,7 @@ object Dependencies {
   val codegen = l ++= Seq(
     Compile.scalapbCompilerPlugin,
     Protobuf.protobufJava, // or else scalapb pulls older version in transitively
+    Protobuf.googleApi, // or else scalapb pulls older version in transitively
```
johanandren, line 100: Why does the codegen need a dependency on the common protos?

If we should always transitively pull it in, that should be from the (Java?) runtime rather, but should we have it as a transitive out of the box at all?

project/Dependencies.scala
```diff
@@ -27,6 +27,7 @@ object Dependencies {
     // If changing this, remember to update protoc plugin version to align in
     // maven-plugin/src/main/maven/plugin.xml and akka.grpc.sbt.AkkaGrpcPlugin
     val googleProtobuf = "3.21.9" // checked synced by VersionSyncCheckPlugin
+    val googleApi = "2.15.0" // checked synced by VersionSyncCheckPlugin
```
johanandren, line 30:
```suggestion
    val googleApi = "2.15.0" 
```
Or actually add it to the version sync check plugin for checking but I'm not sure that is needed.

plugin-tester-java/pom.xml
```diff
@@ -41,6 +44,56 @@
...
+                  <outputDirectory>${project.build.directory}/proto</outputDirectory>
+                  <includes>**/*.proto</includes>
+                </artifactItem>
+              </artifactItems>
+              <overWriteReleases>false</overWriteReleases>
+              <overWriteSnapshots>true</overWriteSnapshots>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
johanandren, line 95: Pull this plugin configs into the docs as an example for maven?

johanandren (commented):
Lookin good, left some comments

plugin-tester-java/pom.xml
sebastian-alfers: Yes, good idea. Working on that  right now.

project/Dependencies.scala
sebastian-alfers: You are right, not needed.

johanandren (approved):

Issue #1728, opened 2022-11-30
Client using netty backend cannot work with TLS and a custom SslContext


sndnv:
### Versions used 
sbt-akka-grpc @ 2.1.6

Akka version: 
akka @ 2.6.20

### Expected Behavior
gRPC client with TLS enabled using the netty backend is able to properly connect

### Actual Behavior
An exception is thrown by netty's `GrpcSslContexts.ensureAlpnAndH2Enabled`:
```
ALPN must be enabled and list HTTP/2 as a supported protocol.
```

### Relevant logs
See above

### Reproducible Test Case
Not a public repo so no PR to share

---
As far as I can tell, these issues are related to this piece of code:
```
new JdkSslContext(
      javaSslContext,
      /* boolean isClient */ true,
      /* Iterable<String> ciphers */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
      IdentityCipherSuiteFilter.INSTANCE,
      /* ApplicationProtocolConfig apn */ ApplicationProtocolConfig.DISABLED, // use JDK default (null would also be acceptable, DISABLED config will select the NONE protocol and thus the JdkDefaultApplicationProtocolNegotiator)
      ClientAuth.NONE, // server-only option, which is ignored as isClient=true (as indicated in constructor Javadoc)
      /* String[] protocols */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
      /* boolean startTls */ false)
```
and more specifically, [this](https://github.com/akka/akka-grpc/blob/v2.1.6/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala#L188) line:

```
 /* ApplicationProtocolConfig apn */ ApplicationProtocolConfig.DISABLED, // use JDK default (null would also be acceptable, DISABLED config will select the NONE protocol and thus the JdkDefaultApplicationProtocolNegotiator)
```

If the comment is to be trusted, `JdkDefaultApplicationProtocolNegotiator` will be used and that negotiator provides an empty list of protocols, triggering the above `GrpcSslContexts.ensureAlpnAndH2Enabled` check to fail.

To me it seems that since 2.1.6 there's no way to provide a custom `SslContext` context via `GrpcClientSettings.withSslContext` and have a working client (in other words, breaking change in #1462) or did I miss something?

johanandren:
What JDK version are you using when you see the error?

sndnv:
[...]

PR #1781, opened 2023-05-02
fix: mTLS server and client setup and docs

johanandren:
Includes fix for https://github.com/akka/akka-grpc/issues/1728


docs/src/main/paradox/server/mtls.md
```diff
@@ -0,0 +1,7 @@
+# Mutual authentication with TLS
+
+Mutual or mTLS means that just like how a client will only connect to servers with valid certificates, the server will
+also verify the client certificate and only allow connections if the client key pair is accepted by the server. This is
+useful for example in microservices where only other services are allowed to interact with a service, and public access
```
patriknw, line 5:
```suggestion
useful for example in microservices where only other known services are allowed to interact with a service, and public access 
```

plugin-tester-scala/src/main/scala/example/myapp/helloworld/MtlsGreeterServer.scala
```diff
@@ -0,0 +1,134 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.io.Source
+
+object MtlsGreeterServer {
+  def main(args: Array[String]): Unit = {
+    // Important: enable HTTP/2 in ActorSystem's config
+    // We do it here programmatically, but you can also set it in the application.conf
+    val conf = ConfigFactory
+      .parseString("akka.http.server.preview.enable-http2 = on")
```
patriknw, line 37: we changed the `preview`, or is this with older Akka HTTP?

docs/src/main/paradox/server/mtls.md
```diff
@@ -0,0 +1,7 @@
+# Mutual authentication with TLS
+
+Mutual or mTLS means that just like how a client will only connect to servers with valid certificates, the server will
+also verify the client certificate and only allow connections if the client key pair is accepted by the server. This is
+useful for example in microservices where only other services are allowed to interact with a service, and public access 
+should be denied.
+
```
patriknw, line 7: is the plan to include the `MtlsGreeterServer` as example here?
do we also need example of the client?

plugin-tester-scala/src/main/scala/example/myapp/helloworld/MtlsGreeterServer.scala
johanandren: Correct, but it seems the full bump was missed in akka-grpc, because we have Akka HTTP 10.5.0-M1 as a dependency also for relased versions, dang.

docs/src/main/paradox/server/mtls.md
johanandren: Yes, that's the plan, with the client, I need to figure out how to do that first though :)

plugin-tester-scala/src/main/scala/example/myapp/helloworld/MtlsGreeterServer.scala
johanandren: Fixing separately https://github.com/akka/akka-grpc/pull/1782

johanandren:
Caveat: not sure the added SSL context setup doesn't break all kinds of things when not doing mTLS, hoping CI will tell me

johanandren:
Some more testing/digging, the client SSLContext translation only happens when a custom SSLContext is defined, so regular use without that is completely unaffected by this, and I think all such customisation was broken based on what @patriknw noticed when trying to set up a custom context to accept a self-signed cert, which then worked with this patch.

So: ready for final review/merge

johanandren:
I'll add a test covering this as well, so we notice if we break TLS setup again

patriknw:
> Caveat: not sure the added SSL context setup doesn't break all kinds of things when not doing mTLS

I tested this with TLS but without mTLS and it works.

johanandren:
We could add an additional test for custom SSL only, but I think what that would catch is already covered by the mTLS test I wrote yesterday.

plugin-tester-java/src/main/java/example/myapp/helloworld/MtlsGreeterClient.java
```diff
@@ -0,0 +1,104 @@
...
+    } catch (Exception ex) {
+      throw new RuntimeException("Failed to set up SSL context for the client", ex);
+    }
+  }
+
+  private static String classPathFileAsString(String path) {
+    try {
+      BufferedInputStream in = new BufferedInputStream(MtlsGreeterServer.class.getResourceAsStream(path));
+      ByteArrayOutputStream bao = new ByteArrayOutputStream();
+      for (int result = in.read(); result != -1; result = in.read()) {
```
patriknw, line 95: Since this is text, shouldn't it use a Reader instead? Random example https://www.baeldung.com/convert-input-stream-to-string

Also, this is missing `close()` on the `in`. I think they are `Closable` so placing in `try (...) {` should automatically close.

plugin-tester-java/src/main/resources/certs/README.md
```diff
@@ -0,0 +1,22 @@
+# Some notes about the certs in this directory
+
+Self-signing sample CA in rootCA.*, CA cert secret: "secret"
+Server cert for localhost signed by rootCA in localhost-server.*, no password for private key
+Client cert for a client to connect in localhost-client.*, no password for private key
+
+Certs used by `MtlsGreeterServer`.
```
patriknw, line 7: For reference, should we include the commands for how the test certs were created?

patriknw (approved):
LGTM, after something minor in the samples

plugin-tester-java/src/main/java/example/myapp/helloworld/MtlsGreeterClient.java
johanandren: [92ee92d](https://github.com/akka/akka-grpc/pull/1781/commits/92ee92d3452d726cec04171c3712024008fdf92f)

plugin-tester-java/src/main/resources/certs/README.md
johanandren: [16c640a](https://github.com/akka/akka-grpc/pull/1781/commits/16c640ace8f8a3ec76e2fb7814d8aea9a5c3bfa5)

ihostage:
@patriknw @johanandren 
Hi, friends! 👋 
Any chance that this fix will be backported to `2.1.x` and released as `2.1.7` with APL 2.0? 

_Context_: Play gRPC stucks on `2.1.5` and can't bump to `2.1.6` by this problem 😞

johanandren:
I'm afraid we have no plans to backport any functionality added in 2.2 and later to the 2.1 branch

ihostage:
> I'm afraid we have no plans to backport any functionality added in 2.2 and later to the 2.1 branch

Will release be possible if I will do the small (without docs and samples) backport itself?

ihostage:
I tested locally and it looks as if we need only this piece of changes, nothing more

```diff
Subject: [PATCH] Backport #1781
---
Index: runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala
--- a/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala	(revision accf37ccc97205d588742ced4bcc7d1ed7499142)
+++ b/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala	(revision 1943f6dc29418f3332203fa3143b959aa60fd512)
@@ -5,7 +5,6 @@
 package akka.grpc.internal
 
 import java.util.concurrent.TimeUnit
-
 import javax.net.ssl.SSLContext
 import akka.{ Done, NotUsed }
 import akka.annotation.InternalApi
@@ -16,7 +15,17 @@
 import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts
 import io.grpc.netty.shaded.io.grpc.netty.NegotiationType
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
-import io.grpc.netty.shaded.io.netty.handler.ssl.{ SslContext, SslContextBuilder }
+import io.grpc.netty.shaded.io.netty.handler.ssl.ApplicationProtocolConfig.{
+  Protocol,
+  SelectedListenerFailureBehavior,
+  SelectorFailureBehavior
+}
+import io.grpc.netty.shaded.io.netty.handler.ssl.{
+  ApplicationProtocolConfig,
+  ApplicationProtocolNames,
+  SslContext,
+  SslContextBuilder
+}
 
 import scala.annotation.nowarn
 import scala.concurrent.duration.FiniteDuration
@@ -172,23 +181,24 @@
    */
   @InternalApi
   private def createNettySslContext(javaSslContext: SSLContext): SslContext = {
-    import io.grpc.netty.shaded.io.netty.handler.ssl.{
-      ApplicationProtocolConfig,
-      ClientAuth,
-      IdentityCipherSuiteFilter,
-      JdkSslContext
-    }
+    import io.grpc.netty.shaded.io.netty.handler.ssl.{ ClientAuth, IdentityCipherSuiteFilter, JdkSslContext }
     // See
     // https://github.com/netty/netty/blob/4.1/handler/src/main/java/io/netty/handler/ssl/JdkSslContext.java#L229-L309
-    new JdkSslContext(
+    val apn = new ApplicationProtocolConfig(
+      Protocol.ALPN,
+      SelectorFailureBehavior.NO_ADVERTISE,
+      SelectedListenerFailureBehavior.ACCEPT,
+      ApplicationProtocolNames.HTTP_2)
+    val context = new JdkSslContext(
       javaSslContext,
       /* boolean isClient */ true,
       /* Iterable<String> ciphers */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
       IdentityCipherSuiteFilter.INSTANCE,
-      /* ApplicationProtocolConfig apn */ ApplicationProtocolConfig.DISABLED, // use JDK default (null would also be acceptable, DISABLED config will select the NONE protocol and thus the JdkDefaultApplicationProtocolNegotiator)
-      ClientAuth.NONE, // server-only option, which is ignored as isClient=true (as indicated in constructor Javadoc)
+      /* ApplicationProtocolConfig apn */ apn,
+      ClientAuth.OPTIONAL, // server-only option, which is ignored as isClient=true (as indicated in constructor Javadoc)
       /* String[] protocols */ null, // use JDK defaults (null is accepted as indicated in constructor Javadoc)
       /* boolean startTls */ false)
+    context
   }
 
   /**

```

PR #1787, opened 2023-05-09
perf: Simple/small optimizations from #1391

johanandren:
References #1391

Two minimal changes and one opt that was never ported to the Java impl.


octonato (approved):
LGTM 🤷

PR #1788, opened 2023-05-09
perf: Thread local thread buffer for unmarshalling

johanandren:
Last optimisation from #1391

Trying to optimise deserialisation by going via byte buffer.
Couldn't see a perf improvement on aarch64 in a microbench, so needs further investigation.


johanandren:
Let's close this for now, we can revisit in the future.

PR #1812, opened 2023-06-29
build: Include release notes url in pom

johanandren:


johanandren:
Missing check is CLA validator which was unavailable because of some github instability yesterday.

PR #1833, opened 2023-09-09
fix: stream leak in akka-http client backend for indefinite server st…

### LINKED ISSUE #1832: Stream leak in akka-http backend (grpc client)
Consider the following scenario.

Server offering indefinite event stream:
```
service SomeService {
  rpc EventStream (EventStreamRequest) returns (stream StreamEvent) {}
}
```
Client is constantly connected to server, opening and cancelling event streams as needed by the application.

**Expected**
When client cancels event stream, stream on server side also gets closed. 

**Actual**
Everything works fine with netty backend configured in GRPC client. With akka-http backend configured, 
```
akka.grpc.client."*" {
  backend = "akka-http"
}
``` 
whenever the client cancels a stream, the server stream (being indefinite) keeps open. In an application scenario, where streams are constantly opened and canceled by the client, this will work until `akka.http.server.max-concurrent-streams` is reached, by default 256 streams. 

The cause of this issues lies in [runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala:231](https://github.com/akka/akka-grpc/blob/c33b9d1a2536f08f963d21fccc2041ce52d6515e/runtime/src/main/scala/akka/grpc/internal/AkkaHttpClientUtils.scala#L231)

**Solution**
CancellationBarrierGraphStage should only be added for non streaming calls.

steffenhaak:
Fixes #1832

### DIFF
--- a/plugin-tester-scala/src/test/scala/example/myapp/helloworld/AkkaHttpClientCancelSpec.scala
+++ b/plugin-tester-scala/src/test/scala/example/myapp/helloworld/AkkaHttpClientCancelSpec.scala
@@ -1,0 +1,81 @@
+/*
+ * Copyright (C) 2009-2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package example.myapp.helloworld
+
+import akka.Done
+import akka.NotUsed
+import akka.actor.testkit.typed.scaladsl.ScalaTestWithActorTestKit
+import akka.grpc.GrpcClientSettings
+import akka.http.scaladsl.Http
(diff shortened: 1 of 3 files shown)


johanandren (commented):
Thanks, would be good with test coverage. 

Could be added in `interop-tests` and have a client and server where the server closes over a promise, uses .watchTermination on the response stream to complete the promise and `future.failed.futureValue` as assert in the test case to wait for client cancelation to be seen on the server side.

patriknw (approved):
LGTM

PR #1847, opened 2023-09-29
build: Disable unidoc

johanandren:
Disabling for now, unidoc expects 2.12 artifacts for some reason, didn't manage to force it onto 2.13 with `scalaVersion`. Can see that it lists 2.12 as `compilers` but didn't get further than that.


patriknw (approved):
LGTM, yeah this build is difficult with the non overlapping cross builds

Issue #1423, opened 2021-08-03
Connection errors not reported for akka-http client backend


jrudolph:
Connection errors are not reported i.e. the response `Future` of a call is not failed.

The reason is that currently the akka-http `PersistentConnection` fails the connection stream if the connection fails ultimately. When that happens, we might have a request on the akka-http side outstanding (and more in the queue on the akka-grpc side).

We could try to report an error on the akka-http side but that is prone to race conditions (send error response vs stream failure, but also possibly send new request vs. stage failed).

To add that feature in akka-grpc, we need to keep track of ongoing requests. Is that what we want?

johanandren:
Trying out a possible upstream solution/precondition to solution here: https://github.com/akka/akka-http/issues/3892

johanandren:
Better upstream fix for the request triggering the failure: https://github.com/akka/akka-http/pull/4322

jtjeferreira:
@johanandren Since https://github.com/akka/akka-http/pull/4322 was merged and released this is no longer needed?

johanandren:
There is still an open PR with further fixes needed waiting (#1851)

PR #1851, opened 2023-10-06
fix: Fail queued requests on connection failure #1423

johanandren:
Without any stuff introduced in Akka streams, a switch over operator that will drain/continue to fail requests put in the buffer. Still doesn't fail the first request triggering the connection fail inside Akka HTTP/2 infra.

References #1423

Superseedes #1831

Related fix for failing the triggering request in Akka HTTP: https://github.com/akka/akka-http/pull/4322

### DIFF
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
@@ -1,3 +1,7 @@
+/*
+ * Copyright (C) 2023 Lightbend Inc. <https://www.lightbend.com>
+ */
+
 package akka.grpc.scaladsl
 
 import akka.actor.ActorSystem
--- a/interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
+++ b/interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
@@ -5,11 +5,12 @@
 package akka.grpc.scaladsl
 
 import akka.actor.ActorSystem
-import akka.grpc.GrpcClientSettings
+import akka.grpc.{ GrpcClientSettings, GrpcServiceException }
 import akka.http.scaladsl.model.HttpResponse
 import akka.testkit.TestKit
 import com.typesafe.config.ConfigFactory
 import example.myapp.helloworld.grpc.helloworld.{ GreeterServiceClient, HelloRequest }
+import io.grpc.Status
 import org.scalatest.Inspectors.forAll
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.matchers.should.Matchers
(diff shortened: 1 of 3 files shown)


johanandren:
Added an upstream fix for the triggering request in https://github.com/akka/akka-http/pull/4299

interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+
+      val futures = (1 to 10).map { _ =>
+        client.sayHello(HelloRequest())
+      }
+      // all should be failed
+      import system.dispatcher
+      // FIXME the first one in the pipe is still not failed :/ needs a fix in Akka HTTP
+      val lifted = Future.sequence(futures.tail.map(_.map(Success(_)).recover {
+        case th: Throwable => Failure[HttpResponse](th)
+      }))
```
johanandren, line 51: I wonder if it shouldn't be responses with StatusCodes.ServiceUnavailable instead of a failed future to align with how the Netty client behaves.

runtime/src/main/scala/akka/grpc/internal/SwitchOnCancel.scala
```diff
@@ -0,0 +1,74 @@
...
+import akka.util.OptionVal
+
+/**
+ * Identity stage that feeds all incoming events to output 1 until it cancels, then switches over to output 2, after
+ * completing the materialized value future.
+ *
+ * INTERNAL API
+ * @tparam T
+ */
+final private[akka] class SwitchOnCancel[T] extends GraphStage[FanOutShape2[T, T, (Throwable, T)]] {
```
johanandren, line 24: Might be useful enough that we should backport it to Akka proper. It's kind of like a Source.recover but in the other direction.

patriknw: What is the status of this PR? Should it be fixed in Akka or here?

johanandren: I think we could keep it as a custom stage here and, maybe decide to port back to Akka as public operator in the future.

interop-tests/src/test/scala/akka/grpc/scaladsl/AkkaHttpClientConnectionFailSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+      // Note that the Akka HTTP client does not strictly adhere to the gRPC backoff protocol but has its own
+      // backoff algorithm
+      val client = GreeterServiceClient(GrpcClientSettings.connectToServiceAt("127.0.0.1", 5).withTls(false))
+
+      val futures = (1 to 10).map { _ =>
+        client.sayHello(HelloRequest())
+      }
+      // all should be failed
+      import system.dispatcher
+      // FIXME the first one in the pipe is still not failed :/ needs a fix in Akka HTTP
```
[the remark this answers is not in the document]
johanandren: Merged and in release 10.6.0

johanandren:
Huh, not sure what is going on here, it is rebased on current main, but for some reason it still shows a ton of bumps from main here, bug in gh pr view?

johanandren:
Except for PR weirdness I think this is ready to merge now

patriknw (approved):
LGTM

Issue #1400, opened 2021-06-30
Publish the akka-grpc-codegen package for Scala 2.13


raboof:
It looks like we lost this in 18bcff9359f2ff4ac58a944c4a0f5e09270e82c5

bmo-at:
Any advancement on this? Or alternatively a workaround on how I can build 2.13 artifacts locally?

raboof:
I'm afraid not - possibly the changes in #1516 might help for this, but I ran into other problems there. Help welcome!

raboof:
> Or alternatively a workaround on how I can build 2.13 artifacts locally?

That you can do with `sbt ++2.13.8 akka-grpc-codegen/publishLocal`

bmo-at:
Ah well you see, I am not using sbt. I'm trying out mill today. I am currently in the process of exploring if I can migrate my "legacy" play-akka-grpc enpoint application to mill.
I am also trying to make use of [this plugin](https://github.com/tjarvstrand/mill-akka-grpc/), but development seems to have been abandoned and I am unsure whether it is worth investing significant time.

bmo-at:
But thank you for all the pointers anyway and the fast response!

bmo-at:
Ok, so that was a pretty stupid comment of me to make I realize now. I cloned the akka-grpc repo and used the local publishing to at least get a copy into my ivy2 cache that is 2.13 scala and 2.1.4 akka-grpc.
This at gets me a chance to work on the aforementioned plugin.
Maybe I can contribute back if the maintainer is still active/

raboof:
> Ok, so that was a pretty stupid comment of me to make I realize now.

No worries ;)

> I cloned the akka-grpc repo and used the local publishing to at least get a copy into my ivy2 cache that is 2.13 scala and 2.1.4 akka-grpc. This at gets me a chance to work on the aforementioned plugin.

Great! In the mean time I also added a workaround so now we publish snapshots to https://oss.sonatype.org/content/repositories/snapshots/com/lightbend/akka/grpc/akka-grpc-codegen_2.13 - and I *think* this should work for 2.1.5 as well

> Maybe I can contribute back if the maintainer is still active/

Would be great!

Let's close this for now and re-open if it turns out publishing for 2.1.5 didn't work after all.

PR #1852, opened 2023-10-09
build: Remove double 2.13 publish

patriknw:
publish fail in main: https://github.com/akka/akka-grpc/actions/runs/6453387739/job/17516874355

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Publish
         run: |-
           sbt +publishSigned
-          CI_RELEASE=akka-grpc-codegen/publishSigned CI_SNAPSHOT_RELEASE=akka-grpc-codegen/publish sbt ++2.13.10\! publishSigned
+          CI_RELEASE=akka-grpc-codegen/publishSigned CI_SNAPSHOT_RELEASE=akka-grpc-codegen/publish
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Publish
         run: |-
           sbt +publishSigned
-          CI_RELEASE=akka-grpc-codegen/publishSigned CI_SNAPSHOT_RELEASE=akka-grpc-codegen/publish sbt ++2.13.10\! publishSigned
+          CI_RELEASE=akka-grpc-codegen/publishSigned CI_SNAPSHOT_RELEASE=akka-grpc-codegen/publish
```
patriknw, line 32: I don't know the background for this, but could have been something with 2.12 and 2.13 cross publishing.
As far as I can see the above `sbt +publishSigned` will already publish for 2.13 and 3.

I don't know what that CI_RELEASE and CI_RELEASE is, but leaving that unchanged

johanandren: Looks like Raboof added that as a workaround for some publishing issues last year (https://github.com/akka/akka-grpc/issues/1400), not sure I understand why that is needed though.

johanandren (approved):

patriknw:
let's try

Issue #1854, opened 2023-10-10
NullPointerException in akka.grpc.internal.GrpcMetadataImpl.asList


jtjeferreira:
### Versions used 

Akka version: 2.8.5
Akka-http: 10.5.2
Akka-grpc version: 2.3.4
scala: 2.13.11

### Actual Behavior

I updated akka-grpc from 2.1.6. to 2.3.4 and I see a `NullPointerException` as shown in the below stack trace. The application is a grpc service (let's call it service A) that calls downstream service (let's call it service B). From the stack trace below one can see this is happening near the `GrpcExceptionHandler` (i.e when an exception is being converted to a grpc response). From what I collected from other telemetry sources this is only happening when service B returns an error (i.e grpc status code != Ok).

### Expected Behavior

No `NullPointerException` should be thrown 

### Relevant logs

```
Error during processing of request: 'java.lang.NullPointerException (No error message supplied)'. Completing with 500 Internal Server Error response. To change default exception handling behavior, provide a custom ExceptionHandler.
at akka.grpc.internal.GrpcMetadataImpl.asList(MetadataImpl.scala:106)
at akka.grpc.internal.GrpcEntityHelpers$.metadataHeaders(GrpcEntityHelpers.scala:74)
at akka.grpc.internal.GrpcEntityHelpers$.trailer(GrpcEntityHelpers.scala:67)
at akka.grpc.internal.GrpcResponseHelpers$.status(GrpcResponseHelpers.scala:101)
at akka.grpc.scaladsl.GrpcExceptionHandler$.$anonfun$from$1(GrpcExceptionHandler.scala:58)
at scala.PartialFunction$AndThen.applyOrElse(PartialFunction.scala:289)
at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:269)
at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:270)
at scala.concurrent.impl.Promise$Transformation.run$$original(Promise.scala:490)
at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:141)
at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:64)
at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:101)
at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[...]

PR #1855, opened 2023-10-11
fix: Avoid NPE in GrpcMetadataImpl #1854

johanandren:
Exception may not have any trailers. Also, fail fast if created with null delegate in some place we don't expected.

References #1854

### DIFF
--- a/runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
+++ b/runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
@@ -7,7 +7,7 @@ package akka.grpc
 import io.grpc.{ Status, StatusRuntimeException }
 import akka.annotation.ApiMayChange
 import akka.grpc.scaladsl.{ Metadata, MetadataBuilder }
-import akka.grpc.internal.{ GrpcMetadataImpl, JavaMetadataImpl, RichGrpcMetadataImpl }
+import akka.grpc.internal.{ JavaMetadataImpl, RichGrpcMetadataImpl }
 import com.google.protobuf.any.Any
 import io.grpc.protobuf.StatusProto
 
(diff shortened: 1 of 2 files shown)


runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
```diff
@@ -50,7 +54,12 @@ object GrpcServiceException {
   }
 
   def apply(ex: StatusRuntimeException): GrpcServiceException = {
-    new GrpcServiceException(ex.getStatus, new RichGrpcMetadataImpl(ex.getStatus, ex.getTrailers))
+    new GrpcServiceException(
+      ex.getStatus,
+      new RichGrpcMetadataImpl(
+        ex.getStatus,
```
jtjeferreira, line 60: Unrelated to this PR, but why we use `RichGrpcMetadataImpl`, but above we just use `GrpcMetadataImpl`?

johanandren: We want to propagate the structured error details if present, but maybe cleaner and more correct to select between GrpcMetadataImpl and RichGrpcMedataImpl on the presence of trailers instead of giving an empty metadata like I did. I'll do that instead.

johanandren: Hmm, looking at it some more, actually, I'm not so sure why we are not passing `RichGrpcMetadataImpl` above, that looks like the right thing to do.

johanandren: [312a11a](https://github.com/akka/akka-grpc/pull/1855/commits/312a11a0dc1b08d0e2657a9210df0dd3e76773a5) should do the trick

runtime/src/main/scala/akka/grpc/GrpcServiceException.scala
```diff
@@ -41,9 +41,7 @@ object GrpcServiceException {
 
     new GrpcServiceException(
       statusRuntimeException.getStatus,
-      new GrpcMetadataImpl(
-        // might not be present
-        Option(statusRuntimeException.getTrailers).getOrElse(new io.grpc.Metadata())))
+      new RichGrpcMetadataImpl(statusRuntimeException.getStatus, statusRuntimeException.getTrailers))
```
johanandren, line 44: we just converted, trailers always present here

jtjeferreira (approved):

patriknw (approved):
LGTM

PR #1868, opened 2023-10-31
build: Build and publish scaladocs for runtime module

johanandren:
We previously had to disable unidoc in #1847 because of the tricky build with 2.12 for plugins but 2.13 only for the runtime. This re-introduces at least scaladoc for the runtime module, published under akka-grpc/api/[version].

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 import akka.grpc.Dependencies
 import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
-import akka.grpc.ProjectExtensions.*
+import akka.grpc.ProjectExtensions._
 import akka.grpc.build.ReflectiveCodeGen
 import sbt.Keys.scalaVersion
 import com.geirsson.CiReleasePlugin
--- a/build.sbt
+++ b/build.sbt
@@ -1,10 +1,11 @@
 import akka.grpc.Dependencies
 import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
-import akka.grpc.ProjectExtensions._
+import akka.grpc.ProjectExtensions.*
 import akka.grpc.build.ReflectiveCodeGen
-import com.typesafe.tools.mima.core._
+import com.typesafe.tools.mima.core.*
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -1,10 +1,11 @@
 import akka.grpc.Dependencies
 import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
-import akka.grpc.ProjectExtensions._
+import akka.grpc.ProjectExtensions.*
 import akka.grpc.build.ReflectiveCodeGen
-import com.typesafe.tools.mima.core._
+import com.typesafe.tools.mima.core.*
```
patriknw, line 5: ?

build.sbt
```diff
@@ -176,11 +177,20 @@ lazy val benchmarks = Project(id = "benchmarks", base = file("benchmarks"))
     scalaVersion := Dependencies.Versions.CrossScalaForLib.head,
     (publish / skip) := true)
 
+// Config to allow only building scaladocs for runtime module but in/from the docs module
+val Runtime = config("runtime")
```
patriknw, line 181: name it something like grpc-runtime? "runtime" sounds like it could be something standard in sbt

patriknw (commented):
advanced, looking good

build.sbt
johanandren: Huh, IntelliJ what on earth are you up to?

johanandren: Also, why was it there to start with? Removed it now.

build.sbt
johanandren: done

build.sbt
```diff
@@ -1,10 +1,10 @@
 import akka.grpc.Dependencies
 import akka.grpc.Dependencies.Versions.{ scala212, scala213 }
-import akka.grpc.ProjectExtensions._
+import akka.grpc.ProjectExtensions.*
```
patriknw, line 3: One more?

patriknw (approved):
LGTM, after one more wildcard?

Issue #1152, opened 2020-10-20
Configure a way call the refresh api AkkaDiscoveryNameResolver preiodically or on event changes in the DNS


Marcus-Rosti:
### Short description

gRPC likes to cache all the dns lookups from AkkaDiscoveryNameResolver. On scale up events in kubernetes for instance, the client has no way then to discover those new endpoints until a cache gets invalidated and it reresolves them.

There should be a configuration to add a way to call the refresh endpoint in AkkaDiscoveryNameResolver periodically. 

### Details

Could just work via a tick in akka streams or it could be event driven but that may be too endpoint specific and more of an akka-management thing:

hand waving akka stream specifics

```scala
Source.tick(clientSettings.refreshDnsInterval.seconds, 0.seconds, ())
  .map(akkaDiscoveryNameResolver.refresh())
  .run()
```

Could also be solved via what @jrudolph mentions in https://github.com/akka/akka-http/issues/3505

danieltahara:
^ This would be very helpful (or at least some default). We just figured out this workaround after a painful incident.

johanandren:
Is this for a specific underlying discovery impl (either of the DNS ones?) If so I'd expect that the current behavior is that it stays in the cache of that until the TTL of the DNS entry runs out. That would mean that this feature request is basically about a way to specifically tell it to throw away the cached DNS entry for a service earlier (periodically or up to a given application to trigger).

I see that the Netty client API for name resolution (`io.grpc.NameResolver#NameResolver`) has a refresh method, but right now that will just hit the cached DNS value, not force-evict it. Not entirely clear to me exactly when Netty will call that (on transient connection failures).

If it is DNS I think we'd need to add support for that in the underlying resolver in Akka first, and then possibly add the capability of periodic forced-evict-re-refresh here. The evict would need to be followed by a call to the `io.grpc.NameResolver.Listener` with the fresh lookup results, so it's not easy to actually workaround with any public API as far as I can see.

danieltahara:
This is specifically for akka discovery. There are enough layers involved in the DNS resolver that I agree it would be a taller task. However, switching to use akka discovery (in particular for integration with kubernetes) allows us to hook into akka discovery itself to force refreshes, as suggested by the example above.

> Not entirely clear to me exactly when Netty will call that (on transient connection failures).

I believe this is correct, empirically.

johanandren:
Upstream cache-evict implemented in https://github.com/akka/akka/pull/32260

PR #1886, opened 2023-12-15
feat: Periodic client discovery refresh #1152

johanandren:
New per-service client config with an interval to trigger (possibly) cache-piercing refresh of discovered endpoints for a service.

Depends on upstream API addition in Akka.

References #1152


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -162,6 +166,12 @@ object GrpcClientSettings {
       case other => Some(other)
     }
 
+  private def getOptionalDuration(config: Config, path: String): Option[FiniteDuration] =
+    config.getString(path) match {
```
patriknw, line 170: toRootLowerCase

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -284,6 +295,28 @@ final class GrpcClientSettings private (
...
 
+  /**
+   * Scala API: Set this to a duration to trigger periodic refresh of the resolved endpoints, evicting cached entries
+   * if the discovery mechanism supports that. The default value `None` disables periodic refresh and instead
+   * only does refresh when the client implementation decides to.
+   *
+   * Currently only supported by the Netty client backend.
+   */
+  @ApiMayChange
+  def withDiscoveryRefreshInterval(refreshInterval: Option[FiniteDuration]): GrpcClientSettings =
```
patriknw, line 306: should this be plain FiniteDuration instead of Option? starting by default as "off" and can then be set with this?

patriknw (commented):
Looking good. I know the request has been periodic refresh, but I wonder if there is a way to trigger a refresh programatically for a specific gRPC client? Let's say there is some external event (maybe something observed from Kubernetes) and the application would like to trigger a refresh immediately when seeing that event.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
johanandren: Hmm, yeah, maybe cleaner. But then otoh it means no way to disable it for one service if you have set a default.

johanandren:
It is a bit hard to provide a regular programmatic hook since the `AkkaDiscoveryNameResolver` instance is created deep behind the scenes depending on backend and discovery used. I guess one way would be to have it listen to a stream or topic or something like that.

Note that the refresh is in the client config so can be set separately for each specific endpoint via service name in config (or with code in the used `GrpcClientSettings`).

johanandren:
@Marcus-Rosti @danieltahara how important is the ability to do a manual call if periodic refresh is in place with this PR?

danieltahara:
> @Marcus-Rosti @danieltahara how important is the ability to do a manual call if periodic refresh is in place with this PR?

I think for our use case, manual isn't necessary.

runtime/src/main/scala/akka/grpc/internal/AkkaDiscoveryNameResolver.scala
```diff
@@ -4,55 +4,101 @@
...
+          if (log.isDebugEnabled)
+            log.debug(
+              "Successful service discovery for service {}, found addresses: {}",
+              serviceName,
+              result.addresses.mkString(", "))
           listener.onAddresses(addresses(result.addresses), Attributes.EMPTY)
         } catch {
           case e: UnknownHostException =>
-            // TODO at least log
+            log.warning(e, s"Unknown host for service $serviceName")
```
patriknw, line 82: log placeholder?

runtime/src/main/scala/akka/grpc/internal/AkkaDiscoveryNameResolver.scala
```diff
@@ -4,55 +4,101 @@
...
             listener.onError(Status.UNKNOWN.withDescription(e.getMessage))
         }
       case Failure(e) =>
-        // TODO at least log
+        log.warning(e, s"Service discovery failed for service $serviceName")
```
patriknw, line 86: ^

patriknw (approved):
LGTM after some log minors

PR #1909, opened 2024-03-07
fix: do not publish sbt-akka-gprc jar twice (under two different names)

sebastian-alfers:
Before this PR, it seems that for the `sbt-akka-grpc` submodule, a `publish` does publish the plugin jar twice as follows:
- `sbt-akka-grpc_2.12_1.0-2.4.1.jar` and
- `sbt-akka-grpc-2.4.1.jar`.

Both files are identical, just carry a different name.

This has caused some issues on the hosting side of things, so this PR removes one of the two artifacts during `publish`.


johanandren (approved):
LGTM i guess

PR #1912, opened 2024-03-08
build: filter out 2.12_1.0 artifacs to fix publishing of sbt-akka-grpc

sebastian-alfers:
Only after https://github.com/akka/akka-grpc/pull/1911 got merged and released, it showed that the wrong artifacts were filtered out.


johanandren (approved):
LGTM

PR #1927, opened 2024-04-02
docs: Build tool with resolver config in more places

johanandren:
Backed out on @octonato s suggestion to add it also to the README, with the three different build tools it just becomes a pages long scroll unless you happened to like the first build tool best.


johanandren:
Yes, but we should merge and release this first or else those links will not point to pages with any details but just another link into the walkthrough. Changes to README is now but docs only update with release.

PR #1906, opened 2024-02-23
Include metadata in StatusRuntimeException

jtjeferreira:
Includes metadata in `StatusRuntimeException` both when metadata is in headers or trailers

References #1814

PR #1929, opened 2024-04-09
Fix binary trailer decoding

jtjeferreira:
In https://github.com/akka/akka-grpc/pull/1906 I introduced a bug in the trailer decoding where I didn't take into consideration that there are "ASCII headers" and "binary headers"

This PR adds a failure test and the fix in separate commits


jtjeferreira:
As expected the tests of 7cca84e fail with `java.lang.IllegalArgumentException: ASCII header is named custom-key-bin.  Only binary headers may end with -bin was not an instance of io.grpc.StatusRuntimeException, but an instance of java.lang.IllegalArgumentException (AkkaHttpClientUtilsSpec.scala:53)`

johanandren (approved):
LGTM, thanks!

Issue #1930, opened 2024-04-16
Add Channel State API 


enforser:
### Short description

I've noticed that there is a significant delay on the first request of a newly instantiated client, which is caused by the extra time it takes to create the ManagedChannel for the client, and then to establish a connection with the server. 

There [exists a solution to this in other GRPC libraries](https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md) where the channel state is exposed, and a method to eagerly transition to a connected state is provided. 

With this functionality a user of akka-grpc could establish a connection early without making an actual RPC request to the server. 

The effect of the connection creation overhead in a multi service architecture can lead to high delays that are not acceptable in low latency environments. Anecdotally, I am working with a system which makes 4 GRPC calls between four different services all synchronously (a -> b -> c -> d -> e). If I exclude the requests which occur on the first call from the client, the average time is far less than a second. On the first request though we consistently see 3-4s latencies.

johanandren:
I haven't really dug into what would be required/is easily doable, but instead of adding such a low level api, I wonder if we could have an eager connect client setting. Or do you think it would be important to also be able to imperatively query-trigger connect later during the lifecycle of a client?

enforser:
Eager connect sounds great and would suit my use case perfectly. In fact I think it would be fairly straightforward to add that. 

It would just be calling `.getState(true)` on the `ManagedChannel` returned here: https://github.com/akka/akka-grpc/blob/main/runtime/src/main/scala/akka/grpc/internal/NettyClientUtils.scala#L112-L126

PR #1931, opened 2024-04-17
feat: Eager connect flag for Netty based client

johanandren:
Will casue the client creation to trigger connection to server rather than first on initial request, can help avoiding latency for the first requests caused by waiting for connection to complete.

References #1930

### DIFF
--- a/runtime/src/main/resources/reference.conf
+++ b/runtime/src/main/resources/reference.conf
@@ -6,6 +6,10 @@ akka.grpc.client."*" {
   # Host to use if service-discovery-mechanism is set to static or grpc-dns
   host = ""
 
+  # Service discovery mechamism to use. The default is to use a static host
+  # and port that will be resolved via DNS.
+  # Any of the mechanisms described in https://doc.akka.io/docs/akka-management/current/discovery/index.html can be used
+  # including Kubernetes, Consul, AWS API
   service-discovery {
     mechanism = "static"
     # Service name to use if a service-discovery.mechanism other than static or grpc-dns
@@ -54,9 +58,9 @@ akka.grpc.client."*" {
   # interpreted as retrying 'indefinitely'.
   connection-attempts = 20
 
-  # Service discovery mechamism to use. The default is to use a static host
-  # and port that will be resolved via DNS.
-  # Any of the mechanisms described in https://doc.akka.io/docs/akka-management/current/discovery/index.html can be used
-  # including Kubernetes, Consul, AWS API
+  # Request that the client try to connect the service immediately when the client is created
+  # rather than on the first request. Only supported for the Netty client backend, the akka-http client backend
+  # is always eager.
+  eager-connection = off
 }
 //#defaults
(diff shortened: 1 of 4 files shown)


runtime/src/main/resources/reference.conf
```diff
@@ -54,9 +58,9 @@ akka.grpc.client."*" {
...
   connection-attempts = 20
 
-  # Service discovery mechamism to use. The default is to use a static host
-  # and port that will be resolved via DNS.
-  # Any of the mechanisms described in https://doc.akka.io/docs/akka-management/current/discovery/index.html can be used
-  # including Kubernetes, Consul, AWS API
+  # Request that the client try to connect the service immediately when the client is created
+  # rather than on the first request. Only supported for the Netty client backend, the akka-http client backend
+  # is always eager.
+  eager-connection = off
```
patriknw, line 64: Shall we add something to the client reference docs about this? Seems like something you want to enable most of the time.

patriknw (commented):
looking good, maybe some more docs

runtime/src/main/resources/reference.conf
johanandren: I'm not entirely sure it is something you want enabled most of the time, it since it would also means you might create a bunch of connections for clients that don't see any traffic so wastes resource usage (at least I'm guessing that is why it is lazy by default). For the record clients will also idle timeout after a default 5 minutes without any requests and switch to an "IDLE" state and only reconnect when a new request is made.

johanandren: Mentioning in docs seems good though.

patriknw: ok, yes we should not change the default

johanandren: Added in [6222128](https://github.com/akka/akka-grpc/pull/1931/commits/6222128c97930bf54839cdc1f28374d00a67da09)

runtime/src/main/resources/reference.conf
```diff
@@ -6,6 +6,10 @@ akka.grpc.client."*" {
   # Host to use if service-discovery-mechanism is set to static or grpc-dns
   host = ""
 
+  # Service discovery mechamism to use. The default is to use a static host
```
ennru, line 9:
```suggestion
  # Service discovery mechanism to use. The default is to use a static host
```

runtime/src/main/resources/reference.conf
```diff
@@ -54,9 +58,9 @@ akka.grpc.client."*" {
   # interpreted as retrying 'indefinitely'.
   connection-attempts = 20
 
-  # Service discovery mechamism to use. The default is to use a static host
-  # and port that will be resolved via DNS.
-  # Any of the mechanisms described in https://doc.akka.io/docs/akka-management/current/discovery/index.html can be used
-  # including Kubernetes, Consul, AWS API
+  # Request that the client try to connect the service immediately when the client is created
+  # rather than on the first request. Only supported for the Netty client backend, the akka-http client backend
+  # is always eager.
```
ennru, line 63:
```suggestion
  # Request that the client tries connecting to the service immediately when the client is created
  # rather than on the first request. Only supported for the Netty client backend, the Akka HTTP client backend
  # is always eager.
```

ennru (approved):
LGTM.

PR #1933, opened 2024-04-18
Add Trailers.apply to construct from code, message, and details

alexklibisz:
The handler generated by akka-grpc takes an error handler in the form of `eHandler: ActorSystem => PartialFunction[Throwable, Trailers]`.

This change just adds a method to make it easier to construct an instance of Trailers containing rich error details, so that we can enrich the error response from within the error handler.

Couple things I'm not sure about in the initial implementation:

1. This implementation delegates to the analogous `GrpcServiceException.apply` method. Is there a preferred way to would abstract out the common code? I.e., create a Utils object somewhere.
2. The runtime/test submodule does not seem to import any instances of GeneratedMessage, so I'm not sure how to test that we're encoding details correctly. Should we modify the subproject so that it imports LocalizedMessage? Should we put the tests in another submodule.


runtime/src/main/scala/akka/grpc/Trailers.scala
```diff
@@ -36,4 +36,9 @@ class Trailers(val status: Status, val metadata: Metadata) {
 object Trailers {
   def apply(status: Status): Trailers = new Trailers(status)
   def apply(status: Status, metadata: Metadata): Trailers = new Trailers(status, metadata)
+
+  def apply(code: com.google.rpc.Code, message: String, details: Seq[scalapb.GeneratedMessage]): Trailers = {
+    val ex = GrpcServiceException(code, message, details)
+    Trailers(ex.status, ex.metadata)
+  }
```
johanandren, line 43: Since this is common API for Java and Scala we should have one that accepts a Java collection as well, and transform/delegate. Usually in Akka land we call those `create` but just another `apply` overload converting the java collection using `.toScala.toSeq` would suffice.

I think it's fine to delegate to a throw-away GrpcServiceException for now, once we need it in a third place we can revisit extracting some utility.

johanandren (commented):
A lot of the test coverage is in the `akka-grpc-plugin-tester-{java,scala}` modules which have more access to actual message generation etc. so maybe sneak in some coverage there in `LoggingErrorHandlingGreeterServer.java` and `LoggingErrorHandlingGreeterServer.scala`. Those also has snippets that goes in the docs which may be useful to show off the new signature. For example the `custom-error-mapping` snippet.

Looks like there is no actual test calling those, but since this mostly adds convenience and no real logic having it used in something that is compiled as a part of the test suite is probably enough.

alexklibisz:
I added the `create` method in https://github.com/akka/akka-grpc/pull/1933/commits/ed7a91359de5b3a28db1a639834ac5e0abf1699e

And added an example to the LoggingErrorHandlingGreeterServer in https://github.com/akka/akka-grpc/pull/1933/commits/18856521d354a8e9030b9d2d26485bd95fb44340.

If that example looks good, I can do the same on the Java side.

johanandren:
That example looks good, please add for Java as well 👍

plugin-tester-java/src/main/java/example/myapp/helloworld/LoggingErrorHandlingGreeterServer.java
```diff
@@ -100,7 +111,19 @@ private static <ServiceImpl> Route loggingErrorHandlingGrpcRoute(
   //#custom-error-mapping
   private final static Function<Throwable, Trailers> customErrorMapping = (throwable) -> {
     if (throwable instanceof IllegalArgumentException) {
-      return new Trailers(Status.INVALID_ARGUMENT);
+      if (throwable.getMessage().equals(nameNonEmptyException.getMessage())) {
+        // We can pass through the message by attaching it to the Status.
+        return new Trailers(Status.INVALID_ARGUMENT.withDescription(throwable.getMessage()));
+      } else if (throwable.getMessage().equals(nameCapitalizedException.getMessage())) {
+        // We can pass through extra error details like localized versions of the message.
+        // FIXME: This subproject uses Java protos, which do not extend GeneratedMessage, so the code below does not compile.
```
alexklibisz, line 119: There's a bit of an issue here @johanandren 

Should we really be using scalapb.GeneratedMessage in the Java APIs? If so, then we need to either make the Scala LocalizedMessage available, or we need to find a GeneratedMessage that is appropriate to use here.

alexklibisz:
Ran into an issue in the Java example. Commented inline.

plugin-tester-java/src/main/java/example/myapp/helloworld/LoggingErrorHandlingGreeterServer.java
johanandren: Yes, good catch, for Java we should indeed have APIs for the Java proto message classes (`com.google.protobuf.Message`) instead of the ScalaPB ones. 

Looks like an oversight already in the pre-existing `GrpcServiceException.create`. I've PR:ed a fix for that in https://github.com/akka/akka-grpc/pull/1937 once that is merged you can rebase this on top and change the Java Trailers.create to accept a list of the Java type instead.

alexklibisz: Nice! Fixed in https://github.com/akka/akka-grpc/pull/1933/commits/f452a6a76d2667bb872c4c76f2ca54c8686bace3

alexklibisz:
I'm seeing an error in the Gradle build: https://github.com/akka/akka-grpc/actions/runs/8851059859/job/24306757026?pr=1933#step:9:131

```
Error:  /home/runner/work/akka-grpc/akka-grpc/plugin-tester-scala/src/main/scala/example/myapp/helloworld/LoggingErrorHandlingGreeterServer.scala:89: class com.google.rpc.LocalizedMessage is not a value
```

on the line:

```scala
Trailers(com.google.rpc.Code.INVALID_ARGUMENT, ex.getMessage, List(LocalizedMessage("en-US", ex.getMessage)))
```

I guess it's possible that gradle isn't configured to know about the scalapb-generated LocalizedMessage. 🤔

alexklibisz:
@johanandren This is ready for another look. I might need a hint on how to get the Gradle Scala build working.

johanandren:
The plugin-tester-scala is used for verifying building Scala projects with gradle and maven in addition to the "normal" sbt, so those two project files needed some ammends, I've pushed fixes that should sort it.

johanandren (approved):
LGTM, thanks! Tricker than you'd think at first.

PR #1937, opened 2024-04-22
fix: Java API for rich errors was not usable from Java

johanandren:
We were accepting the ScalaPB generated top message type in the Java API which likely means no Java project could ever use the detailed errors. 

This breaks binary and source compatibility but since it is unlikely that anyone could really make use of the API I think that is fine.

### DIFF
--- a/docs/src/main/paradox/server/details.md
+++ b/docs/src/main/paradox/server/details.md
@@ -52,6 +52,6 @@ Scala
 :    @@snip[RichErrorModelSpec](/interop-tests/src/test/scala/akka/grpc/scaladsl/RichErrorModelNativeSpec.scala) { #rich_error_model_unary }
 
 Java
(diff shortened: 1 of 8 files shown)


docs/src/main/paradox/server/details.md
```diff
@@ -52,6 +52,6 @@ Scala
 :    @@snip[RichErrorModelSpec](/interop-tests/src/test/scala/akka/grpc/scaladsl/RichErrorModelNativeSpec.scala) { #rich_error_model_unary }
```
sebastian-alfers, line 52: See above:

> every class that is based on `scalapb.GeneratedMessage` can be used

Should we update this as well to align?

sebastian-alfers: At least for Java.

sebastian-alfers (approved):
LGTM, beside the change in the server docs.

PR #1961, opened 2024-07-19
Provide access to akka-http request/response attributes

### LINKED ISSUE #859: Ability to pass strongly typed context through metadata
Consider the following use case:

I would like to be able to be able to apply a filter to all my gRPC calls to report analytics. These analytics should include the currently logged in user.

Currently, this not very practical in Akka gRPC. You can wrap an Akka gRPC service in a filter easily enough. In that filter, you can look at the headers, and extract the currently logged in user - however, doing so very often requires some IO, eg, looking up a session token in a database. You don't want to do this multiple times per request. So, if your filter looks up the current user, and loads that user from the database, you then want the filter to be able to pass that user object as context for the request through to the gRPC service call, which will then use that user object itself for authorization. This is not possible, the only context that Akka gRPC allows to be passed is String or ByteStrings, a strongly typed user object is not possible. In my case, it's much more than a user id, it will include things like execution scopes (eg, like oauth scopes, each service call is constraint to only be executed by a particular scope), roles the user has, the users email address, etc. Ideally I don't want to have to serialize this to bytes so that I can then deserialize it again in the service call.

jroper:
Fixes #859

This adds an `attribute`/`getAttribute` method to `Metadata`, and connects it to the underlying request/response that the metadata has been sourced from.

I haven't implemented passing attributes on from user created metadata, I'm not sure if there's a use case for that.


johanandren (approved):
LGTM

PR #1972, opened 2024-09-04
feat: Allow rotation of client certs

johanandren:


interop-tests/src/main/scala/akka/grpc/interop/AkkaGrpcClientScala.scala
```diff
@@ -22,6 +25,16 @@ final case class AkkaGrpcClientScala(clientTesterFactory: Settings => ActorSyste
...
+    // alternatively `.withSslContext` directly from a SSLContextFactory.createSSLContextFromPem if not
+    // rotating certs
+    GrpcClientSettings
+      .fromConfig(sys.settings.config)
+      .withSslContextProvider(SSLContextFactory.refreshingSSLContextProvider(5.minutes)(() =>
+        SSLContextFactory.createSSLContextFromPem(
+          certificatePath = Paths.get("/some/path/server.crt"),
+          privateKeyPath = Paths.get("/some/path/server.key"),
+          trustedCaCertificatePaths = Seq(Paths.get("/some/path/serverCA.crt")))))
+
```
johanandren, line 37: Temporary API test for now, doesn't actually use it. I need to add some proper test for that, after verifying it actually works at all.

jroper:
LGTM

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -222,18 +224,33 @@ final class GrpcClientSettings private (
...
+  def withSslContextProvider(provider: () => SSLContext): GrpcClientSettings =
+    copy(useTls = true, sslContextProvider = Some(provider))
+
+  /**
+   * Java API: For each new client connection, invoke this provider to set up TLS, useful for example for using rotating
+   * certs for the client provided by `SSLContextFactory.refreshingSSLEngineProvider`.
+   *
+   * When setting this the other TLS settings (`sslContext`,`tlsManager`) must not be set.
+   */
+  def withSslContextProvider(provider: akka.japi.function.Creator[SSLContext]): GrpcClientSettings =
```
patriknw, line 251: no risk that those two Scala and Java signatures are somehow the same?

johanandren: Good catch, they probably end up problematic with SAMs, not sure why examples didn't trigger a warning

plugin-tester-scala/src/main/scala/example/myapp/helloworld/MtlsGreeterClient.scala
```diff
@@ -26,19 +28,25 @@ object MtlsGreeterClient {
...
+      GrpcClientSettings.connectToServiceAt("localhost", 8443).withSslContextProvider(rotatingSslContext())
 
-    val reply = client.sayHello(HelloRequest("Jonas"))
+    val client = GreeterServiceClient(rotatingClientSettings)
 
-    reply.onComplete { tryResponse =>
+    while (true) {
+      val reply = client.sayHello(HelloRequest("Jonas"))
+
+      val tryResponse = Try { Await.result(reply, 5.seconds) }
```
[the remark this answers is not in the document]
johanandren: I think I should change this part from the forever loop, I used to verify that things work but did not intend to commit.

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
johanandren: Verified, what doesn't work is a call in Java with a lambda, then it becomes ambiguous, the sample code explicitly used a `Creator<SSLContext>` created in a method so I didn't notice.

patriknw (approved):
LGTM

PR #1973, opened 2024-09-05
perf: Trailers only support

jroper:
An optimisation exists in the gRPC spec that allows responses that immediately return an error to return it as a "Trailers-Only" response. This is where the response has no body, rather, the `grpc-status`, `grpc-message` and `grpc-status-details-bin` trailers are placed in the response headers frame.

When adding support for this, I found a bug in the Akka gRPC client where trailers only responses were not correctly handled in streamed requests (the error returned to the client code was an `AbruptStageTerminationException`).

### DIFF
--- a/runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
+++ b/runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
@@ -102,6 +102,10 @@ private final class AkkaNettyGrpcClientGraphStage[I, O](
         override def onMessage(message: O): Unit =
           callback.invoke(message)
         override def onClose(status: Status, trailers: Metadata): Unit = {
+          if (!matVal.isCompleted) {
+            // Trailers only response, first invoke onHeaders to setup the materialized value
+            onHeaders(trailers)
+          }
           trailerPromise.success(trailers)
           callback.invoke(Closed(status, trailers))
         }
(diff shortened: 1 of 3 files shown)


runtime/src/main/scala/akka/grpc/internal/GrpcResponseHelpers.scala
```diff
@@ -51,11 +51,7 @@ object GrpcResponseHelpers {
     try writer.encodeDataToResponse(m.serialize(e), responseHeaders, TrailerOkAttribute)
     catch {
       case NonFatal(ex) =>
-        val trailers = GrpcEntityHelpers.handleException(ex, eHandler)
-        writer.encodeDataToResponse(
-          ByteString.empty,
-          responseHeaders,
-          Trailer(GrpcEntityHelpers.trailer(trailers.status, trailers.metadata).trailers))
```
jroper, line 58: This is actually a bug, the empty byte string gets encoded as a gRPC message frame, so this is equivalent to sending an empty gRPC message before sending a failure message, which many/most unary clients ignore. That said, errors encountered during protobuf serialization would not be common, if they're even possible at all, so I'm not sure that this exception handler was ever triggered.

runtime/src/main/scala/akka/grpc/internal/AkkaNettyGrpcClientGraphStage.scala
```diff
@@ -102,6 +102,10 @@ private final class AkkaNettyGrpcClientGraphStage[I, O](
         override def onMessage(message: O): Unit =
           callback.invoke(message)
         override def onClose(status: Status, trailers: Metadata): Unit = {
+          if (!matVal.isCompleted) {
+            // Trailers only response, first invoke onHeaders to setup the materialized value
+            onHeaders(trailers)
```
jroper, line 107: So, `trailers` in this case is the response header, and it contains the trailers. There's no way to determine which headers are for which - of course, `grpc-status` and `grpc-message` etc are obviously trailers, but there could presumably be other things that we can't predict should be in one or the other. So, code that's expecting to read something from the response header needs this metadata made available as a response header, and code that's expecting to read something from the trailer needs the same metadata made available as a trailer. So, that's why I've passed it to both.

PR #1974, opened 2024-09-06
feat: Load client config from arbitrary config block

johanandren:


runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -46,17 +46,22 @@ object GrpcClientSettings {
...
-    }
+    val akkaGrpcClientConfig = actorSystem.classicSystem.settings.config.getConfig("akka.grpc.client")
+    fromConfig(clientName, akkaGrpcClientConfig)
+  }
 
+  /**
+   * Look up client settings from the given configuration. Will look for an entry with the given name client name
+   * directly in the config block. Each client configuration falls back to the defaults defined in reference.conf
+   *
+   * @param clientName of the client configuration to lookup config from the ActorSystem's config
```
patriknw, line 57: update this description

patriknw (commented):
thanks, that would be useful

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
```diff
@@ -46,17 +46,22 @@ object GrpcClientSettings {
...
+  /**
+   * Look up client settings from the given configuration. Will look for an entry with the given name client name
+   * directly in the config block. Each client configuration falls back to the defaults defined in reference.conf
+   *
+   * @param clientName of the client name to lookup config from the given config
+   */
+  def fromConfig(clientName: String, config: Config)(
+      implicit actorSystem: ClassicActorSystemProvider): GrpcClientSettings = {
+    // Use config named "*" by default
+    val defaultServiceConfig = actorSystem.classicSystem.settings.config.getConfig("akka.grpc.client.\"*\"")
```
patriknw, line 62: I was considering if this should be `config.withFallback(actorSystem.classicSystem.settings.config)` in case the given config also overrides default config, but fine to not support that kind of override.

patriknw (approved):
LGTM

runtime/src/main/scala/akka/grpc/GrpcClientSettings.scala
johanandren: Hmm, yeah, maybe not a bad idea.

johanandren: Let's leave that and revisit in case we figure out that it is very useful.

PR #1978, opened 2024-09-25
fix: correct akkaBinary value

sebastian-alfers:
Thanks @ennru for pointing it out!

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,7 +18,7 @@ object Dependencies {
     // themselves. For more information see
     // https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
     val akka = "2.10.0-M1"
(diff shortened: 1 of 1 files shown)


project/Dependencies.scala
```diff
@@ -18,7 +18,7 @@ object Dependencies {
     // themselves. For more information see
     // https://doc.akka.io//docs/akka/current/project/downstream-upgrade-strategy.html
     val akka = "2.10.0-M1"
-    val akkaBinary = "2.11"
+    val akkaBinary = "2.10"
```
ennru, line 21: We avoid these mistakes in the other repos with
```suggestion
    val akkaBinary = VersionNumber(akka).numbers match { case Seq(major, minor, _*) => s"$major.$minor" }
```

ennru (approved):
LGTM.

PR #2014, opened 2025-01-09
build: Scaladoc JDK links handling breaking build

johanandren:
We cross build for 2.12 because of sbt, and the `-jdk-api-doc-base` setting does not exist in 2.12

This only shows in ["check samples" CI build](https://github.com/akka/akka-grpc/actions/workflows/check-samples.yml), and that doesn't have workflow_dispatch, so we have to merge this to see that it works.


ennru (approved):
LGTM.

PR #2015, opened 2025-01-20
feat: Per request instance factory for Java DSL

johanandren:
WIP


octonato (approved):
LGTM, with a question...

Did you call it ScalaHandler because it's based on a Scala PF? 

It's a little weird that a parameter called `generate_scala_handler_factory` generates a Java file.

johanandren:
It's ScalaHandler because the public API of it is Scala (PF and Future) for folding into a Scala HTTP server, compared to the normal handler factory which has API to create `akka.japi.function.Function` and `CompletionStage` to be bound using the Java API of Akka HTTP

johanandren:
And, should I perhaps add, generated as Java to not require Akka SDK projects to have a Scala compilation step.

PR #2018, opened 2025-01-28
feat: A way to define default request headers for a client

johanandren:
Adds a method that accepts a key value header pair and returns a new client, any rpc method called with the returned client includes the added request headers. 

The returned client is a decorator and is backed by the same channel/connection as the initial header-free client.

### DIFF
--- a/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
@@ -46,6 +46,7 @@ public abstract class @{service.name}Client extends @{service.name}ClientPowerAp
       private final io.grpc.CallOptions options;
       private final Materializer mat;
       private final ExecutionContext ec;
+      private final MetadataImpl defaultMetadata;
 
       private Default@{service.name}Client(GrpcChannel channel, boolean isChannelOwned, ClassicActorSystemProvider sys) {
         this.channel = channel;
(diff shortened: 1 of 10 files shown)


codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
```diff
@@ -139,8 +151,30 @@ public abstract class @{service.name}Client extends @{service.name}ClientPowerAp
...
+          defaultMetadata.addEntry(key, value));
+      }
   }
 
+   /**
+    * The same client instance decorated to add the given key and value to the metadata of any request issued.
+    */
+   public @{service.name}Client addRequestHeader(String key, String value) {
+     // default implementation ignoring header for source compatibility
+     return this;
```
efgpinto, line 175: I'm struggling to understand why.. given this does not exist currently, it will only be generated when people upgrade.. so then, what is the possible source of incompatibility if this actually adds an header instead?

efgpinto (approved):
LGTM with a question for my own understanding.

codegen/src/main/twirl/templates/JavaClient/Client.scala.txt
johanandren: They could be extending the client interface with their own stub in tests, and then that would no longer compile because missing the method if we didn't have this default imp.

efgpinto: I see, thanks.

Issue #2021, opened 2025-01-31
Missing implementation for member of trait AkkaGrpcClient: addRequestHeader


johanandren:
In 2.5.2 I added a new method to the AkkaGrpcClient interfaces for both Java and Scala APIs. This method then get a default implementation in the concrete generated class for each client, but this means that a client generated with an older Akka gRPC no longer has all the needed methods in the interface.

I wonder if we should provide a default/guard impl in the interfaces instead, so that it is binary compatible and can be bumped without recompile/regeneration of the client sources.

PR #2022, opened 2025-01-31
fix: Keep source/binary compatibility with 2.5.1 for addRequestHeader

johanandren:
References #2021


johanandren:
Verified that this solves the issue for Scala, a project on akka-grpc-2.4.4 that gets a bump of the akka-grpc-runtime to a snapshot version of this PR works fine whie 2.5.2 causes the compilation error: 
```
[error] Missing implementation for member of trait AkkaGrpcClient:
[error]   def addRequestHeader(key: String, value: String): akka.grpc.scaladsl.AkkaGrpcClient = ???
```

Need to check Java/mvn as well though.

johanandren:
Java verified, works fine 👍

efgpinto (approved):

Issue #2034, opened 2025-02-27
Propagate GRPC Status Codes to Akka Insights


sean-walsh:
This came up with a client. We should demonstrate a pattern to capture GRPC status codes in order to present them to Akka Insights via Custom Metrics etc. and document it.

PR #2035, opened 2025-03-01
Groundwork to support telemetry for Akka gRPC responses

jackgene:
Lays the groundwork to be able to start gathering gRPC response metrics in Cinnamon.

References #2034

### DIFF
--- a/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
+++ b/codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
@@ -11,6 +11,8 @@ import java.util.Iterator;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
+import io.grpc.Status;
+
 import akka.japi.function.Function;
 import akka.actor.ActorSystem;
 import akka.actor.ClassicActorSystemProvider;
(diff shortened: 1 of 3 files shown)


codegen/src/main/twirl/templates/JavaServer/Handler.scala.txt
```diff
@@ -141,7 +143,18 @@ public class @{serviceName}HandlerFactory {
           String method = segments.next();
           if (segments.hasNext()) return notFound; // we don't allow any random `/prefix/Method/anything/here
           else {
-            return handle(spi.onRequest(prefix, method, req), method, implementation, mat, eHandler, system);
+            final akka.japi.Function<ActorSystem, akka.japi.Function<Throwable, Trailers>> instrumentedEHandler = s -> {
```
johanandren, line 146: This means we'll allocate two lambdas for every request, while handling it/matching the path, also when SPI/telemetry is not used. 

It would be better if a single callback per method has created up front on handler creation, and not happen at all if telemetry not used. That will require some tricker changes than these minimal amends though, it's not immediately obvious how to go about it to me, probably a private field per case and then use that inside the generated pattern match branches in the handle method.

jackgene: Good call. I mainly proposed this as an "as simple as possible" solution that disrupts the existing code-base as little as possible. But let me see if I can come up with something that addresses your concerns.

PR #2081, opened 2025-09-27
chore: nightly with java25

sebastian-alfers:
Added also 21 for scala 3.3


sebastian-alfers:
https://github.com/akka/akka-grpc/actions/runs/18057745435

PR #2095, opened 2025-11-04
chore: Updated sample and test certs

johanandren:
New certs have an expiration date in 2035, and are changed from some google wildcard to akkagrpcexample.example.com as CN


johanandren:
Not sure native image build checks certs, but green run in https://github.com/akka/akka-grpc/actions/runs/19065433078

And I have run server+client in native image and the samples manually to verify the self signed pairs and the CA works.

PR #2097, opened 2025-11-06
fix: SDK gRPC endpoint dispatchers

johanandren:
Was still not enough with the previous changes, since it used the system dispatcher


johanandren:
Damn, I didn't think all the way about the compatibility, will need to make it compatible before we can merge/release.

patriknw (approved):
LGTM

PR #2106, opened 2026-02-13
feat: Service codegen filters

johanandren:
Makes it possible to only generate code for specific clients and servers found in protobufs, for more complex project setups.


codegen/src/main/scala/akka/grpc/gen/scaladsl/ScalaCodeGenerator.scala
```diff
@@ -52,6 +52,19 @@ abstract class ScalaCodeGenerator extends CodeGenerator {
...
+    val ClientIncludeRegex = """(?i)client_include=([^,]+)""".r
+    val ClientExcludeRegex = """(?i)client_exclude=([^,]+)""".r
+    val ServerIncludeRegex = """(?i)server_include=([^,]+)""".r
+    val ServerExcludeRegex = """(?i)server_exclude=([^,]+)""".r
+    val clientInclude = ClientIncludeRegex.findFirstMatchIn(rawParams).map(_.group(1).split(";").toList).getOrElse(Nil)
+    val clientExclude = ClientExcludeRegex.findFirstMatchIn(rawParams).map(_.group(1).split(";").toList).getOrElse(Nil)
+    val serverInclude = ServerIncludeRegex.findFirstMatchIn(rawParams).map(_.group(1).split(";").toList).getOrElse(Nil)
+    val serverExclude = ServerExcludeRegex.findFirstMatchIn(rawParams).map(_.group(1).split(";").toList).getOrElse(Nil)
+    val filter = serviceFilter(clientInclude, clientExclude, serverInclude, serverExclude)
+
```
octonato: We could move this to a common place in codegen. We are repeating exactly the same in the `JavaCodeGenerator`

codegen/src/main/scala/akka/grpc/gen/Main.scala
```diff
@@ -46,6 +46,18 @@ object Main extends App {
...
+    parameters.getOrElse("client_include", "").split(";").toList.filter(_ != "")
+
+  private val clientExclude: List[String] =
+    parameters.getOrElse("client_exclude", "").split(";").toList.filter(_ != "")
+
+  private val serverInclude: List[String] =
+    parameters.getOrElse("server_include", "").split(";").toList.filter(_ != "")
+
+  private val serverExclude: List[String] =
+    parameters.getOrElse("server_exclude", "").split(";").toList.filter(_ != "")
```
octonato: Those four vals are never used.

We can just remove them. Those values are passed down in `req` and the java codegen will read directly from there.

docs/src/main/paradox/buildtools/gradle.md
```diff
@@ -61,10 +61,39 @@ Names and default values are provided.
...
+    ```
+    @@@
+
+The trait/interface is generated for any service that passes either the client or server filter.
+
+If `include` is empty, all services are included. The `exclude` patterns are applied to the result of `include`.
+
+Examples:
+- `*` matches any service name
+- `com.example.*` matches all services in the `com.example` package
```
octonato: This will probably match all services in `com.example.whatever.*` as well.

octonato (commented):
Looking good. 
Left some comments.

octonato (approved):

PR #2109, opened 2026-02-26
build: fix local scripted sbt runs

johanandren:
Scripted run in isolation so does not have access to tokenized resolver from main build by default.


johanandren:
Without the token repository the scripted tests cannot find any akka artifacts, so fails. I merged this as it was blocking working with akka-grpc locally.

PR #2110, opened 2026-03-02
bump: protoc 4.34.0 and grpc 1.79.0

johanandren:
Mostly to see how much breaks.


johanandren:
This is blocked by no upstream support in scalapb released yet. Scalapb 1.0.0 with support is still in alpha phase.

PR #2116, opened 2026-06-04
spike: bump to scalaPB v1.0.0-alpha.6

sebastian-alfers:
Branched out of https://github.com/akka/akka-grpc/pull/2110 to test the scalaPB `1.0.0-alpha.5` release.

---

There was a bug was in protoc-jar (the library, version 3.11.4): when told to download and run protoc 4.35.0, its extractStdTypes method fails on macOS ARM because it was released before Apple Silicon support existed. It doesn't know about the osx-aarch_64 classifier.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -253,7 +253,10 @@ lazy val pluginTesterScala = Project(id = "akka-grpc-plugin-tester-scala", base 
     crossScalaVersions := Dependencies.Versions.CrossScalaForLib,
     scalaVersion := Dependencies.Versions.CrossScalaForLib.head,
     Test / parallelExecution := false,
-    ReflectiveCodeGen.codeGeneratorSettings ++= Seq("flat_package", "server_power_apis"))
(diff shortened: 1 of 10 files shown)


johanandren:
Should be green now, but, plenty of things to investigate further:

 * docs about protoc-jar outdated
 * scripted test for scalapb-validate disabled, can that still work?
 * includeStdTypes no longer available, should the plugin automatically be able to add that runtime dependency (protobuf-java) for maven projects? If not we need to document
 * Binary compability of akka-grpc itself seems mostly fine, but what about the upstream proto libraries, this will likely warrant a major akka-grpc bump
 * we ofc need a non alpha of scalapb before we can release anything non release-candidate ourselves

johanandren:
Ok, not quite green, failures are the semi-disabled scalapb-validate, and std protobuf types proto-files no longer available, we may need to make it available for maven somehow anyway before this is ready.

johanandren:
scalapb validate now working with a workaround, includeStdTypes ported from protoc-jar to our maven plugin so it still works the same.

Issue #1582, opened 2022-02-23
Support external `proto` executable for Maven


raboof:
sbt supports using the system-installed `proto` executable (https://doc.akka.io/docs/akka-grpc/current/buildtools/sbt.html#using-a-local-protoc-command).

It would be useful to support the same from the Maven plugin

PR #2117, opened 2026-06-17
feat: allow alternative protoc binary in mvn

octonato:
References
 - https://github.com/akka/akka-grpc/issues/1582


codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
```diff
@@ -0,0 +1,63 @@
...
+      val segments = v.split('.')
+      if (segments(0) == "3" && segments.length >= 3) segments(1).toInt
+      else segments(0).toInt
+    }
+
+  /** The bare version number of a version string, dropping any non-numeric prefix (e.g. `-v3.25.8` -> `3.25.8`). */
+  def display(version: String): String =
+    Option(version).flatMap(VersionRegex.findFirstIn).getOrElse(version)
+
+  /** Runs `<executablePath> --version` and returns its reported version, or None if it cannot be queried. */
```
johanandren: This is not entirely true, `Process(...)` will throw an exception if the binary does not exist or cannot be executed. Not sure if it should be handled and turned to None or that exception is fine.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
```diff
@@ -93,11 +93,13 @@ object AkkaGrpcPlugin extends AutoPlugin {
       Compile / PB.recompile := {
         // hack to get our (dirty) hands on a proper sbt logger before running the generators
         generatorLogger.logger = streams.value.log
+        checkProtocVersion((Compile / PB.protocExecutable).value, (Compile / PB.protocVersion).value, streams.value.log)
```
johanandren, line 96: Not that it has any tangible overhead, but feels strange to do it over and over again, could that check be once on project load instead of on every recompile?

codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
```diff
@@ -0,0 +1,63 @@
...
+  def queryVersion(executablePath: String): Option[String] = {
+    import scala.sys.process.{ Process, ProcessLogger }
+    val output = new StringBuilder
+    val logger = ProcessLogger(line => output.append(line).append('\n'), line => output.append(line).append('\n'))
+    val exitCode = Process(Seq(executablePath, "--version")).!(logger)
+    if (exitCode == 0 && output.nonEmpty) Some(output.toString.trim) else None
```
johanandren: There's a concurrency issue hiding here, string builder is not thread safe, but we are constructing it in this method, handing off to be used in a different thread, and then expect to always be able to see the full output by `output.toString` which is not guaranteed.

Here's a safe version: https://github.com/lightbend/kalix-jvm-sdk/pull/2382/changes

johanandren: I misremembered the why though, Process calls Thread.join before returning, which is a memory barrier. However, the problem is in that there is one thread pumping stderr and one pumping stdout, and both writes to the non-thread-safe string builder which could garble text or worst case cause an ArrayOutOfBounds exception if both threads writes and decides to grow the backing buffer at the same time.

docs/src/main/paradox/buildtools/maven.md
```diff
@@ -141,6 +141,33 @@ which is a relative path to the project basedir. The below configuration overrid
     </plugin>
     ```
 
+## Using a local `protoc` executable
+
+By default, the plugin downloads `protoc` (via [protoc-jar](https://github.com/os72/protoc-jar)). If the
```
johanandren: Isn't that an impl detail, the docs can just say it downloads it instead?

docs/src/main/paradox/buildtools/maven.md
```diff
@@ -141,6 +141,33 @@ which is a relative path to the project basedir. The below configuration overrid
...
+        <version>${akka.grpc.version}</version>
+        <configuration>
+          <protocExecutable>/usr/bin/protoc</protocExecutable>
+        </configuration>
+    </plugin>
+    ```
+
+The value may also be provided on the command line via `-Dakka-grpc.protoc-executable=...`.
+
+The configured executable must belong to the same protobuf release as `protocVersion`. The plugin runs
```
johanandren, line 164: Above it says the `protocVersion` setting is ignored when `protocExecutable` is defined.

codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
octonato: We should throw. 
The user chooses to override it. If we return None we end up logging a warning

codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
octonato: Good catch, but not clear why we need ProcessIO. Should also work with the ProcessLogger as long as `output` is a StringBuffer.

Anyway, I will adapt. Thanks.

docs/src/main/paradox/buildtools/maven.md
octonato: It is and my bad. 

I adapt my CLAUDE.md telling it explicitly to not do that, but it still do it from time to time. Anyway, I should have caught it when reviewing myself.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
octonato: Yeah, I think we can do it. Also, it means that errors with the protoc binary version will be detected when loading the project, not on compile.

codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
octonato: I had a vague memory that we had discussed it before. Found the PR in Kalix. Anyway, I will go with ProcessIO here as well.

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
octonato: @johanandren, I didn't impl this at loading, but at first compilation. See effd65ef7.

At load time is not great because you may load the project in order to do other tasks. Like test an already compiled project. 

It will check the first on the first time. For sbt is quite useful. Less for maven.

johanandren: Isn't this entire block running one every recompile?

codegen/src/main/scala/akka/grpc/gen/ProtocVersion.scala
johanandren: Ok, so the comment is wrong then?

sbt-plugin/src/main/scala/akka/grpc/sbt/AkkaGrpcPlugin.scala
johanandren: Ah, the guard is elsewhere, gotcha

johanandren (approved):

PR #2136, opened 2026-09-04
fix: abandon connections stuck in CONNECTING via configurable connecting-timeout

sebastian-alfers:
Netty-backed akka-grpc clients could get permanently stuck with their ManagedChannel wedged in CONNECTING if a peer accepted the TCP connection but never completed the handshake (grpc-java has no timeout for that phase — grpc-java#1943); fixed by adding a configurable connecting-timeout (GrpcClientSettings.withConnectingTimeout, default 20s) that abandons a stuck connection attempt and retries.


runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -56,14 +57,41 @@ object ChannelUtils {
...
+            if (channel.getState(false) == ConnectivityState.CONNECTING) {
+              log.warning(
+                "gRPC client channel has been CONNECTING for more than {}, abandoning the stuck connection " +
+                "attempt and starting a new one (see akka.grpc.client config setting 'connecting-timeout')",
+                timeout)
+              channel.enterIdle()
+              channel.getState(true)
+            })
+      case _ => // connecting-timeout is 'infinite': watchdog disabled
+    }
```
johanandren: The watchdog is scheduled once per entry into `CONNECTING`, but it is never cancelled when the channel leaves that state. It only checks "is the channel `CONNECTING` right now" when it fires — not "is this the same connecting attempt I was scheduled for." 

If the channel legitimately cycles `CONNECTING → READY → IDLE → CONNECTING` again within the connecting-timeout window (plausible with idle-channel churn or frequent reconnects), a stale timer from the first episode can fire during the second, freshly-started legitimate attempt and abandon it prematurely even though that attempt hasn't been stuck for anywhere near 20s.

Something like capturing a cancellable for the timeout and cancelling that on READY or TRANSIENT_FAILURE seems reasonable.

runtime/src/main/resources/reference.conf
```diff
@@ -59,6 +59,13 @@ akka.grpc.client."*" {
   # interpreted as retrying 'indefinitely'.
   connection-attempts = 20
 
+  # If a connection attempt gets stuck in the CONNECTING state (the TCP connection is
+  # established but the TLS/HTTP2 handshake with the remote peer never completes, for example
+  # because a middlebox silently drops the connection) for longer than this duration, the client
+  # abandons that attempt and starts a new one rather than waiting on it forever.
+  # Only supported for the Netty client backend. Set to 'infinite' to disable.
+  connecting-timeout = 20s
```
johanandren, line 67: `infinite` as default and opt in would be safer, but on the other hand this seem like a genuine useful timeout for anyone. Maybe enable by default but with an even longer timeout?

Also, I think the doc comment above could be more concise and skip the "for example" part.

Worth a link to the upstream [issue](https://github.com/grpc/grpc-java/issues/1943) issue this sort of works around as well? Or would that anyway not solve our issue because using grpc-java in different ways from the vast majority?

johanandren: Ok, looks like a few other impls use 20s as a default, so maybe that's a good default here as well.

johanandren:
Might be worth broadening the scope to not be about that specific connection state, and rather be a full connection timeout, just like https://github.com/grpc/grpc-java/issues/1943 envisions.

That would be easier to understand as well, what does this config flag do, "it's a timeout for how long connecting can take" vs "it's a timeout for how long it can be in the state CONNECTING"

runtime/src/main/resources/reference.conf
sebastian-alfers: Would that warrant a new minor release as it potentially impacts existing behavior?

johanandren: Hmm, no, I think it is ok to have this as a new default.

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
johanandren, line 1: The watchdog's recovery path is `enterIdle() → getState(true) → back to CONNECTING`, which never passes through `TRANSIENT_FAILURE`. `connectionAttempts` is only incremented in the `TRANSIENT_FAILURE` arm, so a peer that permanently black-holes the handshake causes the watchdog to fire every connecting-timeout interval forever — `maxConnectionAttempts` and the `ClientConnectionException` give-up path never trigger. 

A user who sets `connection-attempts = 5` to fail fast against a dead peer will instead retry indefinitely if the peer specifically hangs mid-handshake rather than refusing the connection outright.

I think we should count the timeout against connection attempts as well.

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -56,16 +58,52 @@ object ChannelUtils {
...
+    // if the TCP connection is established but the peer never completes the TLS/HTTP2 handshake,
+    // the channel can stay CONNECTING forever, and since it never reaches TRANSIENT_FAILURE the
+    // connectionAttempts counter below never advances either. Abandon such stuck attempts after
+    // `connectingTimeout` by forcing the channel back to IDLE (tearing down the wedged
+    // subchannel) and immediately requesting a fresh connection.
+    //
+    // The returned Cancellable must be cancelled as soon as this particular CONNECTING episode
+    // ends (see `monitor` below) - otherwise a stale timer from an earlier episode could fire
+    // during a later, legitimate CONNECTING episode (e.g. after a normal idle-and-reconnect
+    // cycle) and abandon it prematurely, even though that one hasn't been stuck at all.
```
johanandren: This mostly just documents what is here in code, skip that and only leave the part about grpc-java not having a native timeout for CONNECTING

plugin-tester-scala/src/test/scala/example/myapp/helloworld/StuckConnectingReproducerSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.util.Failure
+import scala.util.Success
+import scala.util.Try
+
+/**
+ * Reproducer for Zoho case #16428: a Netty-backed akka-grpc client can get permanently wedged
```
johanandren: Skip internal issue tracker reference

plugin-tester-scala/src/test/scala/example/myapp/helloworld/StuckConnectingReproducerSpec.scala
```diff
@@ -0,0 +1,144 @@
...
+ * the client never gives up and retries on its own either - it's just stuck, forever, even once
+ * the peer is perfectly healthy again.
+ *
+ * Both tests below run the exact same scenario - a peer that is initially a black hole and later
+ * becomes a real, healthy `GreeterService` on the same port - and differ only in whether
+ * `GrpcClientSettings.connectingTimeout` (the fix) is enabled:
+ *   - with it disabled (`Duration.Inf`), the client never notices the peer became healthy
+ *   - with it enabled, the client abandons the stuck attempt on its own and a later attempt
+ *     succeeds against the now-healthy peer
+ */
```
johanandren, line 37: Not critical but I think this could explain the core of the problem in like 3-4 lines instead of this. Exactly what the tests do is already in the code for example, that the test fails before the test is also not something to document (even if it is good practice that claude verified that)

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -56,16 +58,52 @@ object ChannelUtils {
...
+          scheduleOnce(
+            timeout,
+            () =>
+              if (channel.getState(false) == ConnectivityState.CONNECTING) {
+                log.warning(
+                  "gRPC client channel has been CONNECTING for more than {}, abandoning the stuck connection " +
+                  "attempt and starting a new one (see akka.grpc.client config setting 'connecting-timeout')",
+                  timeout)
+                channel.enterIdle()
+                channel.getState(true)
```
johanandren: `getState(false) == CONNECTING` is read, then `enterIdle()` is called, this is not atomic. If grpc-java's internal `SynchronizationContext` has a `READY` transition already queued (handshake just completed) when the watchdog task runs, `enterIdle()` could still force a just-established connection back to `IDLE`.

I don't think we can do much given how the API looks, but may be worth mentioning with a comment.

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -56,23 +59,77 @@ object ChannelUtils {
...
+                    // Check state before claiming `advanced`: if we're not actually stuck, we must
+                    // NOT prevent the real notifyWhenStateChanged callback from doing its job below.
+                    //
+                    // Note: getState(false) here and enterIdle() below are not atomic - grpc-java
+                    // exposes no "abandon only if still CONNECTING" primitive - so in a narrow
+                    // window where the handshake completes concurrently with this check, we could
+                    // force a just-READY connection back to IDLE. Harmless (the client immediately
+                    // reconnects) but wasteful; accepted given the API available to us.
+                    if (channel.getState(false) == ConnectivityState.CONNECTING && advanced
+                        .compareAndSet(false, true)) {
```
johanandren, line 110: `advanced` is claimed before `channel.enterIdle()/getState(true)` run, with no `try/catch`. If either throws (e.g. a concurrent `.close()`), the real `notifyWhenStateChanged` callback for that episode is permanently silenced and no new one gets registered — the client hangs forever with no failure signal, which is the exact failure mode this PR exists to fix. Cheap insurance: wrap that block in a `try/catch` that falls back to failing the promises.

runtime/src/main/scala/akka/grpc/internal/ChannelUtils.scala
```diff
@@ -56,23 +59,77 @@ object ChannelUtils {
...
+      } else Some(nextAttempts)
+
     def monitor(currentState: ConnectivityState, connectionAttempts: Int): Unit = {
       log.debug(s"monitoring with state $currentState and connectionAttempts $connectionAttempts")
+
+      // Both the watchdog below and the "normal" state-change callback registered further down
+      // can end up deciding what happens next for this one CONNECTING episode - whichever fires
+      // first should win, and the other must become a no-op (guarded by this flag). Without it, a
+      // watchdog-driven continuation and the notifyWhenStateChanged-driven one could both fire for
+      // the same episode, double-counting the attempt or acting on stale data.
```
johanandren: Again quite verbose and chatty comments here and below, see if you can boil them down to something more concise 2-3 lines should be enough to explain why this exists.

johanandren (approved):
LGTM!

